PR #2, opened 2022-01-19
'Solution'

alexandra-halytska:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
IvanRamyk, line 25: If you multiply the distance by the distance you get the area. So it is expected that other has `float` or `int` type

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            return Distance(self.km * other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
IvanRamyk, line 31: The same case as in the `__mul__` method

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!!

PR #5, opened 2022-01-19
'Solution'

serhii56465:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __eq__(self, other):
+        if not isinstance(other, Distance):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other):
+        if not isinstance(other, Distance):
+            return self.km <= other
+        return self.km <= other.km
```
TPolina, line 55: Probably, it would be better if you use the trick we discussed at the lesson. Try using __gt__ method here. Consider same changes in other methods.

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        )
+
+    def __truediv__(self, other):
+        return Distance(
+            km=round((self.km / other), 2)
+        )
+
+    def __lt__(self, other):
+        if not isinstance(other, Distance):
+            return self.km < other
```
TPolina, line 39: parameter other can be only of type Distance, int or float. Consider adding check for parameter other to be instance of int or float. The same things in further methods.

TPolina (changes requested):

IgorOlifirenko (approved):

PR #8, opened 2022-01-19
Solution

nikon4uk:


IgorOlifirenko (approved):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
IvanRamyk, line 24: `other` can't be an instance of `Distance` only a number (float or int). 
Because if you multiply a kilometer by a kilometer, you get a square kilometer.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
IvanRamyk, line 29: The same case as in the `__mul__` method

IvanRamyk (changes requested):

app/main.py
```diff
@@ -21,14 +21,16 @@ def __iadd__(self, other):
         return self
 
     def __mul__(self, other):
-        if isinstance(other, Distance):
-            return Distance(self.km * other.km)
-        return Distance(self.km * other)
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km * other, 2))
+        else:
+            raise TypeError(f"{type(other)} can`t be multiply on {type(self.km)}")
```
cloudprogramm, line 27: I can't multiply class object by class object. Write it in the condition, and remove this TypeError line.

app/main.py
```diff
@@ -21,14 +21,16 @@ def __iadd__(self, other):
...
 
     def __truediv__(self, other):
-        if isinstance(other, Distance):
-            return Distance(round(self.km / other.km, 2))
-        return Distance(round(self.km / other, 2))
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError(f"{type(other)} can`t be divide on {type(self.km)}")
```
cloudprogramm, line 33: The same problem.

cloudprogramm (changes requested):

app/main.py
nikon4uk: https://github.com/mate-academy/py-distance-class/pull/8#discussion_r787886731 I followed the instructions of the mentor, added an TypeError because, according to the condition from the example, these operations are available only for (int, float) types. Now i little bit confused what i need to fix.

nikon4uk: "other can't be an instance of Distance only a number (float or int)" i check it , if other is not int or float i return error becouse the meaning of the Distance class is to perform operations with distance, not area, also I cannot multiply the distance by a string or another class, so I will check for valid types otherwise I return an exception. I understand that there is a number in another class, but it is written in a variable that denotes kilometers, and a kilometer per kilometer is an area

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
```
Nazar-kob, line 21: you should think about "else" I am not sure it will good for another type

Nazar-kob (approved):

IvanRamyk (approved):
Good job!

PR #10, opened 2022-01-19
Solution

alix-mak:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(
+            km=self.km * other
+        )
+
+    def __truediv__(self, other):
+        return Distance(
+            km=round(self.km / other, 2)
```
Krimat, line 35: Avoid using magic constants

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            )
+        return Distance(
+            km=self.km + other
+        )
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
```
Krimat, line 25: Should add some check on variable 'other' type

Krimat (approved):

app/main.py
IvanRamyk: It is optional for this task

IvanRamyk (approved):
Well done!

PR #16, opened 2022-01-20
Solution

phantom-terminal:


app/main.py
```diff
@@ -1,3 +1,86 @@
...
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
+        else:
+            raise TypeError("The right operand should be int or float.")
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
vldtsw, line 30: We will get area by multiplying distance on distance. I don't think we need that in task.

app/main.py
```diff
@@ -1,3 +1,86 @@
...
+            return Distance(self.km * other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        else:
+            raise TypeError("The right operand should be int or float.")
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
vldtsw, line 38: We also don't need to divide to distances, I guess

vldtsw (changes requested):

vldtsw (approved):

IvanRamyk (approved):
Well done!

PR #17, opened 2022-01-20
Solution

anitaposliedova:


Oleksandr563487953 (changes requested):
Lines #2, 3 should be deleted.

anitaposliedova:
> 
However under the terms of the assignment, we had to implement the magical method of init

Oleksandr563487953:
> However under the terms of the assignment, we had to implement the magical method **init**.
init is OK, I mean the next lines:
    # Write your code here
    pass
it should be deleted

anitaposliedova:
> > However under the terms of the assignment, we had to implement the magical method **init**.
> > init is OK, I mean the next lines:
> > # Write your code here
> > pass
> > it should be deleted

They are already deleted, the second and third lines in the code are the method init.

Oleksandr563487953:
Sorry, I didn`t notice that.

Oleksandr563487953 (approved):

IvanRamyk (approved):
God job!

PR #22, opened 2022-01-20
'Solution'

Zikkolion:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
```
nikon4uk, line 32: The same case as in the __mul__ method

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __iadd__(self, other):
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
```
nikon4uk, line 26: other can't be an instance of Distance only a number (float or int).
Because if you multiply a kilometer by a kilometer, you get a square kilometer.

nikon4uk (changes requested):

liuda-chuzdiuk (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __lt__(self, other):
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
+
+    def __le__(self, other):
+        return not self > other
+
+    def __ge__(self, other):
+        return self.__gt__(other) or self.__eq__(other)
```
MasterpieceElbow, line 49: Why do you use two different styles in `__le__`  and  `__ge__`? Make it in one style

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #24, opened 2022-03-25
Solution Task Distance Class. Katiukha

KatiukhaO:


app/main.py
```diff
@@ -1,9 +1,6 @@
-from typing import Union
+class Distance():
```
Serhii-Leonenko, line 1: $1 i think you don't need ( ) here

app/main.py
```diff
@@ -13,62 +10,53 @@ def __repr__(self):
...
+        return Distance(self.km + other)
 
     def __iadd__(self, other):
-        try:
+        if isinstance(other, Distance):
             self.km += other.km
-        except AttributeError:
-            self.km += other
+        else:
+            self.km = self.km + other
```
Serhii-Leonenko, line 21: $21 it's better to use operator "+="

app/main.py
Serhii-Leonenko, line 1: $1 i think you don't need ( ) here
$21 it's better to use operator "+="
$29 remove pass

Serhii-Leonenko (changes requested):
Please, look at comments and make changes

app/main.py
KatiukhaO: thanks , i fix it

Serhii-Leonenko (approved):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, (Distance)):
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
TPolina, line 30: The same case as in the `__mul__ `method

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
```
TPolina, line 15: You cannot add any type of data to the distance, so make an appropriate check for the data type. Check all methods for this issue

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other):
```
TPolina, line 54: optimize this method as well

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
```
TPolina, line 49: optimize this method: use the method you already wrote

TPolina (changes requested):

TPolina (approved):
Well done!

PR #25, opened 2022-04-29
distance v1

SairusQw:


app/main.py
```diff
@@ -1,3 +1,40 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __add__(self, other):
+        return Distance(other + self.km)
```
IvanRamyk, line 6: It works only if `other` is an integer. If it is a `Distance`, the method tries to add a `Distance` object and an integer inside the constructor. 
As a result, you create a `Distance` object passing another Distance to the constructor.

Could you use an `if` construction to avoid such a problem?

app/main.py
```diff
@@ -1,3 +1,40 @@
...
+
+    def __iadd__(self, other):
+        self.km = other + self.km
```
IvanRamyk, line 9: The same issue

IvanRamyk (changes requested):

PR #29, opened 2022-05-14
Solution

YuriiKindrat:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        if isinstance(other, Distance):
+            return True if self.km < other.km else False
```
Magus6784, line 33: You can use just 
self.km < other.km
because its True or False already
Same to $35, $39, $41 etc

Magus6784 (changes requested):

app/main.py
Magus6784: A lot of redundant code comes out. Exsept this, everything ok

Magus6784 (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
TPolina, line 23: duplicate `return`: take it outside if-else

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            return self.km > other
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other):
+        return any([self.__lt__(other), self.__eq__(other)])
```
TPolina, line 50: not the best way to use magic methods. not for nothing they are magic :)
`__le__` == not `__gt__`

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __ge__(self, other):
+        return any([self.__gt__(other), self.__eq__(other)])
```
TPolina, line 53: same here

TPolina (changes requested):

TPolina (changes requested):
pls, fix remark above

TPolina (approved):
Good!

PR #32, opened 2022-05-16
Solution

4ndyua:


Flashmobber (approved):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, int):
```
TPolina, line 13: `other` can be also the type of `float`. Consider this in all methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
```
TPolina, line 15: what if `other` will be not of the type `int`, `float`, and `Distance`? better check that `other` is `distance` object. Consider this in all methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km > other if (isinstance(other, int))\
+            else self.km > other.km
+
+    def __eq__(self, other):
+        return self.km == other if (isinstance(other, int))\
+            else self.km == other.km
+
+    def __le__(self, other):
+        return self.km <= other if (isinstance(other, int))\
+            else self.km <= other.km
```
TPolina, line 44: optimize this method: use the method you already wrote

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other if (isinstance(other, int))\
+            else self.km >= other.km
```
TPolina, line 48: same here

TPolina (changes requested):

TPolina (changes requested):

TPolina (approved):
Well done!

PR #34, opened 2022-05-16
'Solution'

darten360:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km == other if (isinstance(other, int)) \
+            else self.km == other.km
+
+    def __le__(self, other):
+        return self.km <= other if (isinstance(other, int)) \
+            else self.km <= other.km
+
+    def __ge__(self, other):
+        return self.km >= other if (isinstance(other, int)) \
+            else self.km >= other.km
```
Magus6784, line 47: rewrite using opposite function

Magus6784 (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        if isinstance(other, Distance):
+            self.km += other.km
+        return self
```
Magus6784, line 23: No need $20 because you do the same at $23

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        return self.km == other if (isinstance(other, (int, float))) \
+            else self.km == other.km
+
+    def __le__(self, other):
+        return not self.km > other if (isinstance(other, (int, float))) \
+            else self.km <= other.km
+
+    def __ge__(self, other):
+        return not self.km < other if (isinstance(other, (int, float))) \
+            else self.km >= other.km
```
Magus6784, line 49: Refactore (watch slack for advice)

Magus6784 (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other.km \
+            if isinstance(other, Distance) else self.km > other
+
+    def __eq__(self, other):
+        return self.km == other.km \
+            if isinstance(other, Distance) else self.km == other
+
+    def __le__(self, other):
+        return self.km <= other.km \
+            if isinstance(other, Distance) else self.km <= other
```
TPolina, line 43: here you can use the magic method you already wrote (don't forget to use it as a magic method)
le == not gt

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other.km \
+            if isinstance(other, Distance) else self.km >= other
```
TPolina, line 47: same here: ge == not lt

MasterpieceElbow (approved):

PR #35, opened 2022-05-17
Solution

rotsen18:


KatiukhaO (approved):
intresting way, $5 - 7 lines

rotsen18:
DRY;) thanks for review

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __add__(self, other):
+        return Distance(self.km + self.get_other_km(other))
+
+    def __iadd__(self, other):
+        self.km += self.get_other_km(other)
+        return self
+
+    def __mul__(self, other):
+        self.km *= other
+        return self
```
TPolina, line 24: in this method, you should return other Distance object and not modify given

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __truediv__(self, other):
+        self.km = round(self.km / other, 2)
+        return self
```
TPolina, line 28: same as in `__mul__` method

TPolina (changes requested):

TPolina (approved):
Great! interesting solution

rotsen18:
Thanks ;)

PR #36, opened 2022-05-17
solution

BohdanKryven:
make solution to hw distance class

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,3 @@
-from __future__ import division
-
-
 class Distance:
     def __init__(self, km):
         self.km = km
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,65 @@
+from __future__ import division
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        self.km *= other
+        return self
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            self.km = round(self.km / other.km, 2)
+        else:
+            self.km = round(self.km / other, 2)
+        return self
+
+    def __len__(self):
+        return self.km
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __gt__(self, other):
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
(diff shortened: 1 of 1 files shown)


Chardona (approved):
Good as well

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        self.km *= other
+        return self
```
TPolina, line 27: same as in `__add__` method

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return Distance(self.km)
```
TPolina, line 16: in this method you modify `self.km`, but you shouldn't do this. You have to return different `Distance` object and pass to it new value, that is calculated by `self.km + other.km` or `self.km + other`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        self.km *= other
+        return Distance(self.km)
```
TPolina, line 27: the same issue as in `__add__` method

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __truediv__(self, other):
+        self.km = round(self.km / other, 2)
+        return Distance(self.km)
```
TPolina, line 31: the same issue as in `__add__` method

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km == other.km
+        return self.km == other
+
+    def __gt__(self, other):
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __ge__(self, other):
+        return not self.km < other
```
TPolina, line 47: correct: `return not self < other`. This will call the `__lt__` magic method that will do all necessary checks.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __lt__(self, other):
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __le__(self, other):
+        return not self.km > other
```
TPolina, line 55: same here

TPolina (changes requested):

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
```
MasterpieceElbow, line 25: As Polina said, you can only multiply kilometers on a number. Remove `if` block.

MasterpieceElbow (changes requested):

app/main.py
BohdanKryven: Finally fixed. I fixed this part of code and push on develop branch, but pull request was on master branch. This is finally good solution

PR #38, opened 2022-05-17
All done

Flashmobber:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = kilometer
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return Distance(kilometer=self.km + other.km) \
```
Hectorovich, line 15: Maybe its better to make variable`s name shorter to avoid this line separation.

Hectorovich, line 15: You can just use this construction: `return Distance(self.km + other.km) \`... and then you won`t need line separation in this block.

Hectorovich (changes requested):

app/main.py
Flashmobber: I can't. Because if I want to use somthing like this:

> dst = Distance(10)
> print(dst + 10)

I will get the error AttributeError: 'int' object has no attribute 'km'

Hectorovich (approved):

app/main.py
```diff
@@ -1,3 +1,54 @@
 class Distance:
-    # Write your code here
-    pass
+
+    km = 0
```
TPolina, line 3: what is the purpose of this line?

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other.km \
+            if isinstance(other, Distance) else self.km < other
```
TPolina, line 35: Better not use ternary operator here: it’s good for example for returning one of two strings not for calculations and comparison. Fix this issue throughout your code

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        return self.km > other.km \
+            if isinstance(other, Distance) else self.km > other
+
+    def __eq__(self, other):
+        return self.km == other.km \
+            if isinstance(other, Distance) else self.km == other
+
+    def __le__(self, other):
+        return self.km <= other.km \
+            if isinstance(other, Distance) else self.km <= other
```
TPolina, line 47: here you can use the magic method you already wrote (don't forget to use it as a magic method)
le == not gt

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(kilometer=self.km + other.km)
+        return Distance(kilometer=self.km + int(other))
```
TPolina, line 15: you don't need to convert `other` to int.

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        return not self.__gt__(other)
```
TPolina, line 46: magic methods are used in a different way
Example: `a + b` -> method `__add__` will be called

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __ge__(self, other):
+        return not self.__lt__(other)
```
TPolina, line 49: same as in `__le__` method

TPolina (changes requested):

app/main.py
TPolina: not fixed

PR #39, opened 2022-05-17
Distance-Class Solution

mens0:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        if isinstance(other, int):
+            return Distance(self.km + other)
+
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, int):
+            self.km += other
+            return self
+        self.km += other.km
```
Y-Havryliv, line 21: it's better to use magic method here

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        res = self.km / other
+        return Distance(round(res, 2))
+
+    def __lt__(self, other):
```
Y-Havryliv, line 31: it works only with numbers, what if it's Distance class

Y-Havryliv (changes requested):

Y-Havryliv (approved):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers"
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, int):
```
TPolina, line 12: `other` can be also of type `float`. Check this in all your methods

TPolina (changes requested):

TPolina (approved):
Good!

PR #40, opened 2022-05-17
Solution

Kiawami:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __gt__(self, other):
+        if isinstance(other, int):
+            return True if self.km > other else False
+        if isinstance(other, Distance):
+            return True if self.km > other.km else False
+
+    def __eq__(self, other):
+        if isinstance(other, int):
+            return True if self.km == other else False
```
DmytroSvirsa, line 51: $51 you do not need `if/else`. Result of `self.km == other` already is `bool`

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, int):
+            self.km += other
+
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, int):
+            return Distance(self.km * other)
```
DmytroSvirsa, line 27: $27 not sure if this is necessary, but we need to multiply only by `int` the same for $32`__truediv__`

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, int):
+            return self.km == other
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+    def __le__(self, other):
+        if isinstance(other, int):
+            return self.km <= other
+        if isinstance(other, Distance):
+            return self.km <= other.km
```
TPolina, line 55: optimize this method: use the method you already wrote

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __ge__(self, other):
+        if isinstance(other, int):
+            return self.km >= other
+        if isinstance(other, Distance):
+            return self.km >= other.km
```
TPolina, line 61: same here

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, (int, float)):
+            return self.km == other
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+    def __le__(self, other):
+        if isinstance(other, (int, float)):
+            return not self.km > other
+        if isinstance(other, Distance):
+            return not self.km > other.km
```
TPolina, line 55: optimize this method: `return not self > other`

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __ge__(self, other):
+        if isinstance(other, (int, float)):
+            return not self.km < other
+        if isinstance(other, Distance):
+            return not self.km < other.km
```
TPolina, line 61: same here

TPolina (changes requested):

TPolina (approved):

PR #41, opened 2022-05-17
'Solution'

Hectorovich:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
katevasilenko, line 24: You shoudn't check if other if Distance. As Polina explaned somewhere: if you multiply km to km you'll get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
katevasilenko, line 29: Here too. Just divide to other

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(round((self.km / other.km), 2))
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other):
+        if isinstance(other, Distance):
```
katevasilenko, line 34: I guess here and methods below you don't need to check if other is Distance

katevasilenko (changes requested):

katevasilenko (approved):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
TPolina, line 30: `other` can be also of type `Distance`, then your code will not work. Add necessary checks. Fix this issue throughout your code

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 42: same as in `__le__` method: ge == not lt

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
```
TPolina, line 26: but here `other` can't be an instance of `Distance` only a number (float or int) Because if you multiply a kilometer by a kilometer, you get a square kilometer))

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round((self.km / other.km), 2))
+        return Distance(round((self.km / other), 2))
```
TPolina, line 31: same as in `__mul__` method

TPolina (approved):

PR #42, opened 2022-05-17
'Solution'

Korenely:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __gt__(self, other):
+        return self.km > other.km \
+            if isinstance(other, Distance) else self.km > other
+
+    def __eq__(self, other):
+        return self.km == other.km \
+            if isinstance(other, Distance) else self.km == other
+
+    def __le__(self, other):
+        return self.km <= other.km \
```
Aleksei-Isaev, line 39: You may refactor your code in this way:
return not self > other
*because it's a magic method)

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            if isinstance(other, Distance) else self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other.km \
```
Aleksei-Isaev, line 43: Almost the same situation, as above:
return not self < other
*because it's a magic method too)

Aleksei-Isaev (approved):

TPolina (changes requested):

TPolina:
also, you didn't fix Oleksii remarks

MasterpieceElbow (approved):

PR #45, opened 2022-05-18
Katya's Solution

katevasilenko:


Hectorovich (approved):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, int):
```
TPolina, line 12: other can be also of type `float`. Consider this in all your methods

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
TPolina, line 31: `other` can be of type `Distance`, then your code will not work. Consider this in the following methods as well

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
+
+    def __len__(self):
+        return int(self.km)
```
TPolina, line 46: return just `self.km` here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
```
TPolina, line 40: here use the magic method you already wrote. Don't forget to use it as a magic method
le == not gt

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 43: same here

TPolina (changes requested):

TPolina (approved):

PR #47, opened 2022-05-18
Solution

NOSSUDAK:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, num):
+        self.km *= num
+        return self
```
TPolina, line 25: in this method you modify `self.km`, but you shouldn't do this. You have to return a different `Distance` object and pass to it a new value, that is calculated by `self.km * num`

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __truediv__(self, num):
+        res = round(self.km / num, 2)
+        self.km = res
+        return self
```
TPolina, line 30: same issue as in `__mul__` method

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __lt__(self, num):
+        return self.km < num
```
TPolina, line 33: `num` (better use `other`) can be of type `Distance`, then your code will not work. Consider this in all the following methods

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __gt__(self, num):
+        return self.km > num
+
+    def __eq__(self, num):
+        return self.km == num
+
+    def __le__(self, num):
+        return self.km <= num
```
TPolina, line 42: here use the method you wrote (don't forget to use it as a magic method)
le == not gt

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __ge__(self, num):
+        return self.km >= num
```
TPolina, line 45: same as in `__le__` method

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        return not self.__gt__(other)
```
TPolina, line 46: your magic method does no magic) use it as a magic method

TPolina (changes requested):

TPolina (approved):

PR #48, opened 2022-05-18
'Solution_18/05_#1'

CheshkoSergey:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        self.km = km
+
+    def __str__(self):
+        print(f"Distance: {self.km} kilometers")
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if type(other) == int:
```
katevasilenko, line 13: it could be also float so check if it not Distance(here and below)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return Distance(km=self.km + other)
+        else:
+            return Distance(km=self.km + other.km)
+
+    def __iadd__(self, other):
+        if type(other) == int:
```
katevasilenko, line 19: The same

katevasilenko, line 19: The same

katevasilenko, line 19: The same

katevasilenko (changes requested):

katevasilenko:
and also you hadn't pass all checks

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if type(other) in [int, float]:
+            return self.km <= other
+        else:
+            return self.km <= other.km
+
+    def __ge__(self, other):
+        if type(other) in [int, float]:
+            return self.km >= other
+        else:
+            return self.km >= other.km
```
Magus6784, line 59: rewrite le and ge with maghic methoda gt and lt
le NOT gt
ge NOT lt

Magus6784 (changes requested):

katevasilenko (approved):

app/main.py
```diff
@@ -1,3 +1,62 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        print(f"Distance: {self.km} kilometers")
```
TPolina, line 7: this method should return value, not print it

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if type(other) in [int, float]:
+            return self.km == other
+        else:
+            return self.km == other.km
+
+    def __le__(self, other):
+        if type(other) in [int, float]:
+            return self.km <= other
+        else:
+            return self.km <= other.km
```
TPolina, line 53: optimize this method, use the magic method you already wrote. Don't forget to use it as a magic method

app/main.py
TPolina, line 59: same here

TPolina (changes requested):

TPolina (approved):
Well done!

PR #49, opened 2022-05-18
Solution

lenasharabura:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
-    # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {str(self.km)} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={str(self.km)})"
```
Magus6784, line 9: No need str(self.km) we already in string

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, int):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
```
Magus6784, line 19: we can use
self.km += other.km

Magus6784: Same to $21

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, int):
+            return self.km == other
+
+    def __len__(self):
+        return self.km
+
+
+print(Distance(50) > Distance(50))
```
Magus6784, line 60: Delete "print"

Magus6784 (changes requested):
Some small errors, but in general I see some solutions that are better than in my code
Good Job!

TPolina (approved):
Good job!

PR #50, opened 2022-05-18
'Solution'

dashaprotsko:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other.km \
+            if isinstance(other, Distance) else self.km < other
```
TPolina, line 28: better not use ternary operator here. It is good for returning, for example,  one of two `int` or `string`, not more complex things with comparison or calculations.

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __gt__(self, other):
+        return self.km > other.km \
+            if isinstance(other, Distance) else self.km > other
```
TPolina, line 32: same here

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __eq__(self, other):
+        return self.km == other.km \
+            if isinstance(other, Distance) else self.km == other
```
TPolina, line 36: same here

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __le__(self, other):
+        return self.km <= other.km \
+            if isinstance(other, Distance) else self.km <= other
```
TPolina, line 40: optimize this method: use the method you already wrote. Don't forget to use it as magic method
le == not gt

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other.km \
+            if isinstance(other, Distance) else self.km >= other
```
TPolina, line 44: same as in `__le__` method

TPolina (changes requested):

PR #51, opened 2022-05-18
Solution

RomaTsybulia:


darten360 (approved):
Nice)

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+        return Distance(self.km)
```
TPolina, line 16: in this method you modify `self.km`, but you shouldn't do this. You have to return a different `Distance` object and pass to it a new value, that is calculated by `self.km + other.km` or `self.km + other`

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        else:
+            self.km = self.km + other.km
+        return self
+
+    def __mul__(self, other):
+        if not isinstance(other, Distance):
+            self.km *= other
+        else:
+            self.km *= other.km
+        return Distance(self.km)
```
TPolina, line 30: `other` can't be an instance of `Distance` only a number (float or int)
Also don't modify `self.km`

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __truediv__(self, other):
+        if not isinstance(other, Distance):
+            self.km /= other
+        else:
+            self.km /= other.km
+        return Distance(round(self.km, 2))
```
TPolina, line 37: same issue as in `__mul__` method

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __eq__(self, other):
+        if not isinstance(other, Distance):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other):
+        if not isinstance(other, Distance):
+            return self.km <= other
+        return self.km <= other.km
```
TPolina, line 57: optimize this method: use the method you already wrote. Don't forget to use it as a magic method

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __ge__(self, other):
+        if not isinstance(other, Distance):
+            return self.km >= other
+        return self.km >= other.km
```
TPolina, line 62: same here

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other):
+        if not isinstance(other, Distance):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other):
+        return not self.__gt__(other)
```
TPolina, line 49: not for nothing magic methods are magic)
best usage: `not self > other`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __ge__(self, other):
+        return not self.__lt__(other)
```
TPolina, line 52: same here

TPolina (changes requested):

TPolina (approved):

PR #52, opened 2022-05-18
Distance class Solution 

Ed12816:
Some magical methods to definitely start running on Monday

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,48 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f'Distance: {self.km} kilometers.'
+
+    def __repr__(self):
+        return f'Distance(km={self.km})'
+
+    def __add__(self, other):
+        return Distance(self.km + other.km) \
+            if isinstance(other, Distance) else Distance(self.km + other)
+
+    def __iadd__(self, other):
+        self.km += other.km \
+            if isinstance(other, Distance) else other
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other.km \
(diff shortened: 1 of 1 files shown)


NOSSUDAK (approved):
Mentors say that it is better to use " instead of '

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __str__(self):
+        return f'Distance: {self.km} kilometers.'
+
+    def __repr__(self):
+        return f'Distance(km={self.km})'
+
+    def __add__(self, other):
+        return Distance(self.km + other.km) \
+            if isinstance(other, Distance) else Distance(self.km + other)
```
TPolina, line 14: better not use ternary operator in your methods. It is good for returning, for example,  one of two `int` or `string`, not more complex things with comparison or calculations. Fix this in all your methods

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km > other.km \
+            if isinstance(other, Distance) else self.km > other
+
+    def __eq__(self, other):
+        return self.km == other.km \
+            if isinstance(other, Distance) else self.km == other
+
+    def __le__(self, other):
+        return self.km <= other.km \
+            if isinstance(other, Distance) else self.km <= other
```
TPolina, line 41: optimize this method: use the magic method you already wrote

TPolina, line 41: optimize this method: use the magic method you already wrote (don't forget to use it as a magic method)

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other.km \
+            if isinstance(other, Distance) else self.km >= other
```
TPolina, line 45: optimize this method as well

TPolina, line 45: optimize this method as well

TPolina (changes requested):

app/main.py
Ed12816: rebuild methods to suit your requirements and principles

TPolina (approved):
Well done!

PR #53, opened 2022-05-18
Solution

Chardona:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __iadd__(self, second):
+        if isinstance(second, Distance):
+            self.km += second.km
+            return self
+        elif isinstance(second, int):
+            self.km += second
+            return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
katevasilenko, line 27: You shouldn't check if **other** is in class **Distance** in **__mul__** magic method

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return Distance(self.km * other.km)
+        elif isinstance(other, int):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
katevasilenko, line 33: The same in **trudiv** magic method

katevasilenko (changes requested):

Magus6784 (changes requested):

katevasilenko (approved):

app/main.py
```diff
@@ -1,3 +1,54 @@
 class Distance:
     # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        print(f"Distance: {self.km} kilometers.")
```
TPolina, line 7: this method should return string not print it

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, int):
```
TPolina, line 15: other can be also of type `float`. Consider this throughout your code

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other):
+
+        if isinstance(other, int):
+            return Distance(self.km * other)
+        else:
+            return Distance(self.km * other.km)
```
TPolina, line 30: `other` can't be an instance of `Distance`, only a number (float or int)

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, int):
+            return Distance(round(self.km / other, 2))
+        else:
+            return Distance(round(self.km / other.km, 2))
```
TPolina, line 36: same as in `__mul__` method

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __eq__(self, other):
+        return self.km == other
```
TPolina, line 39: `other` can be of type `Distance`, then your code will not work. Consider this in following methods

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, int):
```
TPolina, line 19: other can be also of type `float`

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __gt__(self, other):
+        return self.km > other
```
TPolina, line 38: `other` can be of type `Distance`, then your code will not work.

TPolina, line 38: `other` can be of type `Distance`, then your code will not work. Consider this in all the following methods

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 44: here use the magic method you wrote: `return not self < other`

TPolina, line 44: here use the magic method you wrote: `return not self < other`

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __le__(self, other):
+        return self.km <= other
```
TPolina, line 47: same as in `__ge__` method

TPolina, line 47: same as in `__ge__` method

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __gt__(self, other):
+        return self.km > other
+
+    def __lt__(self, other):
+        return self.km < other
```
TPolina, line 41: same as in `__gt__` method

TPolina, line 41: same as in `__gt__` method

TPolina (changes requested):

app/main.py
TPolina, line 19: `other` can be of type `float`

TPolina (changes requested):

TPolina (approved):

PR #54, opened 2022-05-18
'Solution'

Yurnerroo:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
Magus6784, line 15: For __add__ and __iadd__ you must add check for type(int)
elif isinstance(other, int):
Not sure about float, but for int - yes

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
Magus6784, line 23: Duplicate code $20 and $23
Remove one return

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        if isinstance(other, Distance):
+            return True if self.km < other.km else False
+        else:
+            return True if self.km < other else False
```
Magus6784, line 35: Dont need so big $33 $35 etc
self.km < other already return True or False

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return True if self.km == other.km else False
+        else:
+            return True if self.km == other else False
+
+    def __le__(self, other):
+        return True if self.__lt__(other) or self.__eq__(other) \
+            else False
```
Magus6784, line 51: Good direction but for __le__ opposite is
not __gt__ 
so rewrite with one another func

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __ge__(self, other):
+        return True if self.__gt__(other) or self.__eq__(other) \
+            else False
```
Magus6784, line 55: same with __le__

Magus6784 (changes requested):

PR #56, opened 2022-05-19
'Solution'

daria-niklaevna:


NOSSUDAK (approved):
It's good. But I don't think you need to use "not"  in line 40 and line 43

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, int):
```
TPolina, line 12: `other` can be also of type float. Consider this throughout your code

TPolina (changes requested):

TPolina (approved):
Good!

PR #57, opened 2022-05-19
Solution

Evgeniy994:


Hectorovich (approved):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
TPolina, line 48: optimize this method, use magic method you wrote: `return not self > other`

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
TPolina, line 53: same here

TPolina (changes requested):

TPolina (changes requested):
see private message

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        return not self.__gt__(other)
```
TPolina, line 46: magic methods are used in a different way.
Example: `a + b` -> method `__add__` will be called

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __ge__(self, other):
+        return not self.__lt__(other)
```
TPolina, line 49: same here

TPolina (changes requested):

MasterpieceElbow (approved):

PR #58, opened 2022-05-19
Solution

vasylhnatiuk:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km != other
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if type(other) is int:
```
katevasilenko, line 19: it could be float also. Just check if it not belong to Distance class

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+            return Distance(self.km + other)
+        else:
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if type(other) is int:
```
katevasilenko, line 25: The same

katevasilenko (changes requested):

app/main.py
vasylhnatiuk: thanks)

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __gt__(self, other):
+        return self.km > other
```
TPolina, line 38: `other` can be also of type `Distance`, then your code will not work. Add necessary checks. Fix this issue throughout your code

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 44: here you can use magic method you wrote (don't forget to use it as a magic method)
ge == not lt

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __le__(self, other):
+        return self.km <= other
```
TPolina, line 47: same as in __ge__ method: le == not gt

TPolina (changes requested):

TPolina (approved):

PR #59, opened 2022-05-19
'Solution'

YuraBeseda:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
-    # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        else:
+            return Distance(km=self.km + other)
```
Magus6784, line 9: You can have TypeError if you take str
So add elif isinstance(other, int):

Magus6784 (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=round(self.km / other.km, 2))
+        elif isinstance(other, (int, float)):
+            return Distance(km=round(self.km / other, 2))
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(km=self.km * other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(km=self.km * other)
```
TPolina, line 30: `other` can't be an instance of `Distance`, only a number (float or int)

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __len__(self):
+        return self.km
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(km=round(self.km / other.km, 2))
+        elif isinstance(other, (int, float)):
+            return Distance(km=round(self.km / other, 2))
```
TPolina, line 24: same as in `__mul__` method

MasterpieceElbow (approved):

PR #60, opened 2022-05-19
Solution

Igor-Cegelnyk:


app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
Abnormaltype, line 48: other can be of type Distance, then your code will not work. Consider this in the following methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
```
Abnormaltype, line 45: other can be of type Distance, then your code will not work. Consider this in the following methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return Distance(round(self.km / other.km, 2))
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
```
Abnormaltype, line 42: other can be of type Distance, then your code will not work. Consider this in the following methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        if not isinstance(other, Distance):
+            return Distance(round(self.km / other, 2))
+        else:
+            return Distance(round(self.km / other.km, 2))
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
```
Abnormaltype, line 39: other can be of type Distance, then your code will not work. Consider this in the following methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if not isinstance(other, Distance):
+            return Distance(round(self.km / other, 2))
+        else:
+            return Distance(round(self.km / other.km, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
Abnormaltype, line 36: other can be of type Distance, then your code will not work. Consider this in the following methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self
+        else:
+            self.km += other.km
+            return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if not isinstance(other, Distance):
```
Abnormaltype, line 30: You need to explicitly specify that if it is int or float type

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return Distance(round(self.km / other, 2))
+        else:
```
Abnormaltype, line 32: You need to explicitly specify that if it is `Distance`

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if not isinstance(other, Distance):
+            return Distance(self.km + other)
+        else:
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if not isinstance(other, Distance):
```
Abnormaltype, line 19: same here

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            self.km += other
+            return self
+        else:
```
Abnormaltype, line 22: same here

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers"
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if not isinstance(other, Distance):
+            return Distance(self.km + other)
+        else:
```
Abnormaltype, line 15: same here

Abnormaltype (changes requested):

app/main.py
Igor-Cegelnyk: Привіт! До кінця не зрозуміло, що не так з методом, він працює як з типом Distance так і з числами. Чи потрібно вказати анотацію типу, а у методі вказати умову для чисел?

Abnormaltype: You compare simply  `other`, and the problem is that if  `other` is a class `Distance`, compare `self.km == other.km`
And if the `other` is a type of `int` or type `float`, then make such a comparison`self.km == other`

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+            return self.km > other
+        else:
+            return self.km > other.km
+
+    def __eq__(self, other):
+        if not isinstance(other, Distance):
+            other == int or other == float
+            return self.km == other
+        else:
+            return self.km == other.km
```
Abnormaltype, line 61: It's not correct, better to write like this:
```
            return self.km == other.km
            return self.km == other
```

Abnormaltype: Implements this in other methods

Abnormaltype (changes requested):

app/main.py
Igor-Cegelnyk: Got it, thanks!

Abnormaltype (approved):
Good job!

Abnormaltype (approved):

PR #61, opened 2022-05-20
Solution

kamilla-boiko:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return True if self.km == other.km else False
+
+    def __le__(self, other):
+        if not isinstance(other, Distance):
+            return True if self.km <= other else False
+        return True if self.km <= other.km else False
+
+    def __ge__(self, other):
+        if not isinstance(other, Distance):
+            return True if self.km >= other else False
```
Abnormaltype, line 51: `other` can be only of type `float`, `int`, `Distance`. Consider this throughout your code

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        if isinstance(other, (float, int)):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, (float, int)):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
```
Abnormaltype, line 21: same here

Abnormaltype (changes requested):

app/main.py
kamilla-boiko: When I changed if-statement in __iadd__ and __add__ , it returned me "TypeError: unsupported operand type(s) for +: 'int' and 'Distance'", so I don't understand, how I can fix it.

Abnormaltype: Try to write like this:

```
            self.km += other.km
            return self
            self.km += other
            return self
```

Abnormaltype (approved):
Well done!

Abnormaltype (approved):

PR #62, opened 2022-05-24
Solution

mariia-movchan:


rotsen18 (changes requested):
Looking good, only run all tests

rotsen18 (approved):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        if isinstance(other, int):
```
TPolina, line 25: `other` can be also the type of `float`. Fix this throughout your code

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, int):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other):
+        if isinstance(other, int):
+            return self.km <= other
+        return self.km <= other.km
```
TPolina, line 42: here you can use magic method you already wrote (don't forget to use it as a magic method)
le == not gt

PR #63, opened 2022-05-26
Solution

LaskoA:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
+            total = self.km * other
+        else:
+            total = self.km * other.km
```
IvanRamyk, line 29: Distance shouldn't be multiplied by other distance
Please, remove this logic

IvanRamyk (changes requested):

LaskoA:
Please check now after __mul__ method was fixed

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        total = round(self.km / other, 2)
+        return Distance(total)
+
+    def __lt__(self, other):
+        if isinstance(other, (int, float)):
+            return self.km < other
+        else:
```
IvanRamyk, line 35: Redundant `else`
Please fix it in other places too.

IvanRamyk (changes requested):

LaskoA:
Else parts were deleted from lt, le, ge... Please check again :)

PR #70, opened 2022-06-08
Solution

artemkazakov947:
Please, check it.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -13,49 +13,49 @@ class Distance:
         return f"Distance(km={self.km})"
 
     def __add__(self, other):
-        if isinstance(other, int):
+        if isinstance(other, (int, float)):
             return Distance(self.km + other)
         return Distance(self.km + other.km)
 
     def __iadd__(self, other):
-        if isinstance(other, int):
+        if isinstance(other, (int, float)):
             self.km += other
             return self
         self.km += other.km
         return self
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, int):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, int):
```
IvanRamyk, line 21:
```suggestion
        if isinstance(other, (int, float)):
```
Distance class should support both numerical types

IvanRamyk (changes requested):

artemkazakov947 (commented):
float was added to isinstat()

app/main.py
```diff
@@ -1,3 +1,63 @@
+from __future__ import division
```
IvanRamyk, line 1: Please, don't leave unused imports in the code

IvanRamyk (changes requested):

app/main.py
artemkazakov947: Ooops

IvanRamyk (approved):
Good job!

IvanRamyk (approved):

PR #83, opened 2022-06-25
Solution

InnaKushnir:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
+            raise TypeError("both variants ^ are possible")
```
Abnormaltype, line 17: this statement never work out

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __len__(self):
+        counter = 0
+        for i in range(self.km + 1):
+            counter += i
+        return self.km
```
Abnormaltype, line 34: what about function `len()`?

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __gt__(self, other):
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
```
Abnormaltype, line 51: if `other` it's `int`? `int` object has atribute `km`?

Abnormaltype: please fix other cases

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
Abnormaltype, line 15: remove unnecessary `else`

Abnormaltype (changes requested):

app/main.py
InnaKushnir: int object has not atribute km, but I don"t know how to correct errors here and other similar cases

Abnormaltype (approved):
good job!

PR #87, opened 2022-07-04
solution

tarkaiev:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return (Distance(self.km + other)
+                if type(other) == int
```
IvanRamyk, line 13: It's better to check type using isinstance

IvanRamyk: Also, consider float type

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #98, opened 2022-07-21
Solution

lksndr307:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __lt__(self, other):
+        if type(other) == int:
+            return self.km < other
+        else:
+            return self.km < other.km
+
+    def __ge__(self, other):
+        if type(other) == int:
+            return self.km >= other
+        else:
```
IvanRamyk, line 58: Please, remove redundant else operators from all methods

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __iadd__(self, other):
+        if type(other) == int:
+            self.km += other
+            return self
+        self.km += other.km
+        return self
+
+    def __mul__(self, other):
+        if type(other) == int:
```
IvanRamyk, line 24: Use `isinstance` to check the variable type. It's important in the case of inheritance that will be discussed in detail on the OOP topic

IvanRamyk: Consider also `float` type

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

IvanRamyk (approved):
Good job!

DmytroSvirsa (approved):

PR #99, opened 2022-07-21
'Solution'

Viktor-Beniukh:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
+
+    def __len__(self):
+        if isinstance(self.km, int):
```
IvanRamyk, line 48: And if it's a float value?

IvanRamyk: I believe you don't need this if-statement

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers"
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, int):
```
Abnormaltype, line 13: And if it's a float value?

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, int):
```
Abnormaltype, line 18: the same here

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            self.km += other
+            return self
+        self.km += other.km
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, int):
```
Abnormaltype, line 25: the same here

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, int):
```
Abnormaltype, line 29: the same here

Abnormaltype (changes requested):

PR #108, opened 2022-07-31
Solution

vladimir-byshevets:


app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __iadd__(self, distance):
+        if (isinstance(distance, int)):
+            self.km += distance
+        else:
+            self.km += distance.km
+        return self
+
+    def __mul__(self, distance):
+        if (isinstance(distance, int)):
```
IvanRamyk, line 27:
```suggestion
        if isinstance(distance, int):
```
stick to the PEP8 standard

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, distance):
+        if (isinstance(distance, int)):
+            self.km = self.km + distance
```
IvanRamyk, line 14: You shouldn't modify the current instance. Only return a new one

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, distance):
+        if (isinstance(distance, int)):
```
IvanRamyk, line 13: Consider float type

IvanRamyk (changes requested):

app/main.py
vladimir-byshevets: You can clarify where inemmono needs to be applied float type?

vladimir-byshevets: You can clarify where inemmono needs to be applied float type?

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __add__(self, distance):
+        if isinstance(distance, int) or isinstance(distance, float):
+            dist = self.km + distance
+        else:
+            dist = self.km + distance.km
+        return Distance(dist)
+
+    def __iadd__(self, distance):
+        if isinstance(distance, int):
```
IvanRamyk, line 20: Consider float type in all methods :)

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, distance):
+        if isinstance(distance, int) or isinstance(distance, float):
```
IvanRamyk, line 13:
```suggestion
        if isinstance(distance, (int, float)):
```
Could be simplified

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+        else:
+            dist = distance.km
+        return self.km > dist
+
+    def __eq__(self, distance):
+        if isinstance(distance, int):
+            dist = distance
+        else:
+            dist = distance.km
+        return self.km is dist
```
IvanRamyk, line 59:
```suggestion
        return self.km == dist
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

vladimir-byshevets:
Спасибо

IvanRamyk (approved):

PR #109, opened 2022-08-01
Added tests for float type

Nattalli:
Added test cases with float type numbers

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -5,12 +5,15 @@ from contextlib import redirect_stdout
 
 from app.main import Distance
 
+
 @pytest.mark.parametrize(
     'kilometers',
     [
         50,
         100,
-        300
+        300,
+        12.5,
+        0.6
     ]
 )
 def test_distance_class_init(kilometers):
@@ -26,7 +29,9 @@ def test_distance_class_init(kilometers):
     [
         (50, "Distance: 50 kilometers.\n"),
         (100, "Distance: 100 kilometers.\n"),
-        (300, "Distance: 300 kilometers.\n")
+        (300, "Distance: 300 kilometers.\n"),
+        (12.5, "Distance: 12.5 kilometers.\n"),
+        (0.6, "Distance: 0.6 kilometers.\n")
     ]
 )
 def test_distance_class_str(kilometers, output):
@@ -49,7 +54,9 @@ def test_distance_class_str(kilometers, output):
     [
         (50, "Distance(km=50)"),
         (100, "Distance(km=100)"),
-        (300, "Distance(km=300)")
+        (300, "Distance(km=300)"),
+        (12.5, "Distance(km=12.5)"),
+        (.6, "Distance(km=0.6)")
     ]
 )
 def test_distance_class_str(kilometers, output):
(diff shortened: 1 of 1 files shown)


tests/test_main.py
```diff
@@ -49,7 +54,9 @@ def test_distance_class_str(kilometers, output):
     [
         (50, "Distance(km=50)"),
         (100, "Distance(km=100)"),
-        (300, "Distance(km=300)")
+        (300, "Distance(km=300)"),
+        (12.5, "Distance(km=12.5)"),
+        (.6, "Distance(km=0.6)")
```
Abnormaltype, line 59:
```suggestion
        (0.6, "Distance(km=0.6)")
```

tests/test_main.py
```diff
@@ -363,11 +410,18 @@ def test_distance_class_le_number(kilometers, kilometers2, result):
...
-    assert len(distance) == distance.km, (
-        f"'len()' for instance should return instance.km"
-    )
+    if isinstance(kilometers, int):
+        assert len(distance) == distance.km, (
+            f"'len()' for instance should return instance.km"
+        )
+    else:
+        assert kilometers == distance.km, (
+            f"'len()' for instance of float type should return instance.km"
```
danylott, line 426: I believe this comment is not really correct.
`len` for instance of float type will be an error. Smth like `len(2.3)` -> error.
I believe the good message is `"'len()' for instance of Distance should return instance.km"`

tests/test_main.py
```diff
@@ -363,11 +410,18 @@ def test_distance_class_le_number(kilometers, kilometers2, result):
...
     distance = Distance(kilometers)
-    assert len(distance) == distance.km, (
-        f"'len()' for instance should return instance.km"
-    )
+    if isinstance(kilometers, int):
+        assert len(distance) == distance.km, (
+            f"'len()' for instance should return instance.km"
+        )
+    else:
+        assert kilometers == distance.km, (
```
danylott, line 425: In this test we check only `len` functionality. No this step needed

danylott (changes requested):

tests/test_main.py
```diff
@@ -363,11 +410,18 @@ def test_distance_class_le_number(kilometers, kilometers2, result):
...
+            f"'len()' for instance of float type should return instance.km"
+        )
```
Abnormaltype, line 427: assert len(distance) == distance.km, (
        f"'len()' for instance should return instance.km"
    )

Abnormaltype (changes requested):

tests/test_main.py
```diff
@@ -363,11 +410,18 @@ def test_distance_class_le_number(kilometers, kilometers2, result):
...
 def test_distance_class_len(kilometers):
     distance = Distance(kilometers)
-    assert len(distance) == distance.km, (
-        f"'len()' for instance should return instance.km"
-    )
+    if isinstance(kilometers, int):
+        assert len(distance) == distance.km, (
+            f"'len()' for instance should return instance.km"
+        )
+    else:
```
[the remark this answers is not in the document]
Nattalli: > It seems that you don't check `len` method for float distance now

But method 'len' in python does not work for float type. It is return such error: 
"TypeError: object of type 'float' has no len()"

tests/test_main.py
Abnormaltype: this one

Abnormaltype (changes requested):

IvanRamyk (approved):
Good job!

tests/test_main.py
Abnormaltype, line 59:
```suggestion
        (0.6, "Distance(km=0.6)")
```

Abnormaltype (changes requested):
🥲

PR #121, opened 2022-08-14
Solution

volodymyr-vereshchak:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
DmytroSvirsa, line 30: The `other` can be not only `(int, float)` but also a `Distance` class object

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __gt__(self, other):
+        return self.km > other
```
DmytroSvirsa, line 33: The `other` can be not only `(int, float)` but also a `Distance` class object

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __eq__(self, other):
+        return self.km == other
```
DmytroSvirsa, line 36: The `other` can be not only `(int, float)` but also a `Distance` class object

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __le__(self, other):
+        return self.km <= other
```
DmytroSvirsa, line 39: The `other` can be not only `(int, float)` but also a `Distance` class object

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other
```
DmytroSvirsa, line 42: The `other` can be not only `(int, float)` but also a `Distance` class object

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __add__(self, other):
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, (int, float)):
+            self.km += other
+        else:
```
DmytroSvirsa, line 19: You don’t need `else` if you have a `return` statement in `if`. Rewrite without `else`

DmytroSvirsa (changes requested):

volodymyr-vereshchak:
Fix all. Why using else is considered bad practice?

IvanRamyk (approved):
Well done!

DmytroSvirsa (approved):
You just don't need `else` when having a `return` statement, it does not do any function, so you have to remove it.

PR #131, opened 2022-08-31
Develop

yessloss:


app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=other.km + self.km)
+        return Distance(km=other + self.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, int) or isinstance(other, float):
+            x = other
+            self.km += x
+        else:
+            x = other.km
```
IvanRamyk, line 21: Using an extra variable here

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
```
IvanRamyk, line 35: Consider that `other` could have different types (including `Distance`, `float`, `int`)

IvanRamyk (changes requested):

app/main.py
yessloss: but tests is done?

IvanRamyk (approved):
Well done!

PR #132, opened 2022-08-31
py_distance_class_Solution

vadim-kosnovskiy:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
+
+    def __le__(self, other):
```
IvanRamyk, line 36: Consider that `other` could be a variable of the different types, such as `int`, `float`, `Distance`, etc

IvanRamyk (changes requested):

app/main.py
vadim-kosnovskiy: Добавил проверку типа.

DmytroSvirsa (approved):

PR #136, opened 2022-09-04
Solution for magic functions.

kkatash:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
DmytroSvirsa, line 37: why don't you check instance there

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __gt__(self, other):
+        return self.km > other
```
DmytroSvirsa, line 40: why don't you check instance there

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __eq__(self, other):
+        return self.km == other
```
DmytroSvirsa, line 43: why don't you check instance there

DmytroSvirsa (changes requested):

app/main.py
kkatash: Added isinstance check for magic methods

app/main.py
```diff
@@ -34,13 +34,22 @@ def __truediv__(self, other):
             return Distance(round(self.km / other, 2))
 
     def __lt__(self, other):
-        return self.km < other
+        if isinstance(other, Distance):
+            return self.km < other.km
+        if isinstance(other, (int, float)):
+            return self.km < other
```
DmytroSvirsa, line 40: You may be sure that you always get one of `int`, `float`, `Distance`, so you can rewrite without a double check (not only here, check for this full file)

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #140, opened 2022-09-06
Solution

OleksiiKarpov:


app/main.py
```diff
@@ -1,3 +1,99 @@
...
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        if isinstance(other, Distance):
+            if self.km < other.km:
+                return True
+            else:
```
DmytroSvirsa, line 42: rewrite without else, the same do at: $47, $59, $66, $71, $78, $90, $95

DmytroSvirsa (changes requested):

PR #141, opened 2022-09-06
Solution

vlad-milkov:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __truediv__(self, other):
+        return Distance(
+            round(self.km / other, 2)
+        )
+
+    def __lt__(self, other):
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
```
DmytroSvirsa, line 40: use existing magick method `ge` to wright less code

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __gt__(self, other):
```
DmytroSvirsa, line 42: use existing magick method `le` to wright less code

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #152, opened 2022-09-13
'Solution'

AVyha:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if type(other) in (int, float):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
```
Nattalli, line 16: But usually `iadd` is `+=`. So it will not change your code working, but better replace `self.km = self.km + other` with `self.km += other` and so on

Nattalli (changes requested):

Nattalli (approved):
Well done!

PR #159, opened 2022-09-16
Solution

kateryna3k:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
```
TPolina, line 26: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round((self.km / other.km), 2))
+        return Distance(round((self.km / other), 2))
```
TPolina, line 31: same as in mul method

TPolina (changes requested):

app/main.py
```diff
@@ -21,38 +21,24 @@ def __iadd__(self, other):
...
 
     def __gt__(self, other):
-        if isinstance(other, Distance):
-            return self.km > other.km
         return self.km > other
 
     def __eq__(self, other):
-        if isinstance(other, Distance):
-            return self.km == other.km
         return self.km == other
```
TPolina, line 36: you spoiled the correct implementation :( You still need to do a type check here

app/main.py
```diff
@@ -21,38 +21,24 @@ def __iadd__(self, other):
...
 
     def __le__(self, other):
-        if isinstance(other, Distance):
-            return self.km <= other.km
         return self.km <= other
 
     def __ge__(self, other):
-        if isinstance(other, Distance):
-            return self.km >= other.km
         return self.km >= other
```
TPolina, line 42: and here you should use the previously created magic methods. Look: `<=` is the same as `not >`

If it's hard to understand tag me on your slack channel or visit qna)

TPolina (changes requested):

TPolina (approved):
Good job)

PR #160, opened 2022-09-16
Solution)

Gr1zzZzzly:


app/main.py
```diff
@@ -1,3 +1,43 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return(Distance(km=self.km + other.km)
+               if isinstance(other, Distance)
+               else Distance(km=self.km + other))
+
+    def __iadd__(self, othe):
```
mltdll, line 17: A typo in word "other". :)

mltdll (changes requested):
There's a repeating pattern in your code that violates the DRY principle:
```python
(other.km if isinstance(other, Distance) else other)
```
I'd strongly suggest to extract it to a new static method.

mltdll (approved):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km == self.obj_or_int(other)
+
+    def __le__(self, other):
+        return self.km <= self.obj_or_int(other)
+
+    def __ge__(self, other):
+        return self.km >= self.obj_or_int(other)
+
+    def __len__(self):
+        return range(self.km)
```
TPolina, line 46: why range?

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return(Distance(km=self.km + other.km)
+               if isinstance(other, Distance)
+               else Distance(km=self.km + other))
```
TPolina, line 15: you can use `obj_or_int` here as well

TPolina (changes requested):

TPolina (approved):
Well done!

PR #161, opened 2022-09-17
Solution

taras-andruschenko:


app/main.py
```diff
@@ -1,3 +1,48 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {str(self.km)} kilometers."
```
TPolina, line 7: `str` method is unnecessary here

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other):
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        self.km += other.km
+        return self
```
TPolina, line 22: `return self` is duplicated

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        return Distance(self.km / other.km)
```
TPolina, line 30: `other` can not be of type `Distance` in this method

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __lt__(self, other):
```
TPolina, line 32: What if `other` is of type `Distance` here?

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km < other
+
+    def __gt__(self, other):
```
TPolina, line 35: same here with `other` type

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
```
TPolina, line 38: same here

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 45: same here and also you should use previously created magic method here to simplify your code

TPolina (changes requested):

TPolina (changes requested):

TPolina (approved):
Well done)

PR #162, opened 2022-09-17
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other):
+        return self.km == other
```
taras-andruschenko, line 31: What if other is of type Distance here?

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __gt__(self, other):
+        return self.km > other
```
taras-andruschenko, line 34: same here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __ge__(self, other):
+        return self.km >= other
```
taras-andruschenko, line 37: same here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __lt__(self, other):
+        return self.km < other
```
taras-andruschenko, line 40: same here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __le__(self, other):
+        return self.km <= other
```
taras-andruschenko, line 43: same here

taras-andruschenko (changes requested):

app/main.py
olia-trofymchuk: It will compare two objects

taras-andruschenko: But if type(other) is Distance, you should check it and compare self.km with other.km, isn't it?

olia-trofymchuk: Yes, thank you for your comment!

taras-andruschenko (approved):
Good job)

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
Y-Havryliv, line 11: add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        else:
+            return self.km == other.km
+
+    def __gt__(self, other):
+        if isinstance(other, (int, float)):
+            return self.km > other
+        else:
+            return self.km > other.km
+
+    def __ge__(self, other):
```
Y-Havryliv, line 42: you can write `__le__` and `__ge__` in a one line using previously created magic methods

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #163, opened 2022-09-18
Solution

Aleksandro777:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
skrypka-dmytro, line 50: forgot if isinstance(other, Distance):

skrypka-dmytro (changes requested):

skrypka-dmytro (approved):

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
```
TPolina, line 28: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+
+        else:
+            return Distance(round(self.km / other, 2))
```
TPolina, line 35: same as in mul method

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
TPolina, line 11: please, add type annotation everywhere

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
TPolina, line 65: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: `<=` == `not >`

TPolina (changes requested):

TPolina (approved):
Good job!

PR #164, opened 2022-09-18
'3rSolution'

ketstap162:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 12: I think you should use isinstance() here, as in every method where you compare types. In any case, it will look more beautiful

AriohBelskij (changes requested):

app/main.py
```diff
@@ -15,7 +15,7 @@ def __add__(self, other):
             return Distance(self.km + other)
 
     def __iadd__(self, other):
-        if type(other) == type(self):
+        if isinstance(other, type(self)):
```
AriohBelskij, line 18: u need use isinstance in all ur methods! and u dont need else.

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, type(self)):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 25: use isinstance() and remove else:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 31: use isinstance() and remove else:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return Distance(round(self.km / other.km, 2))
+        else:
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 37: use isinstance() and remove else:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 43: use isinstance() and remove else:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return self.km > other.km
+        else:
+            return self.km > other
+
+    def __eq__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 49: use isinstance() and remove else:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 55: use isinstance() and remove else:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other):
+        if type(other) == type(self):
```
AriohBelskij, line 61: use isinstance() and remove else:

AriohBelskij (changes requested):
fix it

ketstap162:
> fix it

I fixed it.
What about else... I think it should be.

AriohBelskij:
> > fix it
> 
> I fixed it. What about else... I think it should be.

return ends the function. u can write like a - 
if isinstance(666, int) :
       return some what u want if isinstance()
return some what u want if not isinstance()

U dont need else. Fix it

app/main.py
```diff
@@ -9,7 +9,7 @@ def __repr__(self):
         return f"Distance(km={self.km})"
 
     def __add__(self, other):
-        if type(other) == type(self):
+        if isinstance(other, type(self)):
             return Distance(self.km + other.km)
         else:
```
AriohBelskij, line 14: remove else)

AriohBelskij (changes requested):

app/main.py
```diff
@@ -11,8 +11,7 @@ def __repr__(self):
     def __add__(self, other):
         if isinstance(other, type(self)):
```
AriohBelskij, line 12: explicitly show the type, now type(self) looks very confusing, specify the class like - isinstance(other, Distance)
change it in all ur methods ;D

AriohBelskij (changes requested):

app/main.py
AriohBelskij: and fix flake8

PR #166, opened 2022-09-19
Solution <3

Mowka7:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
TPolina, line 11: please, add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
```
TPolina, line 26: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
TPolina, line 54: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: `<=` == `not >`

TPolina (changes requested):

app/main.py
```diff
@@ -1,57 +1,58 @@
...
+    def __le__(self, other) -> bool:
         if isinstance(other, Distance):
             return self.km <= other.km
         return self.km <= other
 
-    def __ge__(self, other):
+    def __ge__(self, other) -> bool:
         if isinstance(other, Distance):
             return self.km >= other.km
         return self.km >= other
```
TPolina, line 55: you skipped my previous comment:
> try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= == not >

tag me on your slack channel if you don't understand what should you do

TPolina (changes requested):

app/main.py
Mowka7: I didn't notice, sorry. I've already changed it

app/main.py
```diff
@@ -45,14 +45,10 @@ def __eq__(self, other) -> bool:
         return self.km == other
 
     def __le__(self, other) -> bool:
-        if isinstance(other, Distance):
-            return self.km <= other.km
-        return self.km <= other
+        return not self.__gt__(other)
```
TPolina, line 48: great, but we call magic methods with special interfaces (symbols for example)
`+` will call `__add__` method

app/main.py
```diff
@@ -45,14 +45,10 @@ def __eq__(self, other) -> bool:
...
 
     def __ge__(self, other) -> bool:
-        if isinstance(other, Distance):
-            return self.km >= other.km
-        return self.km >= other
+        return not self.__lt__(other)
```
TPolina, line 51: use symbol here as well

TPolina (changes requested):

TPolina (approved):

PR #168, opened 2022-09-19
Solution

matechris:


Mowka7 (changes requested):

Mowka7 (approved):
well done(;

app/main.py
```diff
@@ -1,3 +1,68 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float):
```
Y-Havryliv, line 2: can it be only float?

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        self.km = km
+
+    def __str__(self):
```
Y-Havryliv, line 5: add type annotations everywhere, and for return values as well

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __iadd__(self, other):
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+
+        return self
+
+    def __mul__(self, other):
```
Y-Havryliv, line 25: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+        return Distance(self.km * other.km)
+
+    def __truediv__(self, other):
```
Y-Havryliv, line 31: same here as in mul method

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+        return self.km > other.km
+
+    def __eq__(self, other):
+        if not isinstance(other, Distance):
+            return self.km == other
+
+        return self.km == other.km
+
+    def __le__(self, other):
```

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #169, opened 2022-09-19
Solution

Kitazuka:


Yaroslav-Stetsko (approved):
Looks good to me

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
TPolina, line 30: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 42: it would be great to use previously created magic methods here

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, (int, float)):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other):
+        return __class__(self.km * other)
```
TPolina, line 24: What for do you use `__class__` here?

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __truediv__(self, other):
+        return __class__(round(self.km / other, 2))
```
TPolina, line 27: Why `__class__`?

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other):
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other):
+        return not self.km > other
```
TPolina, line 45: you need to compare `self` with `other`, not `self.km` with `other`. Magic methods accept `self` (Distance instance) not int/float

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __ge__(self, other):
+        return not self.km < other
```
TPolina, line 48: same here

TPolina (changes requested):

TPolina (approved):

PR #170, opened 2022-09-19
SKantor Solution

KantorSerhiy:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __gt__(self, other):
```
IvanRamyk, line 34: Consider that other can be a variable of different types (`int`, `float`, `Distance`)

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __truediv__(self, other):
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other):
+        if isinstance(other, (Distance, int, float)):
+            return self.km < other
+
+    def __gt__(self, other):
+        if isinstance(other, (Distance, int, float)):
+            return self.km > other
```
IvanRamyk, line 37: if `other` is a `Distance` instance then the condition should be different:
```suggestion
            return self.km > other.km
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

IvanRamyk (approved):
Well done!

PR #173, opened 2022-09-19
Develop

ketstap162:


skrypka-dmytro (approved):

app/main.py
```diff
@@ -1,3 +1,59 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int):
```

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 24: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 29: same here as in mul method

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other) -> Distance:
```
Y-Havryliv, line 14: `other` should be annotated as well everywhere

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
Y-Havryliv, line 48: it's a magic method, just use `not self > other`

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __ge__(self, other) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
Y-Havryliv, line 51: same here

Y-Havryliv (changes requested):

app/main.py
ketstap162: How to do annotation `other: int or float orDistance` ?

app/main.py
```diff
@@ -1,3 +1,54 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Y-Havryliv, line 5: `km` can be `int` as well

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, Distance)) -> Distance:
```
Y-Havryliv, line 14: please, use these annotations: `int | float | Distance` or `Union[int, float, Distance]`

Y-Havryliv (changes requested):

PR #174, opened 2022-09-19
Solution

HStakhiv:


Yaroslav-Stetsko (approved):

Yaroslav-Stetsko (approved):
Looks good to me

app/main.py
```diff
@@ -1,3 +1,44 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float):
```

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km <= self.instancecheck(other)
+
+    def __ge__(self, other):
+        return self.km >= self.instancecheck(other)
+
+    def __len__(self):
+        return self.km
+
+    @staticmethod
+    def instancecheck(other) -> int:
```
Y-Havryliv, line 43: gj, but it would be better to name this method `instance_check`, it would be more readable

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+    def __lt__(self, other):
+        return self.km < self.instancecheck(other)
+
+    def __gt__(self, other):
+        return self.km > self.instancecheck(other)
+
+    def __eq__(self, other):
+        return self.km == self.instancecheck(other)
+
+    def __le__(self, other):
```
Y-Havryliv, line 33: you can write `__le__` and `__ge__` in a one line using previously created magic methods, so you won't need `instancecheck`

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return Distance(self.km + self.instancecheck(other))
+
+    def __iadd__(self, other):
+        self.km += self.instancecheck(other)
+        return self
+
+    def __mul__(self, other):
```
Y-Havryliv, line 18: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return Distance(self.km * self.instancecheck(other))
+
+    def __truediv__(self, other):
```
Y-Havryliv, line 21: same here as in mul method

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km < self.instance_check(other)
+
+    def __gt__(self, other) -> bool:
+        return self.km > self.instance_check(other)
+
+    def __eq__(self, other) -> bool:
+        return self.km == self.instance_check(other)
+
+    def __le__(self, other) -> bool:
+        return not self.__gt__(other)
```
Y-Havryliv, line 34: it's a magic method, use `self > other` to call it

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
```
Y-Havryliv, line 37: same here, don't use `self.__lt__(other)`

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
Y-Havryliv, line 11: and annotations still missing

Y-Havryliv (changes requested):

Y-Havryliv:
almost good

HStakhiv:
I think it's good now

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other) -> Distance:
```
Y-Havryliv, line 14: it's almost good, but `other` should be annotated everywhere as well

Y-Havryliv (changes requested):

app/main.py
HStakhiv: Sorry

HStakhiv:
Sorry. Can you check one more time?

Y-Havryliv (approved):
Well done!

PR #175, opened 2022-09-19
Solution

KatrinLazarenko:


matechris (approved):
Good job :)

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return Distance(other + self.km)
+
+    def __iadd__(self, other):
```
TPolina, line 14: please, add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        return Distance(other + self.km)
```
TPolina, line 12: What if `other` is of type `Distance` here?

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
TPolina, line 28: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 40: and it would be great if you use the previously created magic methods in these two methods to simplify your code

TPolina (changes requested):

app/main.py
```diff
@@ -1,43 +1,51 @@
...
-        return self.km == other
+    def __eq__(self, other) -> bool:
+        if isinstance(other, (int, float)) is True:
+            return self.km == other
+        return self.km == other.km
 
-    def __le__(self, other):
-        return self.km <= other
+    def __le__(self, other) -> bool:
+        return True if self.__gt__(other) is False else False
```
TPolina, line 45: You should use magic methods with a special interface. For example, `+` will call `__add__` method
Also, you do an unnecessary job. Your ternary operator is equal to `not self > other`. It's much more readable

app/main.py
```diff
@@ -1,43 +1,51 @@
...
 
-    def __ge__(self, other):
-        return self.km >= other
+    def __ge__(self, other) -> bool:
+        return True if self.__lt__(other) is False else False
```
TPolina, line 48: same things as in above comment

TPolina (changes requested):

app/main.py
```diff
@@ -42,10 +42,10 @@ def __eq__(self, other) -> bool:
         return self.km == other.km
 
     def __le__(self, other) -> bool:
-        return True if self.__gt__(other) is False else False
+        return not self.__gt__(other)
```
TPolina, line 45: wrong usage of magic method again
`not self > other`

app/main.py
```diff
@@ -42,10 +42,10 @@ def __eq__(self, other) -> bool:
...
 
     def __ge__(self, other) -> bool:
-        return True if self.__lt__(other) is False else False
+        return not self.__lt__(other)
```
TPolina, line 48: same

TPolina (changes requested):

TPolina (approved):

PR #176, opened 2022-09-19
Solution

Yaroslav-Stetsko:


HStakhiv (approved):
Nice!

app/main.py
```diff
@@ -1,3 +1,68 @@
 class Distance:
     # Write your code here
-    pass
+    def __init__(self, km: (int, float)):
+        self.km = km
+
+    def __add__(self, other):
```
Y-Havryliv, line 6: add type annotations everywhere, and for return values as well

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
Y-Havryliv, line 9: it would be better to use only `else` here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return Distance(self.km + other)
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 19: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
Y-Havryliv, line 27: it would be better to use only `else` here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+        elif isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
```
Y-Havryliv, line 40: it would be better to use only `else` here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return self.km == other
+
+    def __gt__(self, other):
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
```
Y-Havryliv, line 46: it would be better to use only `else` here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return self.km > other
+
+    def __lt__(self, other):
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
```
Y-Havryliv, line 52: it would be better to use only `else` here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return self.km < other
+
+    def __ge__(self, other):
```

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
Yaroslav-Stetsko: Done

app/main.py
```diff
@@ -1,3 +1,54 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (int, float)) -> None:
```
Y-Havryliv, line 5: gj, but use annotations like this: `int | float` or `Union[int, float]`

Y-Havryliv (changes requested):

app/main.py
Yaroslav-Stetsko: fixed

Y-Havryliv (approved):
GJ!

PR #177, opened 2022-09-19
Solution

rostyak:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, number):
+        if isinstance(number, Distance):
+            return Distance(self.km + number.km)
+        elif isinstance(number, (int, float)):
```
wQuelS, line 14: here and in all magic methods below u don't have to check wether second arg is int or float. Task was to ensure that second arg is Distance instance.

wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return self + Distance(number)
```
ketstap162, line 15: return self + Distance(number) - bad code
You building a new class and calling func __add__ again
Write:
return Distance(self.km + number)

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            self.km += number.km
+        elif isinstance(number, (int, float)):
+            self.km += number
+        return self
+
+    def __mul__(self, number):
+        if isinstance(number, Distance):
+            return Distance(self.km * number.km)
+        elif isinstance(number, (int, float)):
+            return self * Distance(number)
```
ketstap162, line 28: return self * Distance(number)
Same:
return Distance(self.km * number)

ketstap162 (changes requested):

wQuelS (approved):
Looks good, imo:)

ketstap162 (approved):
Nice

app/main.py
```diff
@@ -1,3 +1,59 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km):
```

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __iadd__(self, number):
+        if isinstance(number, Distance):
+            self.km += number.km
+        else:
+            self.km += number
+        return self
+
+    def __mul__(self, number):
+        if isinstance(number, Distance):
```
Y-Havryliv, line 24: `number` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km * number.km)
+        return Distance(self.km * number)
+
+    def __truediv__(self, number):
+        if isinstance(number, Distance):
```
Y-Havryliv, line 29: same here as in mul method

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return self.km > number.km
+        return self.km > number
+
+    def __eq__(self, number):
+        if isinstance(number, Distance):
+            return self.km == number.km
+        return self.km == number
+
+    def __le__(self, number):
```
Y-Havryliv, line 48: you can write `__le__` and `__ge__` in a one line using previously created magic methods

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #178, opened 2022-09-19
Solution

AKrekhovetskyi:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return True if self.km > other.km else False
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, (int, float)):
+            return True if self.km == other else False
+        return True if self.km == other.km else False
+
+    def __le__(self, other) -> bool:
+        if isinstance(other, (int, float)):
+            return True if self.km <= other else False
```
IvanRamyk, line 50:
```suggestion
            return self.km <= other
```
Such construction can be simplified since `self.km <= other` statement already returns a boolean value

IvanRamyk: please, fix all of them

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #179, opened 2022-09-20
'Solution'

hasslieb3:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
```
ketstap162, line 25: 24th row:
I think you should remove this enter.

ketstap162 (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __len__(self):
+        return self.km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
TPolina, line 14: please, add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
```
TPolina, line 29: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+        return Distance(round(self.km / other, 2))
```
TPolina, line 34: same as in mul method

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
TPolina, line 59: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: `<=` == `not >`

TPolina (changes requested):

app/main.py
```diff
@@ -24,36 +24,22 @@ def __iadd__(self, other):
...
 
     def __gt__(self, other):
-        if isinstance(other, Distance):
-            return self.km > other.km
         return self.km > other
 
     def __eq__(self, other):
-        if isinstance(other, Distance):
-            return self.km == other.km
         return self.km == other
```
TPolina, line 39: you spoiled the correct solution. Type checks are still required in these methods

app/main.py
```diff
@@ -24,36 +24,22 @@ def __iadd__(self, other):
...
 
     def __le__(self, other):
-        if isinstance(other, Distance):
-            return self.km <= other.km
         return self.km <= other
 
     def __ge__(self, other):
-        if isinstance(other, Distance):
-            return self.km >= other.km
         return self.km >= other
```
TPolina, line 45: and only these two methods should be rewritten with previously created magic methods:
`<=` is the same as `not >`

Tag me on your slack channel if it's hard to understand

TPolina (changes requested):

TPolina (approved):

PR #181, opened 2022-09-20
'Solution'

OvDima1:


app/main.py
```diff
@@ -1,3 +1,68 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int):
```
Y-Havryliv, line 3: add type annotations everywhere, and for return values as well

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 26: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        else:
+            return self.km < other
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other):
```

Y-Havryliv (changes requested):

app/main.py
ketstap162: > add type annotations everywhere, and for return values as well

def __add__(self, other) -> Distance:

NameError: name 'Distance' is not defined

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int):
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance:
```
Y-Havryliv, line 15: `other` can be not only `int`, fix it everywhere

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return self.km < other
+
+    def __eq__(self, other: int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: int) -> bool:
+        if isinstance(other, Distance):
```
Y-Havryliv, line 53: don't call magic methods like this, use `self > other` to call it, and you don't even need to check the `other` type

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return not self.__gt__(other.km)
+        else:
+            return not self.__gt__(other)
+
+    def __ge__(self, other: int) -> bool:
+        if isinstance(other, Distance):
```
Y-Havryliv, line 59: same here, don't use `self.__lt__(other.km)`

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int | float):
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
Y-Havryliv, line 15: `other` can be `Distance` as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float):
```
Y-Havryliv, line 21: `other` can be `Distance` as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __gt__(self, other: int | float) -> bool:
```
Y-Havryliv, line 34: `other` can be `Distance` as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
+
+    def __lt__(self, other: int | float) -> bool:
```
Y-Havryliv, line 40: `other` can be `Distance` as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __eq__(self, other: int | float) -> bool:
```
Y-Havryliv, line 46: `other` can be `Distance` as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
```
Y-Havryliv, line 52: `other` can be `Distance` as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return not self > other
+
+    def __ge__(self, other: int | float) -> bool:
```
Y-Havryliv, line 55: `other` can be `Distance` as well

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #182, opened 2022-09-20
Solution

Vasyl-Poremchuk:


ketstap162 (approved):
Well done!

app/main.py
```diff
@@ -1,3 +1,53 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (int, float)):
+        self.km = km
+
+    def __str__(self):
```
Y-Havryliv, line 5: add type annotations everywhere, and for return values as well

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 26: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+            return Distance(self.km * other.km)
+
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
```

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+            return Distance(round(self.km / other.km, 2))
+
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
```
Y-Havryliv, line 37: you need to check `other` type in `__lt__`, `__gt__` and `__eq__` as well

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
```
Y-Havryliv, line 46: you can write `__le__` and `__ge__` using previously created magic methods

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
+
+    def __len(self):
```
Y-Havryliv, line 52: fix this method

Y-Havryliv (changes requested):

app/main.py
Vasyl-Poremchuk: Everything is fixed)

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other) -> Distance:
```
Y-Havryliv, line 14: please, annotate `other` as well, and consider that it can be `int` or `float` or even `Distance` somewhere

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return self.km > other
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+        return self.km == other
+
+    def __le__(self, other) -> bool:
+        return not self.__gt__(other)
```
Y-Havryliv, line 53: it's a magic method, use `self > other` to call it

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
```
Y-Havryliv, line 56: same here, don't use `self.__lt__(other)`

app/main.py
```diff
@@ -1,3 +1,59 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (int, float)) -> None:
```
Y-Havryliv, line 5: and annotate using `int | float` or `Union[int, float]`

Y-Havryliv (changes requested):

app/main.py
Vasyl-Poremchuk: Done!)

Y-Havryliv (approved):
GJ!

PR #183, opened 2022-09-20
Develop

VitaliiPiaskivskyi:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
```
misha-doroshenko, line 26: Other's type can't be Distance, because Distance * Distance gives us square km. You should multiply Distance only  by number

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+        return Distance(round(self.km / other, 2))
```
misha-doroshenko, line 31: The same as __mul__

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
misha-doroshenko, line 51: You can initialize some comparison operators using others (e.g. (<=) the same as (not >))

misha-doroshenko (changes requested):

app/main.py
VitaliiPiaskivskyi: It is possible, the cost of computer resources will be the same. I would like to leave everything as it is

app/main.py
VitaliiPiaskivskyi: ок

app/main.py
VitaliiPiaskivskyi: ок

app/main.py
```diff
@@ -1,3 +1,52 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (int, float)):
```
misha-doroshenko, line 2: I'm not sure that it's correct to use (int, float) as type annotation in this case. I have always used Union[int, float].
Everything else is good. Well done!

misha-doroshenko (approved):

app/main.py
Y-Havryliv: agree, use `int | float` or `Union[int, float]`

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f'Distance: {self.km} kilometers.'
+
+    def __repr__(self) -> str:
+        return f'Distance(km={self.km})'
+
+    def __add__(self, other):
```
Y-Havryliv, line 11: add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other) -> bool:
```
Y-Havryliv, line 44: DRY, rewrite it using previously created magic method

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other) -> bool:
```
Y-Havryliv, line 49: DRY, rewrite it using previously created magic method

app/main.py
```diff
@@ -1,3 +1,55 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float):
```
Nattalli, line 2: you have to add that this function returns nothing: `-> None`

app/main.py
Nattalli: yes, please, add type annotation for:
1) `other` variable
2) function. If function returns nothing then `-> None`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __repr__(self) -> str:
+        return f'Distance(km={self.km})'
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other):
```
Nattalli, line 16: yes, please, add type annotation for:

1. `other` variable
2. function. If function returns nothing then `-> None`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other):
```
Nattalli, line 23: yes, please, add type annotation for:

1. `other` variable
2. function. If function returns nothing then `-> None`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
```
Nattalli, line 26: yes, please, add type annotation for:

1. `other` variable
2. function. If function returns nothing then `-> None`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other) -> bool:
```
Nattalli, line 29: add type annotation for `other` variable

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other) -> bool:
```
Nattalli, line 34: add type annotation for `other` variable

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other) -> bool:
```
Nattalli, line 39: add type annotation for `other` variable

Nattalli (changes requested):

PR #184, opened 2022-09-20
Solution

misha-doroshenko:


Yaroslav-Stetsko (approved):
Looks good to me

app/main.py
```diff
@@ -1,3 +1,59 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km):
```
TPolina, line 2: add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
TPolina, line 23: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __mul__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
```
TPolina, line 29: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+        else:
+            return Distance(round(self.km / other, 2))
```
TPolina, line 35: same here as in mul method

TPolina (changes requested):

TPolina (approved):
Good job!

PR #185, opened 2022-09-20
'Solution'

Ilyakson:


Yaroslav-Stetsko (approved):
Looks good to me

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return self + Distance(other)
```
TPolina, line 14: better return Distance(self.km + other)

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
TPolina, line 21: `return self` is duplicated

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
TPolina, line 52: you can write these two methods in one line each. Use previously created magic methods

TPolina (changes requested):

TPolina (approved):
Well done)

PR #188, opened 2022-09-20
solved the task

Cerne13:


YevheniiDavydenko (approved):
cool, better then mшту

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other):
+        return self.km < other
```
TPolina, line 29: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km > other
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
TPolina, line 41: it would be great if you use previously created magic methods in these two methods to simplify your code

TPolina (changes requested):

app/main.py
Cerne13: Thank you! Somehow it slipped out of my mind to create a separate static method for the ckecks.

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __add__(self, other):
+        return Distance(self.km + self.check_type(other))
+
+    def __iadd__(self, other):
+        self.km = self.__add__(other)
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * self.check_type(other))
```
TPolina, line 25: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __truediv__(self, other):
+        return Distance(round(self.km / self.check_type(other), 2))
```
TPolina, line 28: same as in `mul` method

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km < self.check_type(other)
+
+    def __gt__(self, other):
+        return self.km > self.check_type(other)
+
+    def __eq__(self, other):
+        return self.km == self.check_type(other)
+
+    def __le__(self, other):
+        return self.__lt__(other) or self.__eq__(other)
```
TPolina, line 40: use magic method with magic interface (symbols)

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __ge__(self, other):
+        return self.__gt__(other) or self.__eq__(other)
```
TPolina, line 43: same here

TPolina (changes requested):

app/main.py
```diff
@@ -2,43 +2,45 @@ class Distance:
...
-        if isinstance(other, Distance):
-            km = self.km + other.km
-        else:
-            km = self.km + other
-        return Distance(km)
+        return Distance(self.km + self.check_type(other))
 
     def __iadd__(self, other):
-        self.km = self.__add__(other).km
+        self.km = self.__add__(other)
```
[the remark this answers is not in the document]
Cerne13: Regarding object assignment - you've requested to make a type check earlier. 
So now we have a type check in __add__ method. If Distance object is passed, it will pass Distance.km

Regarding magic methods - understood, thank you.

TPolina (approved):

PR #189, opened 2022-09-20
Solution

linaSam:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if type(other) == int or type(other) == float:
```
mltdll, line 12: You should extract type checking into a separate methods that returns int | float, because right now your code violates DRY principle.
Also, it's easier to check for `isinstance(obj, Distance)` instead of int/float.

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return self.km < other
+        return self.km < other.km
+
+    def __gt__(self, other):
+        if type(other) == int or type(other) == float:
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other):
```
mltdll, line 45: There's nothing interesting here.

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if type(other) == int or type(other) == float:
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other):
```
mltdll, line 50: You can also optimize `__le__` and `__ge__` methods by returning `not __gt__` and `not __lt__` respectively.
It was on today's lecture, actually. :)

mltdll (changes requested):

mltdll:
The issue with DRY remains. You should implement method:
```python
@staticmethod
def some_method(other) -> int | float:
    # Some actions to make it return either other or other.km
```
so that you can use it in other methods:
```python
def __lt__(self, other):
    return self.km < some_method(other) 
```

linaSam:
> The issue with DRY remains. You should implement method:
> 
> ```python
> @staticmethod
> def some_method(other) -> int | float:
>     # Some actions to make it return either other or other.km
> ```
> 
> so that you can use it in other methods:
> 
> ```python
> def __lt__(self, other):
>     return self.km < some_method(other) 
> ```
> 
> P.S. You shouldn't pick name `some_method`. :)

mltdll (approved):

mltdll (changes requested):
-

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km == self.check_type(other)
+
+    def __le__(self, other) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
+
+    def __len__(self) -> int:
+        return len(self.km)
```
mltdll, line 48: Sorry, one more thing i missed. You don't need to call len() of number. Just return the number itself.

linaSam: Done) Thanks by the way for all comments - I've learnt a lot!

mltdll (approved):

app/main.py
```diff
@@ -1,3 +1,48 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km) -> None:
```
Y-Havryliv, line 2: add annotations everywhere, fo `km` and `other` as well

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __lt__(self, other) -> bool:
+        return self.km < self.check_type(other)
+
+    def __gt__(self, other) -> bool:
+        return self.km > self.check_type(other)
+
+    def __eq__(self, other) -> bool:
+        return self.km == self.check_type(other)
+
+    def __le__(self, other) -> bool:
```
Y-Havryliv, line 41: it's a magic method, use `self > other` to call it

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return not self.__gt__(other)
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
```
Y-Havryliv, line 45: same here

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #190, opened 2022-09-20
Solution

musakovskiy:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __lt__(self, other):
+        return self.km < other
+
+    def __gt__(self, other):
+        return self.km > other
```
IvanRamyk, line 35: Consider that `other` can be a variable of the different types: (`int`, `float`, `Distance`)

IvanRamyk (changes requested):

app/main.py
musakovskiy: add loop for for two case:  int/float and class object

PR #191, opened 2022-09-20
Solution

pavlo-kapitanets:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        if type(other) == Distance:
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if type(other) == Distance:
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Ilyakson, line 24: return self is duplicated

Ilyakson (changes requested):

app/main.py
pavlo-kapitanets: Done

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km == other
+
+    def __le__(self, other) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other) -> bool:
+        return self.km >= other
+
+    def __len__(self) -> int:
+        return len(self.km)
```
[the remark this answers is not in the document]
pavlo-kapitanets: Done

Ilyakson (approved):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self
+
+    def __mul__(self, other) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other) -> Distance:
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other) -> bool:
+        return self.km < other
```
TPolina, line 33: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km > other
+
+    def __eq__(self, other) -> bool:
+        return self.km == other
+
+    def __le__(self, other) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other) -> bool:
+        return self.km >= other
```
TPolina, line 45: try to use the previously created magic method in these two methods to simplify your code

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if type(other) == Distance:
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other) -> Distance:
```
TPolina, line 26: add type annotation everywhere

TPolina (changes requested):

TPolina (changes requested):

app/main.py
```diff
@@ -45,14 +45,10 @@ def __eq__(self, other: int | float | Distance) -> bool:
         return self.km == other
 
     def __le__(self, other: int | float | Distance) -> bool:
-        if isinstance(other, Distance):
-            return not self.km > other.km
-        return not self.km > other
+        return not Distance.__gt__(self, other)
```
TPolina, line 48: magic method `__gt__` is called by `>` symbol. And also we call only classmethods through class names. So the correct solution is:
`return not self > other`

app/main.py
```diff
@@ -45,14 +45,10 @@ def __eq__(self, other: int | float | Distance) -> bool:
...
 
     def __ge__(self, other: int | float | Distance) -> bool:
-        if isinstance(other, Distance):
-            return not self.km < other.km
-        return not self.km < other
+        return not Distance.__lt__(self, other)
```
TPolina, line 51: same here

TPolina (changes requested):

TPolina (approved):

PR #192, opened 2022-09-20
Solution

Qusert17:


app/main.py
```diff
@@ -1,68 +1,71 @@
...
             return self.km <= other.km
         else:
             return self.km <= other
 
-    def __ge__(self, other):
+    def __ge__(self, other: Union[Distance, int, float]) -> bool:
         if isinstance(other, Distance):
             return self.km >= other.km
         else:
             return self.km >= other
```
fsocie7y, line 68: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,68 +1,71 @@
...
+    def __mul__(self, other: Union[int, float]) -> Union[Distance, None]:
+        if isinstance(other, (int, float)):
             return Distance(self.km * other)
 
-    def __truediv__(self, other):
+    def __truediv__(self, other: Union[int, float]) -> Union[Distance, None]:
         if isinstance(other, Distance):
             return Distance(round(self.km / other.km, 2))
         else:
             return Distance(round(self.km / other, 2))
```
fsocie7y, line 39: Rewrite this metod as __mul__. Other can`t be type Distance same as in mul. Beaucouse if u do this - u get km-square.

app/main.py
```diff
@@ -1,68 +1,71 @@
...
             return Distance(self.km + other)
 
-    def __iadd__(self, other):
+    def __iadd__(self, other: Union[Distance, int, float]) -> Distance:
         if isinstance(other, Distance):
             self.km += other.km
             return self
         else:
             self.km += other
             return self
```
fsocie7y, line 28: return self is duplicated, take it outside if-else

fsocie7y (changes requested):

app/main.py
Qusert17: Done)

fsocie7y (approved):

fsocie7y:
U amazing :)

TPolina (approved):
Good job!

PR #193, opened 2022-09-20
Solution

ShevchukVitalii:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
Yaroslav-Stetsko, line 53: You can write __le__ and __ge__ in a one line using previously created magic methods

app/main.py
```diff
@@ -1,3 +1,56 @@
 class Distance:
     # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
```
Yaroslav-Stetsko, line 9: Add type annotations everywhere (for return values as well)

Yaroslav-Stetsko (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return self.km < other.km
+        return self.km < other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return not self < other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return not self > other
+
+    def __len__(self) -> int:
```
Y-Havryliv, line 53: here, the return value can be `int` or `float`

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other) -> Distance:
```
Y-Havryliv, line 29: add type annotation, `other: int | float`

Y-Havryliv (changes requested):

PR #194, opened 2022-09-20
'Solution'

Nikyxa:


app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, (int, float)):
```
mltdll, line 12: __add__ method should return new `Distance` instance, not change it in place.

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+            self.km = self.km + other
+        else:
+            self.km = self.km + other.km
+        return self
+
+    def __iadd__(self, other):
+        if isinstance(other, (int, float)):
```
mltdll, line 19: Since you're using this check several times across your code, you should implement a new method that checks if other is a number or `Distance` object, and returns `other` or `other.km` respectively.

mltdll (changes requested):

TPolina (approved):
Well done!

PR #195, opened 2022-09-20
Solution

YevheniiDavydenko:


app/main.py
```diff
@@ -1,3 +1,61 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km):
```
TPolina, line 2: please, add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        else:
+            return self.km <= other.km
+
+    def __ge__(self, other):
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        else:
+            return self.km >= other.km
```
TPolina, line 58: try to rewrite these two methods in one line each. Use previously created magic methods

TPolina (changes requested):

app/main.py
```diff
@@ -1,61 +1,60 @@
...
         return Distance(self.km * other)
 
-    def __truediv__(self, other):
+    def __truediv__(self, other: Union[int, float]) -> Distance:
         return Distance(round(self.km / other, 2))
 
-    def __lt__(self, other):
-        if isinstance(other, (int, float)):
+    def __lt__(self, other: Union[int, float]) -> bool:
+        if isinstance(other, (int, float, Distance)):
```
TPolina, line 36: Distance?

TPolina (changes requested):

app/main.py
YevheniiDavydenko: missed with copy-past)

TPolina (approved):

PR #196, opened 2022-09-20
'Solution'

serviza:


app/main.py
```diff
@@ -1,3 +1,60 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self):
```
olia-trofymchuk, line 6: Add type annotations everywhere

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other):
```
olia-trofymchuk, line 54: You can write __le__ and __ge__ in a one line using previously created magic methods :)

olia-trofymchuk (changes requested):

app/main.py
serviza: > You can write **le** and **ge** in a one line using previously created magic methods :)

Nice lifehack, thanks)

serviza:
type annotations added to methods

olia-trofymchuk (approved):
Looks good to me 👍

serviza:
> 

Thank you 4 reviewing :)

app/main.py
```diff
@@ -1,3 +1,59 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
Y-Havryliv, line 6: add annotations everywhere, for return values as well, and consider that `km` can be `int` or `float`, and `other` can be `int` or `float` or even `Distance` somewhere

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __add__(self, other) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
Y-Havryliv, line 23: it's better to use `else` here to avoid duplicating `return self`

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other) -> Distance:
+        if isinstance(other, Distance):
```
Y-Havryliv, line 28: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km * other.km)
+        return Distance(self.km * other)
+
+    def __truediv__(self, other) -> Distance:
+        if isinstance(other, Distance):
```
Y-Havryliv, line 33: same here as in mul method

Y-Havryliv (changes requested):

app/main.py
serviza: > it's better to use `else` here to avoid duplicating `return self`

Ok, 'else' added

app/main.py
serviza: > `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

Done

app/main.py
serviza: > add annotations everywhere, for return values as well, and consider that `km` can be `int` or `float`, and `other` can be `int` or `float` or even `Distance` somewhere
I changed annotations for 'other' and for 'km'
And for return values all the annotations are there...

olia-trofymchuk (approved):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __iadd__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[int, float, Distance]) -> Distance:
```
Y-Havryliv, line 30: `other` can't be `Distance` here, only `int` or `float`, and you don't need to check its type

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        return Distance(self.km * other.km)
+
+    def __truediv__(self, other: Union[int, float]) -> Distance:
+        if isinstance(other, (int, float)):
```
Y-Havryliv, line 36: `other` can't be `Distance` here, only `int` or `float`, and you don't need to check its type

app/main.py
Y-Havryliv: and remove `Distance` from annotation here

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Union[int, float, Distance]) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Union[int, float, Distance]) -> bool:
+        return not self.__gt__(other)
```
Y-Havryliv, line 56: it's a magic method, use `self > other`

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __ge__(self, other: Union[int, float, Distance]) -> bool:
+        return not self.__lt__(other)
```
Y-Havryliv, line 59: same here, don't use `self.__lt__(other)`

Y-Havryliv (changes requested):

app/main.py
serviza: changed

app/main.py
serviza: > `other` can't be `Distance` here, only `int` or `float`, and you don't need to check its type

changed

app/main.py
serviza: > and remove `Distance` from annotation here

changed

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return self.km == other
+
+    def __le__(self, other: Union[int, float, Distance]) -> bool:
+        return not self > other
+
+    def __ge__(self, other: Union[int, float, Distance]) -> bool:
+        return not self < other
+
+    def __len__(self) -> int:
+        return len(self)
```
Y-Havryliv, line 58: you don't need to return `len(self)`, just return `self.km`

Y-Havryliv (changes requested):

app/main.py
serviza: Ok! Changed.

Y-Havryliv (approved):
Well done!

serviza:
> Well done!

In God we trust! )))

PR #197, opened 2022-09-20
'Solution'

LiudmylaKulinchenko:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
olia-trofymchuk, line 12: you should add type annotation here

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if type(other) == Distance:
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other):
```
olia-trofymchuk, line 18: same here

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if type(other) == Distance:
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other):
```
olia-trofymchuk, line 25: same here

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if type(other) == Distance:
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int):
```
olia-trofymchuk, line 31: same here

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __le__(self, other) -> bool:
+        if type(other) == Distance:
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
+
+    def __len__(self):
```
olia-trofymchuk, line 55: and here :)

olia-trofymchuk (changes requested):

olia-trofymchuk (approved):
as for me, eveything looks good :)

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other) -> Distance:
+        if type(other) == Distance:
+            return Distance(self.km * other.km)
+        else:
+            return Distance(self.km * other)
```
TPolina, line 32: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if type(other) == Distance:
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other) -> bool:
+        return not self.__le__(other)
+
+    def __eq__(self, other) -> bool:
+        return self.km == other
```
TPolina, line 47: add type check here as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other) -> bool:
+        if type(other) == Distance:
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other) -> bool:
+        return not self.__le__(other)
```
TPolina, line 44: we use magic methods with a special interface. For example `+` will call `__add__` method

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return self.km == other
+
+    def __le__(self, other) -> bool:
+        if type(other) == Distance:
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
```
TPolina, line 56: same here with magic method as in `gt` method

LiudmylaKulinchenko: And may you answer one question, please: how to name my new commit with changes correctly? I wrote just 'Changed' and don't really understand how to name it better

TPolina: "Fix magic methods usage", "Fix type checks in all methods" for example
You need to describe in the commit what was added, fixed, or changed, so the other people understand what is in this commit.

TPolina (approved):

PR #198, opened 2022-09-21
Solution

FranchukS:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > Distance(other).km
+
+    def __ge__(self, other):
+        return self.km >= Distance(other).km
+
+    def __le__(self, other):
+        return self.km <= Distance(other).km
+
+    def __eq__(self, other):
+        return self.km == Distance(other).km
```
Yaroslav-Stetsko: Please add type annotations everywhere, and for return values as well

Yaroslav-Stetsko (changes requested):

app/main.py
FranchukS: Annotation added

Yaroslav-Stetsko (approved):
Looks good now

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, Distance)) -> Distance:
+        return Distance(self.km + other.km) if \
+            isinstance(other, Distance) else \
+            Distance(self.km + Distance(other).km)
```
TPolina, line 17: really bad style. Don't use `\`. Instead, use a simple if-else block. Please, change all methods

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __ge__(self, other: (int, float, Distance)) -> bool:
+        return self.km >= other.km if \
+            isinstance(other, Distance) else \
+            self.km >= other
+
+    def __le__(self, other: (int, float, Distance)) -> bool:
+        return self.km <= other.km if \
+            isinstance(other, Distance) else \
+            self.km <= other
```
TPolina, line 50: try to rewrite these two methods in one line each. Use previously created magic methods

TPolina (approved):
Well done!

PR #199, opened 2022-09-21
Solution

dmytro-petliovanyi:


Aleksandro777 (approved):
Good job!😊

app/main.py
```diff
@@ -1,3 +1,60 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
Y-Havryliv, line 6: `km` can be float as well

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other) -> Distance:
```
Y-Havryliv, line 15: add annotation everywhere (for `other` as well)

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return self.km > other.km
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Distance):
+            return self.km == other
+
+        return self.km == other.km
+
+    def __le__(self, other) -> bool:
+        return not self.__gt__(other)
```
Y-Havryliv, line 54: it's a magic method, you don't need to call it directly, use `self > other`

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __ge__(self, other) -> bool:
+        return not self.__lt__(other)
```
Y-Havryliv, line 57: same here, don't use `self.__lt__(other)`

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __len__(self) -> int:
```
Y-Havryliv, line 59: this method can return `float` as well

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
Y-Havryliv, line 15: almost good, but `other` can be `int` and `float` as well, fix everywhere

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+
+        return self
+
+    def __mul__(self, other: Distance) -> Distance:
```
Y-Havryliv, line 29: `other` can't be `Distance` here

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance) -> Distance:
```
Y-Havryliv, line 32: `other` can't be `Distance` here

Y-Havryliv (changes requested):

Y-Havryliv (approved):
🔥

PR #200, opened 2022-09-21
Solution

illiariaboy94:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __gt__(self, other) -> bool:
+        return self.km > other.real if isinstance(other, (int, float)) \
+            else self.km > other.km
+
+    def __eq__(self, other) -> bool:
+        return self.km == other.real if isinstance(other, (int, float)) \
+            else self.km == other.km
+
+    def __le__(self, other) -> bool:
+        return self.km <= other.real if isinstance(other, (int, float)) \
```
Kitazuka, line 54: U can use previously created magic methods here

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            else self.km <= other.km
+
+    def __ge__(self, other) -> bool:
+        return self.km >= other.real if isinstance(other, (int, float)) \
```
Kitazuka, line 58: And here too

Kitazuka (changes requested):

app/main.py
illiariaboy94: Thanks

app/main.py
illiariaboy94: OK

app/main.py
```diff
@@ -1,3 +1,60 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km):
```
Y-Havryliv, line 6: add annotations everywhere, for `km` and `other` as well

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other.real
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other) -> Distance:
+        return Distance(
```
Y-Havryliv, line 30: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            self.km * other.real if isinstance(other, (int, float))
+            else self.km * other.km
+        )
+
+    def __truediv__(self, other) -> Distance:
+        return Distance(
```
Y-Havryliv, line 36: same here as in mul method

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return self.km == other.real if isinstance(other, (int, float)) \
+            else self.km == other.km
+
+    def __le__(self, other) -> bool:
+        return not self > other
+
+    def __ge__(self, other) -> bool:
+        return not self < other
+
+    def __len__(self) -> float:
```
Y-Havryliv, line 59: it can return `int` as well

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #201, opened 2022-09-21
Solution

fsocie7y:


ketstap162 (approved):
Nice job!

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
```
TPolina, line 27: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km * other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km * other)
```
TPolina, line 33: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(round(self.km / other.km, 2))
+        elif isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
TPolina, line 39: same as in mul method

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
```
TPolina, line 63: try to rewrite these two methods in one line each. Use previously created magic methods

TPolina (changes requested):

TPolina (approved):
Well done!

PR #202, opened 2022-09-21
Solution

meloman-13:


app/main.py
```diff
@@ -1,3 +1,66 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
ketstap162, line 58: Use your magic method `gt` for `le`

```
le:
return not self > other
```

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
ketstap162, line 63: Same:
`return not self < other`

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+
+    def __len__(self) -> int:
+        return self.km
```
ketstap162, line 66: Annotation to len:
`int | float`

app/main.py
```diff
@@ -1,3 +1,66 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
```
ketstap162, line 6: Annotation for `km`:
`int | float`

ketstap162 (changes requested):

app/main.py
Y-Havryliv: I agree

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        )
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
```
Y-Havryliv, line 30: `other` can't be `Distance` in here

app/main.py
Y-Havryliv: I agree

app/main.py
Y-Havryliv: I agree

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return not self.__gt__(other)
```
Y-Havryliv, line 56: it's a magic method, it's better to call it using `return not self > other`

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return not self.__lt__(other)
```
Y-Havryliv, line 59: same here, don't use `self.__lt__(other)`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #203, opened 2022-09-21
Solution

Vadosh:


app/main.py
```diff
@@ -1,11 +1,11 @@
 class Distance:
-    def __init__(self, km):
+    def __init__(self, km: (int, float)) -> None:
```
Yaroslav-Stetsko, line 2: These annotations will not work unfortunately :(
Use either one of these instead
if python 3.10+ use: int | float
if older : Union[int, float]

Yaroslav-Stetsko (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __le__(self, other):
```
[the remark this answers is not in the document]
Yaroslav-Stetsko: Great gob!

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other):
```
Yaroslav-Stetsko, line 26: returns Distance :   -> Distance
You need to import:   from __future__ import annotations

Yaroslav-Stetsko (changes requested):
Fix these and you should be Done

Yaroslav-Stetsko (approved):
=)
You did great!

Yaroslav-Stetsko (approved):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other) -> bool:
```
Nattalli, line 32: the `other` value needs a type annotation like in the methods above. So, add type annotation for `other`  for this and next methods, please

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __lt__(self, other) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __ge__(self, other) -> bool:
+        return not self < other
+
+    def __le__(self, other) -> bool:
+        return not self > other
```
Nattalli, line 51: also you have to add a `__len__` method
From the task:

```python
__len__
distance = Distance(20)
len(distance) == 20
```

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #204, opened 2022-09-21
Solution

mariia-movchan:


app/main.py
```diff
@@ -1,3 +1,56 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int):
+        self.km = km
+
+    def __str__(self):
```
Mowka7, line 5: add annotations everywhere

ShevchukVitalii (approved):
gj

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
Mowka7, line 53: you can rewrite these two methods in one line each. Use previously created magic methods
Hint: <= == not >

Mowka7 (changes requested):

app/main.py
mariia-movchan: fixed

Mowka7 (approved):
Well done!

app/main.py
```diff
@@ -1,3 +1,54 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Nattalli, line 5: the `km` variable can be a `float` too.
So if you have python 3.10 then use `int | float` or Union[int, float] for earlier versions

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other) -> Distance:
```
Nattalli, line 14: you have to add a type annotation for the `other` variable in each function. Also, don't forget that now it can be `int`, `float` and `Distance`

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other) -> Distance:
```
Nattalli, line 19: add type annotation for `other` here too

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other) -> Distance:
```
Nattalli, line 26: add type annotation for `other` here too

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other) -> Distance:
```
Nattalli, line 29: add type annotation for `other` here too

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other) -> bool:
```
Nattalli, line 32: add type annotation for `other` here too

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other) -> bool:
```
Nattalli, line 37: add type annotation for `other` here too

Nattalli (changes requested):

app/main.py
mariia-movchan: fixed

app/main.py
mariia-movchan: Added type annotations for the other variable in each function

Nattalli (approved):
Good job! 😸

PR #205, opened 2022-09-22
distance first attempt

VitaliyBychevoy:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if not isinstance(other, Distance):
```
linaSam, line 13: you can write static function which will check type of other. so you won't need to write additional isinstance in every (or almost every) function

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            if type(other) is int or type(other) is float:
+                return Distance(km=self.km + other)
+            raise TypeError(f"unsupported operand type(s) for + :"
+                            f"'Distance' and {type(other)}")
+        return Distance(km=self.km + other.km)
+
+    def __iadd__(self, other):
+        if type(other) is Distance:
```
linaSam, line 21: same: static function for checking type of other

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return self
+        elif type(other) is int or type(other) is float:
+            self.km += other
+            return self
+        else:
+            raise TypeError(f"{type(other)} is error type."
+                            f" Need type Distance, int or float.")
+
+    def __mul__(self, other):
+        if type(other) is int or type(other) is float:
```
linaSam, line 32: same: static function for checking type of other

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return Distance(self.km * other)
+        else:
+            raise TypeError(f"{type(other)} is error type."
+                            f" Need type int or float.")
+
+    def __truediv__(self, other):
+        if type(other) is int or type(other) is float:
```
linaSam, line 39: same: static function for checking type of other

app/main.py
```diff
@@ -1,3 +1,61 @@
 class Distance:
     # Write your code here
-    pass
+    def __init__(self, km: float):
```
linaSam, line 3: add annotations for functions

linaSam (changes requested):

app/main.py
linaSam: still, code needs annotations. for example in this one it should look like this:

linaSam (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __add__(self, other):
+        return Distance(km=self.km + self.check_argument(other))
+
+    def __iadd__(self, other):
+        self.km += self.check_argument(other)
+        return self
+
+    def __mul__(self, other):
+        return Distance(self.km * self.check_argument(other))
```
TPolina, line 19: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __truediv__(self, other):
+        return Distance(km=round(self.km / self.check_argument(other), 2))
```
TPolina, line 22: same as in `mul` method

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
TPolina, line 11: add type annotation everywhere

TPolina (changes requested):

TPolina (approved):

PR #207, opened 2022-09-22
Solution

OleksiiKarpov:


app/main.py
```diff
@@ -1,3 +1,74 @@
...
+
+        if isinstance(other, (int, float)):
+            return True if self.km < other else False
+
+    def __gt__(self, other):
+        if isinstance(other, Distance):
+            return True if self.km > other.km else False
+
+        if isinstance(other, (int, float)):
+            return True if self.km > other else False
```
IvanRamyk, line 50:
```suggestion
            return self.km > other
```
You can just return the value of the condition. It's already a boolean value

IvanRamyk (changes requested):

DmytroSvirsa (changes requested):
read the checklist to avoid else with return statement

PR #214, opened 2022-09-28
Solution

NVMBR13:
V2

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,57 @@
 class Distance:
     # Write your code here
-    pass
+    def __init__(self, km):
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
DmytroSvirsa, line 15: read the checklist to avoid else with return statement

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+        else:
```
DmytroSvirsa, line 22: read the checklist to avoid else with return statement

DmytroSvirsa (changes requested):
read the checklist to avoid else with return statement

PR #227, opened 2022-10-08
Develop

Thirteenthskyi:


Thirteenthskyi:
About the magical function __add__, in my opinion "else" there is not superfluous

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
```
Abnormaltype, line 11:
```suggestion
    def __add__(self, other: int) -> Distance:
```

Abnormaltype: fix it in other cases, If you got a problem, please use this import:
`from __future__ import annotation`

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #236, opened 2022-10-12
Fix tests for distance in mul and truediv

TPolina:


tests/test_main.py
```diff
@@ -192,11 +192,14 @@ def test_distance_class_mul_number(kilometers, number, result):
...
     distance2 = Distance(3)
-    with pytest.raises(TypeError) as error_info:
-        distance1 * distance2
-    assert "unsupported operand" in str(error_info.value), (
-        "'__mul__' method should not accept Distance instance"
-    )
+    try:
+        result = distance1 * distance2
+    except TypeError:
+        assert True
```
IvanRamyk, line 198: why not just use `pass`?

IvanRamyk: If you use this approach

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #242, opened 2022-10-14
Solution

Ida83:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, (float, int)):
+            return self.km <= other
+        if isinstance(other, Distance):
+            return self.km <= other.km
+
+    def __ge__(self, other: float) -> bool:
+        if isinstance(other, (float, int)):
+            return self.km >= other
+        if isinstance(other, Distance):
+            return self.km >= other.km
```
DmytroSvirsa, line 62: you may be sure that it will be one of float, int, or Distance so you can avoid additional checks, look whole code for it

DmytroSvirsa (changes requested):

app/main.py
Ida83: I deleted additional check

DmytroSvirsa (approved):

PR #243, opened 2022-10-15
Solution

dmitriy-kds:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __mul__(self, num: int) -> Distance:
+        distance = Distance(self.km * num)
+        return distance
+
+    def __truediv__(self, num: int) -> Distance:
+        distance = Distance(round(self.km / num, 2))
+        return distance
+
+    def __lt__(self, num: int) -> bool:
```
DmytroSvirsa, line 37: provide logick for num type float and Distance

DmytroSvirsa (changes requested):

app/main.py
dmitriy-kds: sorry, I don't understand what I need to do here – should I rewrite the code so that num type is float/Distance? Or should I just explain why I used "int" as the type? Could you please explain a little more – I'm stuck

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> int or float or Distance:
+        if not isinstance(other, Distance):
+            distance = Distance(self.km + other)
+        else:
```
Abnormaltype, line 18: remove unnecessary else

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> int or float or Distance:
```
```suggestion
    def __add__(self, other: Distance) -> Union[int, float, Distance]:
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __iadd__(self, other: Distance) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, num: int) -> Distance:
+        distance = Distance(self.km * num)
+        return distance
```
Abnormaltype, line 31:
```suggestion
        return Distance(self.km * num)
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __truediv__(self, num: int) -> Distance:
+        distance = Distance(round(self.km / num, 2))
+        return distance
+
+    def __lt__(self, num: int or float or Distance) -> bool:
+        if isinstance(num, Distance):
+            return self.km < num.km
+        else:
```
Abnormaltype, line 40: remove unnecessary else

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            return self.km < num
+
+    def __gt__(self, num: int) -> bool:
```
Abnormaltype, line 43: it can be not only int

Abnormaltype (changes requested):

IvanRamyk (approved):

PR #246, opened 2022-10-23
Solution

tinabushuieva:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union) -> object:
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
Abnormaltype, line 18: remove unnecessary else

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: Union) -> object:
+        return Distance(round(self.km / number, 2))
+
+    def __lt__(self, number: Union) -> bool:
+        inst = Distance(number)
+        return self.km < inst.km
+
+    def __gt__(self, number: Union) -> bool:
```
Abnormaltype, line 39: number it's Union?

Abnormaltype (changes requested):

app/main.py
tinabushuieva: if I do this I get AssertionError: Result of sum of Distance instances should be Distance instance

app/main.py
tinabushuieva: I thought yes, because number can be int, float or object

app/main.py
Abnormaltype: def __add__(self, other: Union) -> object:

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
+from typing import Union
+
```
Abnormaltype, line 2: from __future__ import annotations

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[int, float, Distance]) -> Distance:
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #249, opened 2022-10-26
Solution

Foksanv:


inrock1 (approved):
I don't have any remarks.

P.S.
by the task and by tests all comparisons more/less/equal work only for number attribute.

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 30: `other` can not be of type `Distance` in this method - make a check for this

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #258, opened 2022-11-05
Solution

ViktorKrUA:


app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __mul__(self, other: Union[int, float]) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float]) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union[int, float, Distance]) -> bool:
+        return self.km < Distance(other).km
```
DmytroSvirsa, line 36: what if other if int or float, these types have no attribute `.km`

DmytroSvirsa (changes requested):

app/main.py
ViktorKrUA: > what if other if int or float, these types have no attribute `.km`

I was trying to compare an instance with int, float and other instances and the result was generated correctly without any Error.  Please let me know what did I miss. 

def __lt__(self, other: Union[int, float, Distance]) -> bool:
    return self.km < other

distance = Distance(3)
distance1 = Distance(4)

print(distance.__lt__(distance1))
print(distance.__lt__(Distance(4)))
print(distance1.__lt__(4.05))
print(distance1.__lt__(2))

(True, <class '__main__.Distance'>)
(True, <class '__main__.Distance'>)
(True, <class 'float'>)
(False, <class 'int'>)

DmytroSvirsa (approved):

PR #259, opened 2022-11-05
Solution

NVMBR13:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __mul__(self, other: Union[int, float]) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float]) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union[int, float, Distance]) -> bool:
+        return self.km < Distance(other).km
```
DmytroSvirsa, line 37: what will happen if the other is int or float type?

DmytroSvirsa (changes requested):

app/main.py
NVMBR13: fixed everything(maybe), also used "|" insted of deprecated Union
Should be good this time
Hopefully

PR #261, opened 2022-11-06
'Solution'

contritum:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            km=self.km * other
+        )
+
+    def __truediv__(self, other: Union[int, float]) -> Distance:
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: Union[int, float]) -> bool:
+        return self.km < other
```
DmytroSvirsa, line 42: other also can be the Distance instance

DmytroSvirsa (changes requested):

app/main.py
DmytroSvirsa: so you need to implement aditional logick that will work if other is Distance instanse.

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #265, opened 2022-11-09
Solution

Oleksiy-Liubchenko:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            self.km * other
+        )
+
+    def __truediv__(self, other: int or float) -> callable:
+        return Distance(
+            round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: int or float) -> callable:
+        return self.km < other
```
DmytroSvirsa, line 39: other can be a Distance instance provide logic for it

DmytroSvirsa (changes requested):

app/main.py
DmytroSvirsa: but it is not the only place with such a problem, also add Distance to type annotation
use `from __future__ import annotations` to provide the correct return annotation

DmytroSvirsa (changes requested):

DmytroSvirsa (changes requested):
the logic is correct now
I want you to correct type annotation
is that method resive only int or float, and is it return callable?
`def __lt__(self, other: int or float) -> callable:`

Oleksiy-Liubchenko:
> 

do you mean that in this method and methods after this i need to write like this? def __lt__(self, other: int or float) -> Distance(int or float).

Oleksiy-Liubchenko:
and now the only issue in this code that i do:  -> callable?

app/main.py
```diff
@@ -21,47 +21,47 @@ def __add__(self, other: int or float) -> callable:
...
         )
 
-    def __truediv__(self, other: int or float) -> callable:
+    def __truediv__(self, other: int or float) -> Distance(int or float):
         return Distance(
             round(self.km / other, 2)
         )
 
-    def __lt__(self, other: int or float) -> callable:
+    def __lt__(self, other: int or float) -> (int or float):
```
DmytroSvirsa, line 41: the comparison operatorі like >, <, == return bool - cange for all comparison magick methods

app/main.py
```diff
@@ -21,47 +21,47 @@ def __add__(self, other: int or float) -> callable:
...
         return self
 
-    def __mul__(self, other: int or float) -> callable:
+    def __mul__(self, other: int or float) -> Distance(int or float):
         return Distance(
             self.km * other
         )
 
-    def __truediv__(self, other: int or float) -> callable:
+    def __truediv__(self, other: int or float) -> Distance(int or float):
```
DmytroSvirsa, line 36: just `-> Distance:`

DmytroSvirsa (changes requested):
In general, you understand me correctly

app/main.py
```diff
@@ -1,3 +1,67 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int or float) -> callable:
```
Abnormaltype, line 6:
```suggestion
    def __init__(self, km: int or float) -> None:
```

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        self.km = km
+
+    def __str__(self) -> callable:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> callable:
+        return f"Distance(km={self.km})"
```
Abnormaltype, line 13: these methods return the string

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __add__(self, other: int or float) -> Distance(int or float):
```
```suggestion
    def __add__(self, other: int or float) -> Distance(int or float):
```
use Union[Distance, int, float]

Abnormaltype (changes requested):
fix it in other cases

PR #266, opened 2022-11-10
Solution

aarrtemm:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __truediv__(self, other: int | float) -> callable:
+        if isinstance(other, int | float):
+            return Distance(km=round(self.km / other, 2))
+        else:
+            raise TypeError("TypeError()")
+
+    def __lt__(self, other: int | float) -> callable:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
DmytroSvirsa, line 40: remove else everywhere you can

DmytroSvirsa (changes requested):

app/main.py
aarrtemm: Oh, I forgot

PR #267, opened 2022-11-11
Solution

Isles22:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance) -> bool:
+        return self.km < other
```
DmytroSvirsa, line 34: other can also be int or float add logic for this in all comparison methods

DmytroSvirsa (changes requested):

Isles22:
fixed

Isles22:
Solution fixed

пт, 11 нояб. 2022 г. в 16:11, Dmytro Svirsa ***@***.***>:

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        else:
```
OleksandrPozden, line 36: redundant ```else```

OleksandrPozden (changes requested):

app/main.py
Isles22: removed else

Isles22: Please need help
How to do without else

Isles22: I got it

OleksandrPozden (approved):
good!

PR #268, opened 2022-11-13
Solution

andriy-md:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, int) or isinstance(other, float):
```
Abnormaltype, line 17:
```suggestion
        elif isinstance(other, float):
```

Abnormaltype (changes requested):
fir it also in other cases

app/main.py
andriy-md: If I write `elif isinstance(other, float):`, I obtain errors during testing (in cases when int is entered).
In Internet I've found that it is possible to use tuple to check for multiple types with `isinstance`

So, maybe this solution will be better:
`if isinstance(other, (int, float))`

Thanks!

Abnormaltype: `if isinstance(other, (int, float))`
yes it's a solution, GitHub deleted int🧐

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return True if self.km < other.km else False
```
Abnormaltype, line 36: you can write like this:
```suggestion
            return self.km < other.km
```
fix it in other cases

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return self.km <= other
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        if isinstance(other, (int, float)):
+            return self.km >= other
+
+    def __len__(self) -> int:
+        return len(range(int(self.km)))
```
Abnormaltype, line 65:
```suggestion
        return self.km
```

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, (int, float)):
+            return self.km < other
+
+    def __gt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        if isinstance(other, (int, float)):
+            return self.km > other
+
+    def __eq__(self, other: Distance | int) -> bool:
```
Abnormaltype, line 46:
```suggestion
    def __eq__(self, other: Distance | int | float) -> bool:
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #269, opened 2022-11-15
Distance class solved

Polyakiv-Andrey:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
Abnormaltype, line 19: Unnecessary else

Abnormaltype (changes requested):

app/main.py
Polyakiv-Andrey: I deleted "else", but without that the conditions will not work. In the check, "other" could be classe instance and number

Abnormaltype (approved):
Well done!

PR #274, opened 2022-11-18
distanse

VItalik1292:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, number: int) -> Distance:
+        return Distance(km=self.km * number)
```
Serhii-Leonenko, line 29: I think number can be float too. And you should add check for type of input number. Also fix annotations and add check in other similar cases

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #289, opened 2022-11-28
Solution

Nexxel05:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, int | float):
```
IvanRamyk, line 15:
```suggestion
        if isinstance(other, (int, float)):
```
The correct way to use `isinstance`

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return Distance(self.km + other)
+        else:
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, int | float):
+            self.km += other
+            return self
+        else:
```
IvanRamyk, line 24: Please, remove the redundant else

app/main.py
IvanRamyk: fix everywhere

app/main.py
IvanRamyk: fix everywhere

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #298, opened 2022-12-05
Solution

Andrii-Kozlovskyi:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+        else:
+            self.km = self.km + other
+            return self
```
YuriiKindrat, line 27: U can use only one return here, cause u return the same thing every time

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
```
Serhii-Leonenko, line 15: Fix annotations, `other `can be not only int, also it doesn't return int. Fix in all similar cases

Serhii-Leonenko (changes requested):

IvanRamyk (approved):

PR #308, opened 2022-12-19
1st solution

AnnaKabatova:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> None:
+        if isinstance(other, int | float):
+            return Distance(km=self.km * other)
+        else:
```
Abnormaltype, line 30: Unnecessary else

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            raise TypeError()
+
+    def __truediv__(self, other: int | float) -> None:
+        if isinstance(other, int | float):
+            return Distance(km=round(self.km / other, 2))
+        else:
```
Abnormaltype, line 36: the same

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> None:
+        if isinstance(other, int | float):
+            return Distance(km=self.km * other)
+        else:
+            raise TypeError()
```
Abnormaltype, line 31: Do you need ()?

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> None:
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> None:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> None:
+        return self.km >= other
```
YuriiKindrat, line 50: How will it work if other will be an Distance class?

YuriiKindrat (changes requested):

Abnormaltype (approved):
Well done!

app/main.py
Abnormaltype: it's magic under the hood, and it works without `.km` attr)

PR #314, opened 2022-12-25
Solution

Tania-Kharchuk:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance | int:
```
IvanRamyk, line 14: fix type annotations:
- you always return a `Distance` instance
- function could accept `float` type

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #324, opened 2023-01-02
Solution

avkpol:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> None:
```
Abnormaltype, line 12: ass float as well

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> float:
```
Abnormaltype, line 12:
```suggestion
    def __add__(self, other: int | float) -> float:
```
you return `Distance`

fix it in other cases

Abnormaltype (changes requested):

app/main.py
Abnormaltype: also, you take `Distance` as well

avkpol: Do you mean this way: def __add__(self, other: int | Distance) -> Distance: ? But " NameError: name 'Distance' is not defined'', why it occurs and how to fix it?

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int) -> bool:
+        return self.km >= other
```
Abnormaltype, line 49: what about it?

Abnormaltype (changes requested):

app/main.py
avkpol: do you mean annotation?

Abnormaltype: yes

avkpol: nothing passes except bool! If other - give a hint, please!

Abnormaltype: other can be float or Distance

katevasilenko: You should fix requested changes all over the project. `other` can be not only `int`

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> Distance:
```
olena-hul, line 20: `other` can be Distance too

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | Distance) -> Distance:
+        return Distance(self.km * other)
```
olena-hul, line 28: Add check for isinstance. `other` can be not only `int`

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __round__(self) -> int:
+        return round(self.km)
```
olena-hul, line 31: Is it needed? I can't see it in the condition

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > other
+
+    def __eq__(self, other: float | Distance) -> bool:
+        return self.km == other
+
+    def __le__(self, other: float | Distance) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: float | Distance) -> bool:
+        return self.km >= other
```
olena-hul, line 49: Add checks for `isinstance`

olena-hul (changes requested):

app/main.py
avkpol: why to check in such way, why to add redundant extra rows if the code works well and passes all the tests?

avkpol: Please explain!

olena-hul: These checks are needed because you can pass Distance as well as int of float instances to this method. And the logic of comparison differs. But for your case - I checked the code locally and the tests really run successfullly and I wonder why) I need to find an explanation, need to dig in more into `__ge__` internal implementation

olena-hul: Anyway, you are missing `int` in annotations, add it please

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self
+
+    def __mul__(self, other: int | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __round__(self) -> int:
+        return self.km
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
```
ArsenPidhoretskyi, line 34: As i see it only possible int and float, not Distance

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
```
ArsenPidhoretskyi, line 37: If other is the Distance we need to take km not object

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        return self.km == other
```
ArsenPidhoretskyi, line 43: If other is the Distance we need to take km not object

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.km >= other
```
ArsenPidhoretskyi, line 49: If other is the Distance we need to take km not object

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __iadd__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | Distance) -> Distance:
+        return Distance(self.km * other)
```
ArsenPidhoretskyi, line 28: @avkpol check also this one
I we do distance * distance we can get error

avkpol: No errors, all tests pass

ArsenPidhoretskyi: Then please remove Distance in other. Because due to your type hint you can get the error. Just do:
```
distance1 = Distance(10)
distance2 = Distance(10)
distance3 = distance1 * distance2
```

avkpol: if I remove Distance in other,  I  get  TypeError: unsupported operand type(s) for *: 'int' and 'Distance'

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __round__(self) -> int:
+        return round(self.km, 2)
```
ArsenPidhoretskyi, line 31: Also, You do not use this magic method. You can remove it

app/main.py
ArsenPidhoretskyi: Due to your typing, we already have a possible error:
If you say that we can accept Distance in other, that mean you code must work ok with this
image

ArsenPidhoretskyi: image

ArsenPidhoretskyi (changes requested):

app/main.py
avkpol: but not such test # distance3 = distance1 * distance2 in test_main.py, thats why all tests passed. I fulfilled this task following the readme.

ArsenPidhoretskyi: Okay but you say that your function will correctly work in that case(due to type hinting). Just remove “Distance” type hint for “other” variable, than it will be correct. 
Btw, If test does not exist it does not mean all works correct)

avkpol: do you mean :     def __mul__(self, other: int | str) -> Distance:
                                      return Distance(self.km * other)?
but still TypeError: unsupported operand type(s) for *: 'int' and 'Distance'

avkpol: > Btw, If test does not exist it does not mean all works correct)
no doubt, you are right, but your team wrote these tests

ArsenPidhoretskyi: But tests are passed when you set the str type hint.
If you set that for __mul__ possible only int that you need only test if someone does Distance(10) * some_int
Before you have that other can be int or Distance. That mean your code must work if you do Distance(10) * Distance(10)

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | Distance) -> Distance:
```
ArsenPidhoretskyi, line 27: You need just do:
```suggestion
    def __mul__(self, other: int) -> Distance:
```
And that all will be okay. This action say that your code must work okay if you only do Distance() * any_int

app/main.py
ArsenPidhoretskyi: Yeah
Check here
https://github.com/mate-academy/py-distance-class/pull/324/files#r1072485772

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
ArsenPidhoretskyi, line 49: Due to task requirements, you can compare two distances.
For example:
if isinstance(other, Distance) we need to compare other by km, otherwise just by value

ArsenPidhoretskyi (changes requested):

olena-hul (changes requested):

avkpol:
@olena-hul I'm really tired of this task... Could you please specify what exactly should I do?

app/main.py
ArsenPidhoretskyi: Oh really, It works nice. As I see if we use distance1 > distance2 it will enter gt and then it will enter lt due to `other`

olena-hul:
@avkpol
The final suggestion
1. Update all annotations, to include int, float and Distance (where needed and when these types can be passed)
2. Add isinstance() checks in methods for comparisons (gt, lt, etc.). Because you expect `other` to be Distance too and the logic should be appropriate

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __mul__(self, other: int | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __round__(self) -> int:
+        return round(self.km, 2)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
ArsenPidhoretskyi, line 36:
```suggestion
    def __lt__(self, other: int | float | Distance) -> bool:
```

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km < other
+
+    def __gt__(self, other: int | float) -> bool:
```
ArsenPidhoretskyi, line 39:
```suggestion
    def __gt__(self, other: int | float | Distance) -> bool:
```

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
```
ArsenPidhoretskyi, line 42:
```suggestion
    def __eq__(self, other: int | float | Distance) -> bool:
```

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
```
ArsenPidhoretskyi, line 45:
```suggestion
    def __le__(self, other: int | float| Distance) -> bool:
```

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
```
ArsenPidhoretskyi, line 48:
```suggestion
    def __ge__(self, other: int | float | Distance) -> bool:
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
Abnormaltype, line 6:
```suggestion
    def __init__(self, km: int | float) -> None:
```

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance) -> Distance:
```
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | Distance) -> Distance:
```
Abnormaltype, line 20:
```suggestion
    def __iadd__(self, other: int | float | Distance) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
```
Abnormaltype, line 18:
```suggestion
        if isinstance(other, (int, float)):
            return Distance(self.km + other)
```

Abnormaltype (changes requested):

Vasyl-Smutok (changes requested):
please correct what other mentors wrote to you)

Vasyl-Smutok (approved):
Well Done 👍

PR #327, opened 2023-01-04
Solution

serg121192:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, int) or isinstance(other, float):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, int) or isinstance(other, float):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(self, Distance) and isinstance(other, Distance):
+            return True if self.km < other.km else False
```
Abnormaltype, line 38:
```suggestion
            return self.km < other.km
```
fix it in other cases

Abnormaltype (changes requested):

app/main.py
serg121192: > fix it in other cases

Fixed) And could you check my solution of decorators?)

Abnormaltype: check guidelines on how to do a re-request review.

Vasyl-Smutok (approved):
Good Job 👍

PR #330, opened 2023-01-07
solution

VladyslavOladko:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int or Distance) -> int:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
Vasyl-Smutok, line 17: it is not necessary to use `else` here

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
```
Vasyl-Smutok, line 24: and here

Vasyl-Smutok (changes requested):

katevasilenko (approved):
Looks good to me

PR #332, opened 2023-01-09
'Solution'

Vitalii2906:


app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> object:
```
olena-hul, line 11: Add better annotation for `other`. `Union[Distance, int]` or `Distance | int` if you're using python 3.10 suits better. Fix in other places
The same for return annotation - it should be `Distance` class

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        if isinstance(other, Distance):
+            result = Distance(self.km + other.km)
+        else:
+            result = Distance(self.km + other)
+        return result
```
olena-hul, line 16: And also, this can be done in 3 lines without additional variable

olena-hul (changes requested):

olena-hul:
Good job!

olena-hul (approved):

PR #333, opened 2023-01-09
Solution

xtina999:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> Any:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> Any:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Any:
```
olena-hul, line 14: `other` can be `Distance` too. And specify concrete class for return annotation, not just `Any`

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Any:
+        return Distance(
+            km=self.km * other
+        )
```
olena-hul, line 29: No need to split it in multiple lines if it fits into one line

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: int | float) -> Any:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
olena-hul, line 24: You can use `else` and one `return`

olena-hul (changes requested):

IvanRamyk (approved):

PR #335, opened 2023-01-09
Develop

OleksandrGrk:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Distance) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Distance) -> bool:
+        return self.km >= other
```
olena-hul, line 47: These methods are wrong.
1. `other` can be as Distance as int
2. You need to change methods' implementation to support either int and Distance

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+    def __add__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
olena-hul, line 24: You can use `else` and only one `return`

Vasyl-Smutok (approved):
Well Done 👍

PR #336, opened 2023-01-10
'Solution'

andy77andy:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int or float) -> Distance:
```
olena-hul, line 14: `or` is not used for annotations. Use `Union` or if you're using python 3.10 - `|` operator. And annotation itself is wrong. other can be Distance class too
Fix in other occurrences

olena-hul (changes requested):

Vasyl-Smutok (changes requested):
Well done 👍

Vasyl-Smutok (approved):

PR #337, opened 2023-01-13
Solution

Rename925:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km + other.km) \
+            if isinstance(other, Distance) else Distance(self.km + other)
```
Vasyl-Smutok, line 16: i think it will look better
generally try to avoid `/`
```
        return (
            Distance(self.km + other.km)
            if isinstance(other, Distance)
            else Distance(self.km + other)
        )
```

also please fix this in other parts of the code

Vasyl-Smutok (changes requested):

app/main.py
Rename925: i fixed the mistakes

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        return (Distance(self.km + other.km)
```
ArsenPidhoretskyi, line 15:
```suggestion
        return (Distance(self.km + other.km)
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+                if isinstance(other, Distance)
+                else Distance(self.km + other))
```
ArsenPidhoretskyi, line 17:
```suggestion
        return (
            Distance(self.km + other.km)
            if isinstance(other, Distance)
            else Distance(self.km + other)
        )
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 25: 1. try to rewrite in two rows. You can use inline if.
2. Also usage of two same return is bad practice

app/main.py
ArsenPidhoretskyi: Also formate the same in all places

ArsenPidhoretskyi (changes requested):

app/main.py
Rename925: how shoud i rewrite it in two rows?

Vasyl-Smutok (approved):
Good Job 👍

PR #338, opened 2023-01-18
Distance class

VKungin:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __lt__(self, other: Any) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: Any) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: Any) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Any) -> bool:
```
Vasyl-Smutok, line 42: I think it is better to write like this 
`Union[int, float, Distance]` or `int | float | Distance`
fix it here and other places

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: Any) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: Any) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: Any) -> bool:
+        return self.km == other
```
Vasyl-Smutok, line 40: also worth adding a check 
`isinstance(other, Distance)`

here and other places

Vasyl-Smutok (changes requested):

ArsenPidhoretskyi (approved):
Well done.
Also 
 ```
def __lt__(self, other: Union[int, float, Distance]) -> bool:
            return self.km < other.km
        return self.km < other
```
Will work the same as:
    ```
def __lt__(self, other: Union[int, float, Distance]) -> bool:
        return self.km < other
```

PR #345, opened 2023-01-25
Solution

Roman28101:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[int, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
olena-hul, line 28: Remove return above and move this return on the same level with `if` and `else`

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Union[int, Distance]) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Union[int, Distance]) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Union[int, Distance]) -> bool:
+        return self.km >= other
```
olena-hul, line 50: Add float to annotation and add isinstance checks because other can be distance too

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #347, opened 2023-01-28
'Solution'

bohdan-yatsyna:


ArsenPidhoretskyi (changes requested):
This is wrong `other is not Distance` use isinstance istead
The boolean `__lt__, __gt__, etc` will also work without check for isinstance. (HINT for future)

bohdan-yatsyna:
> e isinstance istead

E           AssertionError: '__mul__' method should not accept Distance instance
E           assert Distance(km=15) is None

bohdan-yatsyna:
1)  This is wrong `other is not Distance` use isinstance istead   - I tried already with isinstance(other, Distance) for __mul__ and __truediv__, but it is not works. Next errors:
E           AssertionError: '__mul__' method should not accept Distance instance
E           assert Distance(km=15) is None
E           AssertionError: '__truediv__' method should not accept Distance instance
E           assert Distance(km=10.0) is None
And due to this I left "is not Distance", as in this case all tests are passed and it is acceptable for __mul__ and __truediv__. 

2) The boolean `__lt__, __gt__, etc` will also work without check for isinstance. (HINT for future) - Thank you for such an info, all worked without isinstance.

ArsenPidhoretskyi:
@bohdan-yatsyna yeah it is mean you code must not work with Distance in this method. mul and truediv must only accept int and float
You can just remove checks in those methods and expect only int and float

bohdan-yatsyna:
> @bohdan-yatsyna yeah it is mean you code must not work with Distance in this method. mul and truediv must only accept int and float You can just remove checks in those methods and expect only int and float

Noted with thanks! Please see my last commit "Change"

olena-hul (approved):

PR #349, opened 2023-01-30
'Solution'

natalie-goriela:


app/main.py
```diff
@@ -1,3 +1,50 @@
+from __future__ import annotations
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 6: float possible

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> Distance:
```
ArsenPidhoretskyi, line 15: Distance(10) + "some_string"
Will not work but, due to your code you say it must works)
Set in all places correct type hinting

ArsenPidhoretskyi (changes requested):
Usage Any is bad Practice for type hinting. Due to your code it not Any

app/main.py
ArsenPidhoretskyi: @natalie-goriela not fixed

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        self.km = self.km * other
+        return self
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
```
ArsenPidhoretskyi, line 32: `__truediv__` can not accept Distance
the same with `__mull__`

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+        else:
+            self.km = self.km + other
+        return self
```
ArsenPidhoretskyi, line 19: add must return new instance

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km = self.km * other
```
ArsenPidhoretskyi, line 29: mull return new instance

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self
+
+    def __truediv__(self, other: int | float) -> Distance:
+        self.km = round(self.km / other, 2)
+        return self
```
ArsenPidhoretskyi, line 34: mul return new instance

ArsenPidhoretskyi (changes requested):

stanhjr (approved):

PR #350, opened 2023-01-30
Solution

mkobylianskyi:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float) -> Distance:
```
Abnormaltype, line 19:
```suggestion
    def __iadd__(self, other: int | float | Distance) -> Distance:
```

Abnormaltype (changes requested):
fix it in other cases as well

katevasilenko (approved):
Well done

PR #354, opened 2023-02-03
Solution

MaybeTI:


AndereLion:
Good job !.

AndereLion (approved):
Good job

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km *= other
+        return self
+
+    def __truediv__(self, other: int | float) -> Distance:
+        self.km = round(self.km / other, 2)
+        return self
```
YuriiKindrat, line 36: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km
also this methods should return new instance of a class

YuriiKindrat (changes requested):

MikhailLyvak (approved):
GJ!

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km *= other
+        return self
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 35: `other` can not be of type `Distance` in this method - make a check for this

YuriiKindrat: mull method also should return new distance class

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 60: you can write these two methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #356, opened 2023-02-03
1 commit

AndereLion:


DavidGlusch (approved):

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance or int or float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
```
wQuelS, line 17: no need to double check with `if`, since if line 15 is not true, you can be sure, that other is not `Distance` type

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __lt__(self, other: Distance or int or float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        if isinstance(other, (int, float)):
+            return self.km < other
+
+    def __gt__(self, other: Distance or int or float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        if isinstance(other, (int, float)):
```
wQuelS, line 45: same with double checks

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            return self.km > other
+
+    def __eq__(self, other: Distance or int or float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, (int, float)):
```
wQuelS, line 51: same

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        if isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        if isinstance(other, (int, float)):
+            return self.km >= other
```
wQuelS, line 64: you can write these two methods in one line each. Use previously created magic methods

wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance or int or float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance or int or float) -> bool:
+        return self.km <= other
```
Y-Havryliv, line 50: either add a check for a `Distance` instance here as well or use previously created magic methods

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __ge__(self, other: Distance) -> bool:
+        return self.km >= other
```
Y-Havryliv, line 53: same here

Y-Havryliv (changes requested):

app/main.py
AndereLion: Fixed.

app/main.py
AndereLion: Also fixed.

wQuelS (approved):
Nice work

PR #358, opened 2023-02-04
Solution

bihunva:


manjustice (approved):
GJ!

YuriiKindrat (approved):
GJ!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> Distance:
+        return Distance(round(self.km / other, 2))
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #359, opened 2023-02-05
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -16,40 +16,40 @@ class Distance:
             return Distance(self.km + other.km)
         return Distance(self.km + other)
 
-    def __iadd__(self, other):
+    def __iadd__(self, other: (int, float, Distance)) -> Distance:
         if isinstance(other, Distance):
             self.km += other.km
         else:
             self.km += other
         return self
 
-    def __mul__(self, other):
+    def __mul__(self, other: (int, float)) -> Distance:
         return Distance(self.km * other)
 
-    def __truediv__(self, other):
+    def __truediv__(self, other: (int, float)) -> Distance:
         return Distance(round(self.km / other, 2))
 
-    def __lt__(self, other):
+    def __lt__(self, other: (int, float, Distance)) -> bool:
         if isinstance(other, Distance):
             return self.km < other.km
         return self.km < other
 
-    def __gt__(self, other):
+    def __gt__(self, other: (int, float, Distance)) -> bool:
         if isinstance(other, Distance):
             return self.km > other.km
         return self.km > other
 
-    def __eq__(self, other):
+    def __eq__(self, other: (int, float, Distance)) -> bool:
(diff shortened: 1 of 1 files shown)


bihunva (approved):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: (int, float)) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: (int, float)) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 30: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: u a right
didn't read the terms carefully
fixed it

NEkropo1 (approved):
I would say, you can return != self.method(other), for some methods, which will improve longer and stronger comparisons typing, but still your work is great )

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __lt__(self, other: (int, float)) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: (int, float)) -> bool:
+        return self.km > other
```
wQuelS, line 36: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __eq__(self, other: (int, float)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: (int, float)) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: (int, float)) -> bool:
+        return self.km >= other
```
wQuelS, line 45: same. you didn't type checked.
you can write this methods in one line even with type checks. Just use previously created magic methods

app/main.py
SafonovVladimir: In my first solution I check it, but @YuriiKindrat  told me that other can not be of type Distance in this method, because if you multiply km * km you get square km.
https://github.com/mate-academy/py-distance-class/pull/359#discussion_r1097762848
According our condition distance < Distance(60), so other can be just class int.

app/main.py
SafonovVladimir: Yes, I know. For example 
return not self.__lt__(other.km) if isinstance(other, Distance) else not self.__lt__(other)
I fixet it.
Please check

NEkropo1: > same. you didn't type checked. you can write this methods in one line even with type checks. Just use previously created magic methods

I'm not that sure )
image

Y-Havryliv (approved):
Nice!

PR #360, opened 2023-02-05
Solution

vlamuha:


app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
```
Vasyl-Smutok, line 51: 1) `else` redundant here
2) add additional info for annotation 
fix it here and other places 

Example:
```suggestion
    def __gt__(self, other: Distance | int | float) -> bool:
         if isinstance(other, Distance):
             return self.km > other.km
         return self.km > other
```

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km *= other
+        return self
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(
+            round(self.km / other, 2)
+        )
```
Vasyl-Smutok, line 39:
```suggestion
        return Distance(round(self.km / other, 2))
```

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+
+        if isinstance(other, Distance):
+            return Distance(
+                self.km + other.km
+            )
```
Vasyl-Smutok, line 19:
```suggestion
            return Distance(self.km + other.km)
```

Vasyl-Smutok (approved):
Good Job 👍 

also check comment

PR #361, opened 2023-02-05
Solution

DavidGlusch:


ZabFTFT (changes requested):
Everything is good except of double return

ZabFTFT (approved):
Good job! Nothing to add ;)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km = self.km + other
+        elif isinstance(other, Distance):
+            self.km = self.km + other.km
+        return self
```
YuriiKindrat, line 19: add method should return new instance
image

YuriiKindrat: Also you can make it with only one condition

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km = self.km + other
+        elif isinstance(other, Distance):
+            self.km = self.km + other.km
+        return self
```
YuriiKindrat, line 26: no need to use isinstace for the 2nd time, u can use else here

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        elif isinstance(other, Distance):
```
YuriiKindrat, line 37: unnecessary elif here, use just else here

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            self.km = self.km + other
+        elif isinstance(other, Distance):
+            self.km = self.km + other.km
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 32: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        elif isinstance(other, Distance):
+            return self.km <= other.km
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        elif isinstance(other, Distance):
+            return self.km >= other.km
```
YuriiKindrat, line 62: remove unnecessary elif from this methods, u can use just else here

YuriiKindrat (changes requested):

oigod (approved):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            self.km = self.km + other
+        else:
+            self.km = self.km + other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 32: `other` can not be of type `Distance` in this method - make a check for this

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #364, opened 2023-02-06
Solution

manjustice:


Zenionoi (approved):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        else:
+            self.km += other.km
+
+        return self
+
+    def __mul__(self, number: int) -> Distance:
+        return Distance(km=self.km * number)
+
+    def __truediv__(self, number: int | float) -> Distance:
+        return Distance(km=round(self.km / number, 2))
```

YuriiKindrat (changes requested):

app/main.py
Zenionoi: @YuriiKindrat but there `other` is only int #L27 and int/float in #L30

DmytryTong (approved):
Everything is accurate and easy to read) Great code!

YuriiKindrat (approved):

PR #365, opened 2023-02-06
Distance solution

Maytreyaya:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
```
SergiiMachulin, line 12: This method could add not only integers, but instances also. And you return instance, but annotated int.

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if type(other) == int or type(other) == float:
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int) -> int:
```
SergiiMachulin, line 17: The same logic with annotations and return in this method

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
+        if type(other) == int or type(other) == float:
```
SergiiMachulin, line 13: In annotation you use only int, but here and below in other methods you check float

SergiiMachulin (changes requested):

SergiiMachulin (approved):
Well done!

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km = self.km + other
+            return self
+        self.km = self.km + other.km
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        if type(other) == int or type(other) == float:
+            return Distance(self.km * other)
+        return Distance(self.km * other)
```
YuriiKindrat, line 29: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __le__(self, other: int | float) -> bool:
+        if type(other) == int or type(other) == float:
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: int | float) -> bool:
+        if type(other) == int or type(other) == float:
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 59: you can write these two methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):
also it is better to use isinstance method for type checking

wQuelS (approved):
GJ

PR #367, opened 2023-02-06
'Solution'

Morty67:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```

YuriiKindrat (changes requested):

viktoria-rybenchuk (approved):

wQuelS (approved):
Good job

PR #368, opened 2023-02-06
Solution

Yevheniy-Lototskiy:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
```
SergiiMachulin, line 17: I'm not sure about additional new line between returns here, and below

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float) -> bool:
```
SergiiMachulin, line 34: According to README.md __lt__, __gt__, __eq__, __le__, __ge__ should compare not only instance and numbers, but also with other insctances

SergiiMachulin (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Nattalli, line 5: `km` can be int as well

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
```
Nattalli, line 14: can be int as well

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float) -> Distance:
```
Nattalli, line 19: can be int as well

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+            return self
+
+        self.km += other
+        return self
+
+    def __mul__(self, other: Distance | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
Nattalli, line 31: can't be Distance

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Distance | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float) -> bool:
+        return self.km >= other
```
Nattalli, line 46: can be int as well

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self
+
+    def __mul__(self, other: Distance | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | float) -> bool:
+        return self.km < other
```
Nattalli, line 34: why don't you check the type of `other`? It will not work correctly for Distance instance

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __gt__(self, other: Distance | float) -> bool:
+        return self.km > other
```
Nattalli, line 37: and here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __eq__(self, other: Distance | float) -> bool:
+        return self.km == other
```
Nattalli, line 40: and here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __le__(self, other: Distance | float) -> bool:
+        return self.km <= other
```
Nattalli, line 43: and here

app/main.py
Nattalli, line 46: and here

Nattalli (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km < other
+
+    def __gt__(self, other: int) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int) -> bool:
+        return self.km <= other
```
wQuelS, line 43: still not fixed. You have to check the type of `other`

wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+
+        self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
Serhii-Leonenko, line 27: `other` can be float

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+
+        self.km += other
+        return self
```
Serhii-Leonenko, line 25: Do not duplicate `return`

app/main.py
Serhii-Leonenko: Still not fixed. You should check the type of other, and implement the correct comparison. The same in other similar cases

Serhii-Leonenko (changes requested):

app/main.py
wQuelS: Not fixed...
`if isinstance(other, Distance)` - here you check if other is a type of `Distance`. You have to provide such logic here and in other similar methods.

wQuelS (changes requested):

YuriiKindrat (approved):

PR #369, opened 2023-02-06
Solution

WDemidenko:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if (isinstance(other, Distance)):
+            self.km = self.km + other.km
+        else:
+            self.km = self.km + other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
kuznetsb, line 26: You annotate that other might be Distance, but don't handle such occasion in code. I think it might be better or to remove Distance from type annotation or handle such case in code like raise some Error. And with  __truediv__ the same

kuznetsb (changes requested):

kuznetsb (approved):
Nice work!

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            self.km = self.km + other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km = self.km * other
+        return self
+
+    def __truediv__(self, other: int | float) -> Distance:
+        self.km = round((self.km / other), 2)
+        return self
```
YuriiKindrat, line 32: `other` can not be of type `Distance` in this method - make a check for this
also, this methods should return a new instance of Distance class

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Your solution is ok, but don't use redundant brackets in conditions

PR #370, opened 2023-02-06
Solution

SergiiMachulin:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> None:
```
usrofgh, line 14: You need to change return type annotation

usrofgh (changes requested):
Probably you forgot to change return type annotations in different methods

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int | float | Distance) -> None:
+        if isinstance(other, int | float):
+            self.km += other
+            return self
+        self.km += other.km
+        return self
```
YuriiKindrat, line 24: u can use here only 1 return

YuriiKindrat: Also are u sure, that this method reutrns None?

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
```
YuriiKindrat, line 27: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, int | float):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, int | float):
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 57: you can write these two methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):

wQuelS (approved):
Great

PR #372, opened 2023-02-06
Solution

NEkropo1:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return not self.__ge__(other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        self.km = round(self.km / other, 2)
+        return self
```
ada-krav, line 50: Looks good

ada-krav (approved):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, distance: int | float) -> None:
+        self.km = distance
+
+    @staticmethod
+    def is_int(other: int | float | Distance) -> bool:
```
Y-Havryliv, line 9: seems like is_int name is not good enough, you're also checking for float.

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km == other if self.is_int(other) else self.km == other.km
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return self.km > other if self.is_int(other) else self.km > other.km
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.km >= other if self.is_int(other) else self.km >= other.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
```
Y-Havryliv, line 43: it's not a good way to call magic methods, you don't add numbers using `__add__`, right?
it's better to use return `not self > other`

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return not self.__ge__(other)
```
Y-Havryliv, line 46: same here

Y-Havryliv (changes requested):

app/main.py
NEkropo1: In python u need only implement 2 equality methods for comparison functioning in normal(pythonic) way, but since we implemented Distance, which can't be <= 
image
it's exact default way for python, when function which checks something is overloaded, and u cant implement easy way checking it by not this method. But to pass tests I needed to implement those methods(since we invented class Distance), so `not other dunder` is great solution as for me.

app/main.py
NEkropo1: it's local function, implementing for checking if value is in int | float, yes, but overall idea is not_distance, 
while is_int_or_float and not_distance is either 2 long, or unnecessary for method name, while is_int is pretty intuitive.
I agree u could make even a little bit more descriptive, but I'm sure it's intuitive enough for understanding, what method does, so further refactoring is unnecessary, as for me.

app/main.py
NEkropo1: same here

app/main.py
NEkropo1: By the way I still disagree on this exact case. I mean since we already overloaded built-in magic for defining our own magic, it is pretty exact logic of python. I mean mundane | magic.
Really seems more logical, it's like we try to turn steering wheel, it doesn't work the way we want, and after replacing it on other wheel, we still hang out our old wheel on this wheel. If you understand the idea.
If the implementation is easy to explain, it may be a good idea )

app/main.py
Y-Havryliv: your code should be clear rather than intuitive

app/main.py
Y-Havryliv: https://www.tutorialsteacher.com/python/magic-methods-in-python
you can read this
btw, the last question, do you use `int.__str__(1)` or `str(1)`?

Y-Havryliv (approved):

app/main.py
Y-Havryliv: why don't you want to name it `is_distance` for example, is it bad for you?

app/main.py
NEkropo1: > https://www.tutorialsteacher.com/python/magic-methods-in-python you can read this btw, the last question, do you use `int.__str__(1)` or `str(1)`?

mostly I use __str__ if I want to see class logic which is more defined, or print(1) if it's builtin
I read that, and idc, there's no much suggestion either we should use magic methods inside or not, only suggestions to implement override if we need it:
image

As for me, it's some efficient encapsulation, like when we use lq, in core python gets to method, reads logic, after that goes to dunder method and applies him, while if we introduce that logic inside dunder, python will just use other defined dubder, which seems more efficient to me...
That's my overall idea in using functions, like if we don't need magic methods - better to use basic logic, but if we need magic, use other magic inside magic for encapsulation and efficiency.

app/main.py
NEkropo1: > why don't you want to name it `is_distance` for example, is it bad for you?

Yup it's great, similar implementation, more intuitive, more clear, idk, sometimes production matters, but for better practice I totally agree

PR #373, opened 2023-02-06
solution

amber-marichi:


alexlega (approved):

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -55,13 +55,7 @@ def __eq__(self, other: Distance | int | float) -> bool:
             return self.km == other.km
 
     def __le__(self, other: Distance | int | float) -> bool:
-        if isinstance(other, (int, float)):
-            return self.km <= other
-        elif isinstance(other, Distance):
-            return self.km <= other.km
+        return not self.__gt__(other)
```
Y-Havryliv, line 58: it's not a good way to call magic methods, you don't add numbers using __add__, right?
it's better to use `return not self > other`

app/main.py
```diff
@@ -55,13 +55,7 @@ def __eq__(self, other: Distance | int | float) -> bool:
...
 
     def __ge__(self, other: Distance | int | float) -> bool:
-        if isinstance(other, (int, float)):
-            return self.km >= other
-        elif isinstance(other, Distance):
-            return self.km >= other.km
+        return not self.__lt__(other)
```
Y-Havryliv, line 61: rewrite here as well

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return self
+
+    def __mul__(self, multiplier: int | float) -> Distance:
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, divider: int | float) -> Distance:
+        divided_distance = Distance(self.km)
+        if divider != 0:
+            divided_distance.km = round(divided_distance.km / divider, 2)
+        return divided_distance
```
YuriiKindrat, line 37: `multiplier/divider` can not be of type `Distance` in this method - make a check for this

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return self
+
+    def __mul__(self, multiplier: int) -> Distance:
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, divider: int) -> Distance:
+        divided_distance = Distance(self.km)
+        if divider != 0:
+            divided_distance.km = round(divided_distance.km / divider, 2)
+        return divided_distance
```
Serhii-Leonenko, line 37: No, it can be int | float. But it can't be an instance of Distance, so please add a check that  multiplier and divider are not instances of Distance

Serhii-Leonenko (changes requested):

Y-Havryliv (approved):
Nice!

PR #374, opened 2023-02-07
Solution DonchykD

DmytryTong:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if type(other) is Distance:
+            return Distance(
+                self.km + other.km
+            )
+        self.km += other
+        return self
```
oksana-feshchenko, line 14: I think when we are implementing _add_ we should return new instance

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = self.km * other
+        return self
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if type(other) == Distance:
+            Distance(
+                self.km / other.km
+            )
+        self.km = round((self.km / other), 2)
+        return self
```
oksana-feshchenko, line 39: i think here should be new instance too

oksana-feshchenko (changes requested):
iadd definitely  needs change, about truediv can't find correct info

app/main.py
DmytryTong: https://stackoverflow.com/questions/1047021/overriding-in-python-iadd-method   We do not return a new one, like in __add__, we overriting using "+=" or similar inside current (self) object.

app/main.py
DmytryTong: Thank you, fixing that)

inrock1 (approved):

inrock1 (approved):
.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 55: what if other will be a Distance class ? 
image

YuriiKindrat (changes requested):

app/main.py
DmytryTong: Done.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        self.km = self.km * other
+        return self
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if type(other) == Distance:
+            Distance(
+                self.km / other.km
+            )
+        self.km = round((self.km / other), 2)
+        return self
```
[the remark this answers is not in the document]
DmytryTong: Done.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if type(other) == Distance:
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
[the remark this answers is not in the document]
DmytryTong: Done.

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Serhii-Leonenko, line 28: Use only one return here

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, Distance):
+            print("Error: object other can not be of type Distance")
+        self.km = self.km * other
+        return self
```
Serhii-Leonenko, line 34: You should return a new instance of Distance here, not self.

Serhii-Leonenko (changes requested):

app/main.py
DmytryTong: Done.

app/main.py
DmytryTong: Done.

Y-Havryliv (approved):
Well done!

PR #375, opened 2023-02-07
Solution

viktoria-rybenchuk:


Danil228 (approved):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 32: `other` can not be of type `Distance` in this methods - make a check for this

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Distance | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 47: what if other will be a Distnace instance ? 
image

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __lt__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
```
wQuelS, line 43: esle is redundant here, since if line 41 is `True` it will end the method

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return self.km > other
+
+    def __eq__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
```
wQuelS, line 49: same, fix everywhere

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
wQuelS, line 62: you can write these two methods in one line each. Use previously created magic methods

wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __eq__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float) -> bool:
+        return not self > other
+
+    def __ge__(self, other: Distance | float) -> bool:
+        return not self < other
```
wQuelS, line 53: either add a check for a `Distance` instance here as well or use previously created magic methods

wQuelS (changes requested):

wQuelS (approved):
GJ

PR #379, opened 2023-02-07
Solution

inrock1:


oksana-feshchenko (approved):

bohdan-sain (approved):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+            self.km = self.km + other
+        else:
+            self.km = self.km + other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 30: `other` can not be of type `Distance` in this methods - make a check for this

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 45: What if other will be a Distance instance ? 
image

YuriiKindrat (changes requested):

app/main.py
inrock1: done

app/main.py
inrock1: added processing Distance instance.

Serhii-Leonenko (approved):
Good job!

PR #380, opened 2023-02-07
Solution

Danil228:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
```
oksana-feshchenko, line 33: If you could take Distance instance you should  implement this in the code

oksana-feshchenko (changes requested):

app/main.py
oksana-feshchenko: its about all methods that above this lines

Danil228: Done

oksana-feshchenko (approved):
Actually, im not sure if methods mull and truedivision dont take distance instance, so i will leave it for you to decide .

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
```
Yulia-Mala, line 41: you have the same check "if isinstance(other, Distance):" almost in all the methods. And then two returns. Don't you want to create additional method which return value you need (other if other is int or float or Distance.km if other is Distance)? You will need only one return ( for example self.km > your_method(other)). Such a way will make this code mush more laconic. If something is unclear - we can have a chat or call tomorrow and discuss it!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 31: `other` can not be of type `Distance` in this method - make a check for this

YuriiKindrat (changes requested):

Y-Havryliv (approved):
Well done!

app/main.py
inrock1: done

PR #381, opened 2023-02-07
'Solution'

RuslanLopatin:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        else:
+            return self.km == other
+
+    def __le__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: int) -> int:
```
oksana0410, line 54: return type bool

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        else:
+            return self.km > other
+
+    def __eq__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: int) -> int:
```
oksana0410, line 48: return type bool

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        else:
+            return self.km < other
+
+    def __gt__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
+
+    def __eq__(self, other: int) -> int:
```
oksana0410, line 42: return type bool

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __truediv__(self, other: int) -> int:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other: int) -> int:
```
oksana0410, line 36: return type bool

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> int:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> int:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> int:
```
oksana0410, line 30: return type bool

oksana0410 (approved):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
Serhii-Leonenko, line 15: Fix annotation, `other` can be not only int. Also, you can remove redundant else. Fix in all other similar cases

app/main.py
Serhii-Leonenko: agree, fix this everywhere in all similar cases

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __le__(self, other: bool) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: bool) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
wQuelS, line 53: you can use previously created magic methods to write those two in one line.
with such approach, u will not need type check, since the methods that u've created already have it

wQuelS (changes requested):

YuriiKindrat (changes requested):
are u sure about type annotation of "other" in all of ur methods ? Also take a look to type annotation of return in ur methods

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> bool:
```
YuriiKindrat, line 11: is it return bool?

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> bool:
```
YuriiKindrat, line 17: the same, is self a boolean value ?

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> bool:
+        return Distance(self.km * other)
```
YuriiKindrat, line 25: is distance a boolean type ?)

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __truediv__(self, other: int) -> bool:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 28: same

YuriiKindrat (changes requested):
Also check ur typr anntoation for other, u are writing that other is an int type and checking for Distance class below this annotation :)

wQuelS (approved):
Nice

PR #382, opened 2023-02-07
Solution

MikhailLyvak:


app/main.py
```diff
@@ -1,3 +1,46 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
yura-hudzovskyi, line 5: I think `int | float` is better

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | float) -> Distance:
+        return self.km == other
+
+    def __le__(self, other: Distance | float) -> Distance:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float) -> Distance:
+        return self.km >= other
```
yura-hudzovskyi, line 46: same `Distance | float | int`

yura-hudzovskyi (changes requested):

yura-hudzovskyi (approved):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+
+        self.km += other
+        return self
```
YuriiKindrat, line 25: u can make this using 1 return

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 31: `other` can not be of type `Distance` in this methods - make a check for this

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> Distance:
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> Distance:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> Distance:
+        return self.km >= other
```
YuriiKindrat, line 46: Also u can write le,ge methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        error = f"You cannot multiply {type(self)} by {type(other)}"
```
wQuelS, line 27: it is not needed here

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km < other
+
+    def __gt__(self, other: Distance | int | float) -> Distance:
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> Distance:
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> Distance:
+        return self.km <= other
```
wQuelS, line 45: You still don't type check in these methods. Read Yurii's comment

wQuelS (changes requested):

app/main.py
YuriiKindrat: u didn't fix the problem with lt/gt/eq/le/ge methods

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #383, opened 2023-02-07
Solution

BohdanLazaryshyn:


oksanaaam (approved):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            km=self.km + other.km
+        )
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
+            return self
+        else:
+            self.km += other.km
+            return self
```
YuriiKindrat, line 29: no need to use 2 returns here

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
```
YuriiKindrat, line 35: `other` can not be of type `Distance` in this method - make a check for this

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __le__(self, other: Distance) -> bool:
+        if not isinstance(other, Distance):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance) -> bool:
+        if not isinstance(other, Distance):
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 60: you can write these two methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):

wQuelS (changes requested):

wQuelS (approved):
Great

PR #386, opened 2023-02-07
'Solution'

ykopatko:


ostapT (approved):
GJ!

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self
+
+    def __mul__(self, other: [int, float, Distance]) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: [int, float, Distance]) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: [int, float, Distance]) -> bool:
+        return self.km < other
```
Nattalli, line 33: does it work good even if we send `Distance`?

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __gt__(self, other: [int, float, Distance]) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: [int, float, Distance]) -> bool:
+        return self.km == other
```
Nattalli, line 39: the same question for these methods

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __le__(self, other: [int, float, Distance]) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: [int, float, Distance]) -> bool:
+        return self.km >= other
```
Nattalli, line 45: and these

Nattalli: please, add a check for `other` type

Nattalli (changes requested):

app/main.py
```diff
@@ -11,35 +11,45 @@ def __str__(self) -> str:
...
             self.km += other
         return self
 
-    def __mul__(self, other: [int, float, Distance]) -> Distance:
+    def __mul__(self, other: int | float | Distance) -> Distance:
         return Distance(self.km * other)
 
-    def __truediv__(self, other: [int, float, Distance]) -> Distance:
+    def __truediv__(self, other: int | float | Distance) -> Distance:
         return Distance(round(self.km / other, 2))
```
Nattalli, line 30: btw, `__mul__ ` and `__truediv__ ` methods can't have a `Distance` type of `other` variable

Nattalli (changes requested):
good job, 1 comment left. Please, fix it!

app/main.py
ykopatko: done

Nattalli (approved):
Well done!

PR #387, opened 2023-02-07
'Solution'

NazarHladaniuk:


Anatolii-Poznyak (approved):
nice

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 58: you can write these two methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km = self.km + other
+        else:
+            self.km = self.km + other.km
+        return Distance(self.km)
+
+    def __iadd__(self, other: int | float) -> Distance:
```
Serhii-Leonenko, line 21: Fix annotations here, other can be Distance

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km *= other
+        return self
```
Serhii-Leonenko, line 30: Add a check that the other is not a Distance instance and the method should return a new instance not self

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= other
```
Serhii-Leonenko, line 49: Will these methods work if the other is Distance?

Serhii-Leonenko (changes requested):

wQuelS (approved):
GJ

PR #388, opened 2023-02-07
Solution

bohdan-sain:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise ArithmeticError("right operand must be int,"
+                                  " float or instance of the class")
```
AlessaQ, line 21: This conditions are not required cause we have just two stacking options - first of them you already described upper, another can be described using condition `else`, but better return it directly. Also you can print error description, but it is not necessary.

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
+        else:
+            raise ArithmeticError("right operand must be int,"
+                                  " float or instance of the class")
```
AlessaQ, line 32: Same like in `__add__`

AlessaQ (changes requested):

bohdan-sain:
like this?

ProdyRodion (approved):

PR #390, opened 2023-02-07
Solution

AlessaQ:


me1nyk (approved):
Well done)

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float):
```
Y-Havryliv, line 26: seems like other can't be `Distance`, right?

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float):
```
Y-Havryliv, line 29: same here

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float):
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __len__(self) -> int:
```
Y-Havryliv, line 57: can it be not only `int`?

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float):
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 48: you can use `return not self > other` here, but it's not necessary

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float):
+        if isinstance(other, Distance):
```
Y-Havryliv, line 53: you can use `return not self < other` here

Y-Havryliv (changes requested):

app/main.py
AlessaQ: Yes, cause if we will multiply `distance1 * distance2` we will receive square area. So after if we will print answer `"Distance: {self.km} kilometers."`,  statement `kilometers` will be wrong.

app/main.py
AlessaQ: Same here, statement `"Distance: {self.km} kilometers."` will be wrong

app/main.py
AlessaQ: changed

AlessaQ: changed

app/main.py
AlessaQ: changed

AlessaQ: changed

app/main.py
AlessaQ: changed

app/main.py
AlessaQ: changed

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float):
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float):
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 30: `other` can not be of type `Distance` in this method - add a check for this

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):
also all tests should pass, fix flake8 errors

app/main.py
AlessaQ: added

Serhii-Leonenko (changes requested):
Tests still not passed(

Serhii-Leonenko (changes requested):
Don't change tests. It is not the first time. 
Fix it

tests/test_main.py
```diff
@@ -428,4 +428,4 @@ def test_distance_class_le_number(kilometers, kilometers2, result):
     distance = Distance(kilometers)
     assert (distance <= kilometers2) is result, (
         f"'Distance({kilometers}) <= {kilometers2}' should equal to {result}"
-    )
+    )
```
wQuelS, line 431: you've done changes to test fail, but didn't add new line

wQuelS (changes requested):

wQuelS (approved):
Good

PR #391, opened 2023-02-07
Solution

TeliksPG:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if other is not Distance:
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if other is not Distance:
+            return self.km >= other
+        return self.km >= other.km
```
destr34, line 56: add newline at end of file

destr34 (changes requested):

app/main.py
ProdyRodion: @destr34 it does not show it

destr34 (approved):

app/main.py
ProdyRodion: @destr34 thanks

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int) -> Distance:
```
Y-Havryliv, line 19: you have `int | float` in init, so it can be float as well, fix everywhere, please

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #392, opened 2023-02-07
solution. 

Yulia-Mala:
I am curious whether it exist shorter approach to define __gt__ and __ge__ (we have __lt__ and __le__ but its matter do we have number and Distance on the left and right or vise versa).

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,44 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    @staticmethod
+    def get_operand(element: Distance | int | float) -> int | float:
+        return element.km if isinstance(element, Distance) else element
+
+    def __str__(self) -> str:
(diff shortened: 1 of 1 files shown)


VladyslavCherkashyn (approved):
It seems to me that in this approach (calling a magic method -> calls the check function -> calls isinstance()), the program's work increases, although it visually looks more concise.
But I'm not sure how much it matters in this case.

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km += self.get_operand(other)
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 29: `other` can not be of type `Distance` in this methods - make a check for this

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #394, opened 2023-02-07
Solution

Anatolii-Poznyak:


app/main.py
```diff
@@ -1,3 +1,66 @@
...
-    pass
+
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return "Distance(km=" + str(self.km) + ")"
```
wQuelS, line 13: it is almost always better to use f-strings instead of concatenation

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+            if isinstance(other, Distance)
+            else self.km <= other
+        )
+
+    def __ge__(self, other: [int, float] | Distance) -> bool:
+        return (
+            self.km >= other.km
+            if isinstance(other, Distance)
+            else self.km >= other
+        )
```
wQuelS, line 66: you can write these two methods in one line each. Use previously created magic methods

wQuelS (changes requested):

app/main.py
Anatolii-Poznyak: Nice idea, actually. thnx. I ddnt know we can do it that way

wQuelS (approved):
Very well

PR #396, opened 2023-02-07
Solution

VladyslavCherkashyn:


RomanRega (approved):
I think i have very similiar code , so i have not looked mistakes, you have good long strings, and flake8 say you yes.

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        self.km += other.km
+        return self
```
wQuelS, line 25: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
```
wQuelS, line 31: `other` can not be of type `Distance` in these methods

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        return self.km >= other.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
```
wQuelS, line 56: you can make these methods in one line using previously created magic methods

wQuelS (changes requested):

app/main.py
VladyslavCherkashyn: if I understood correctly, then the __ ge__ method couldn`t be implemented at all, simply because when checking for a >= b, the python will swap the values b <= a and return the right result. 
In the same way why we do not implement the method here __ne__ -> not (a == b)

app/main.py
```diff
@@ -46,11 +46,7 @@ def __lt__(self, other: Distance | int | float) -> bool:
         return self.km < other.km
 
     def __ge__(self, other: Distance | int | float) -> bool:
-        if isinstance(other, (int, float)):
-            return self.km >= other
-        return self.km >= other.km
+        return self == other or self > other
```
Y-Havryliv, line 49: return not self < other

app/main.py
```diff
@@ -46,11 +46,7 @@ def __lt__(self, other: Distance | int | float) -> bool:
...
 
     def __le__(self, other: Distance | int | float) -> bool:
-        if isinstance(other, (int, float)):
-            return self.km <= other
-        return self.km <= other.km
+        return self == other or self < other
```
Y-Havryliv, line 52: return not self > other

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv: it would be better

app/main.py
VladyslavCherkashyn: You are right, thanks.

Y-Havryliv (approved):
Nice!

PR #397, opened 2023-02-07
Solution

dogmoony:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, number: any) -> float:
+        if isinstance(number, Distance):
+            return Distance(self.km + number.km)
+        else:
```
AnnaBoichenko, line 17: you don`t have to write else after return. this block is not reachable if isinstance(number, Distance) return True

AnnaBoichenko: the same applies to each else block below

AnnaBoichenko: I would also recommend checking for type errors like:
self.km + "a string or another unsupportable type"

AnnaBoichenko (approved):

app/main.py
wQuelS: Yep, that is correct. Fix everywhere

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            return Distance(self.km + distance.km)
+
+    def __iadd__(self, number: any) -> float:
+        if isinstance(number, Distance):
+            self.km += number.km
+            return self
+        else:
+            distance = Distance(number)
+            self.km += distance.km
+            return self
```
wQuelS, line 28: As it was said above - no need for double `return` statements

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: int) -> float:
+        return Distance(round(self.km / number, 2))
+
+    def __lt__(self, number: int) -> bool:
+        return self.km < number
+
+    def __gt__(self, number: int) -> bool:
+        return self.km > number
```
wQuelS, line 40: What if `other` is of type `Distance` here and in all the following methods? Add type checks

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __iadd__(self, number: any) -> float:
+        if isinstance(number, Distance):
+            self.km += number.km
+            return self
+        else:
+            distance = Distance(number)
+            self.km += distance.km
+            return self
+
+    def __mul__(self, number: int) -> float:
```
wQuelS, line 30: by convention we call it `other` not `number`. Fix everywhere

wQuelS (changes requested):

Serhii-Leonenko (changes requested):
All checks should be passed. Fix it

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: any) -> float:
+        if Distance.__typecheck__(other):
+            return Distance(self.km + other.km)
+        distance = Distance(other)
```
YuriiKindrat, line 21: no need to make new variable, u can just use other

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __le__(self, other: any) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: any) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 61: try to implement this method using already implemented methods

dogmoony: I don`t know how to do that...

Nattalli: @dogmoony look:
`>=` is the same that `not <`
So for `__ge__` you can use `not __lt__`

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __eq__(self, other: any) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: any) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km <= other.km
+        return self.km <= other
+
```
Nattalli, line 57: the same here

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float | int) -> None:
+        self.km = km
+
+    @staticmethod
+    def __typecheck__(other: any) -> None:
```
Nattalli, line 9: please, don't use `any` type annotation. I recommend you use `Union[int, float, Distance]` or `int | float | Distance` (Idk your python version)

Nattalli: and fix it in every method!

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: any) -> float:
+        if Distance.__typecheck__(other):
+            return Distance(self.km + other.km)
+        other = Distance(other)
+        return Distance(self.km + other.km)
```
Nattalli, line 22: you can just `return Distance(self.km + other)` instead of `$21-22` lines. WDYT?

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __iadd__(self, other: any) -> float:
+        if Distance.__typecheck__(other):
+            self.km += other.km
+        else:
+            distance = Distance(other)
+            self.km += distance.km
```
Nattalli, line 29: and here

Nattalli: `self.km += other`

Nattalli (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> float:
+        if Distance.__typecheck__(other):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> float:
```
Nattalli, line 30: `other` can be float as well

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> float:
```
Nattalli, line 33: and here

Nattalli (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> float:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> float:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> bool:
```
Serhii-Leonenko, line 36: There is still an incorrect annotation, other not only `int`

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if self.__eq__(other):
+            return True
+        return not self.__le__(other)
```
wQuelS, line 59: do both of them in one line using previously created magic methods, using `not` keyword and properly chosen sign

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return not self.__lt__(other)
```
wQuelS, line 55: very good. Just use `<` here, since it will automatically call you brand new magic method `__lt__`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if Distance.__typecheck__(other):
+            return self.km <= other.km
+        return self.km <= other
```
wQuelS, line 52: implement approach that u've done in `__ge__` magic method

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return isinstance(other, Distance)
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> float:
+        if Distance.__typecheck__(other):
```
wQuelS, line 19: this can be done with `isinstance`. fix everywhere

wQuelS (changes requested):

app/main.py
dogmoony: I don't understand where to use <

YuriiKindrat: u can use < between self and other object

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __truediv__(self, other: int | float) -> float:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return not self.__le__(other)
```
YuriiKindrat, line 38: the same is for this method, no need to call method like this,` __<some method>__ ` u can use this methods as in `__ge__` method

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
YuriiKindrat, line 46: the same

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #399, opened 2023-02-08
Solution

StreetNik:


app/main.py
```diff
@@ -1,3 +1,55 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
vasylhnatiuk, line 5: If the number is a float, a` | float` should be added.

ArsenPidhoretskyi, line 5: float possible

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
```
vasylhnatiuk, line 27: `other` cannot be an instance of Distance in this method, please add checks for it

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
```
vasylhnatiuk, line 29: other cannot be an instance of Distance in this method, please add checks for it

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
```
vasylhnatiuk, line 49: to make your code simpler, you should use a magic method that has already been defined

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            self.km >= other.km
```
vasylhnatiuk, line 54: the same here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self < other or self == other
```
vasylhnatiuk, line 50: will be correct - `return self.__lt__(other) or self.__eq__(other)`, because you wrote methods which do it already

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self > other or self == other
```
vasylhnatiuk, line 53: `return self.__gt__(other) or self.__eq__(other)` here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -47,7 +47,7 @@ def __eq__(self, other: int | float | Distance) -> bool:
         return self.km == other
 
     def __le__(self, other: int | float | Distance) -> bool:
-        return self < other or self == other
+        return self.__lt__(other) or self.__eq__(other)
 
     def __ge__(self, other: int | float | Distance) -> bool:
-        return self > other or self == other
+        return self.__gt__(other) or self.__eq__()
```
vasylhnatiuk, line 53: must be `return self.__gt__(other) or self.__eq__(other)`

vasylhnatiuk (changes requested):

app/main.py
StreetNik: AssertionError: '__truediv__' method should not accept Distance instance
AssertionError: '__mul__' method should not accept Distance instance
Tests say it can not

vasylhnatiuk (approved):
Good job

PR #401, opened 2023-02-08
Solution

ostapT:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, number: int | float) -> Distance:
+        if isinstance(number, Distance):
+            return Distance(km=self.km + number.km)
+        else:
+            number = Distance(number)
+            return Distance(km=self.km + number.km)
```
myroslavatrschk, line 19: remove "km=" and redundant else. Also, why do you need this variable at line 18? You can just write `return Distance(self.km + number)` at line 19. https://github.com/mate-academy/py-car-wash-station/blob/master/checklist.md can help (search "Clean Code").

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __iadd__(self, number: int | float) -> Distance:
+        if isinstance(number, Distance):
+            self.km += number.km
+            return self
+        else:
+            number = Distance(number)
```
myroslavatrschk, line 26: Same problem with redundant variable "number".

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > number
+
+    def __eq__(self, number: int | float) -> bool:
+        return self.km == number
+
+    def __le__(self, number: int | float) -> bool:
+        return self.km <= number
+
+    def __ge__(self, number: int | float) -> bool:
+        return self.km >= number
```
myroslavatrschk, line 49: Add isinstance check for int | float in number.

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        else:
+            number = Distance(number)
+            self.km += number.km
+            return self
+
+    def __mul__(self, number: int | float) -> Distance:
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: int | float) -> Distance:
+        return Distance(km=round((self.km / number), 2))
```
myroslavatrschk, line 34: Same problem with "km=".

myroslavatrschk (changes requested):

oksanaaam (approved):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, number: int | float) -> bool:
+        if isinstance(number, Distance):
+            return self.km <= number.km
+        return self.km <= number
+
+    def __ge__(self, number: int | float) -> bool:
+        if isinstance(number, Distance):
+            return self.km >= number.km
+        return self.km >= number
```
wQuelS, line 55: you can do this in one line using previously created magic methods. This will also allow you to get rid of those type checks

wQuelS (changes requested):

YuriiKindrat (approved):

PR #402, opened 2023-02-08
Develop

NickSavchyn:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
skmiwa1, line 18: u can ignore else here!

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, (int, float)):
+            self.km += other.km
+            return self
+        else:
```
skmiwa1, line 25: same else

skmiwa1 (changes requested):

app/main.py
NickSavchyn: i can't, just tried to do that and it returned 10 mistakes
image

skmiwa1 (approved):
GJ

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, (int, float)):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Nattalli, line 25: don't repeat yourself: use `return` statement only 1 time in this method)

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
Nattalli, line 46: make these methods using already implemented `__lt__` and `__gt__` methods

Nattalli (changes requested):

app/main.py
wQuelS: Natalli meant that you already has implemented magic methods, for example `__lt__`.
So when you write `<` it will automatically call this brand new magic method. You can use such behaviour in `__ge__` and `__le__` methods. Just think about how to do it. Maybe some keywords, like `not`, can help u !?

wQuelS (changes requested):

wQuelS (approved):
3 last methods can be done a lot easier, but still valid implementation. GJ

PR #403, opened 2023-02-08
Solution

tytomyr:


app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        else:
+            return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+        else:
+            self.km = self.km + other
+            return self
```
wQuelS, line 24: you duplicating your returns

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km > other
+
+    def __lt__(self, other: int | float) -> bool:
+        return self.km < other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
```
wQuelS, line 51: What if other is `Distance` type? 

Also, implement `__le__` and `__ge__` methods using previously created magic methods. With such approach you also can skip type checking

wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int | float | Distance) -> (int, float):
+        self.km = km
```
Serhii-Leonenko, line 7: Are you sure that `km` can be Distance and the method returns int,float?)

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+        else:
+            self.km = self.km + other
+        return self
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(km=round((self.km / other), 2))
```
Serhii-Leonenko, line 25: other cannot be Distance, fix annotation

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(km=self.km * other)
```
Serhii-Leonenko, line 31: other cannot be Distance, fix annotation

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km > other
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.km > other or self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self.km < other or self.km == other
```
Serhii-Leonenko, line 49: What if `other` is of type `Distance` here and in all the following methods?

Serhii-Leonenko (changes requested):

app/main.py
Nattalli: no, type annotations were OK. 
You had to add checking for type, as it is in the `__add__` method. So `other` can be a Distance, int, or float types, and your code should work in every case

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __gt__(self, other: Distance) -> bool:
+        return self.km > other
+
+    def __lt__(self, other: Distance) -> bool:
+        return self.km < other
+
+    def __ge__(self, other: Distance) -> bool:
+        return self.km > other or self.km == other
+
```
Nattalli, line 47: you can combine these two checking in one!

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __le__(self, other: Distance) -> bool:
+        return self.km < other or self.km == other
```
Nattalli, line 49: the same here

Nattalli (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #404, opened 2023-02-08
Solution

RomanRega:


app/main.py
```diff
@@ -1,3 +1,44 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (int, float)) -> (int, float):
```
Nattalli, line 2: does it really returns `(int, float)`?

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float)) -> (int, float):
```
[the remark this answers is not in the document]
Nattalli: and does it really returns int or float? What is the type of the `self`?

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self
+
+    def __iadd__(self, other: (int, float)) -> (int, float):
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: (int, float)) -> float:
```
Nattalli, line 25: it doesn't return float

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return Distance(km=round(self.km * other, 3))
+
+    def __truediv__(self, other: (int, float)) -> float:
+        return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: (int, float)) -> bool:
+        return self.km < other
```
Nattalli, line 32: what if `other` has a `Distance` type? You should add a checking of the `other` type

Nattalli (changes requested):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: (int, float)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
YuriiKindrat, line 24: no need to use 2 returns here, u can use "else" instead

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __eq__(self, other: (int, float)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: (int, float)) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: (int, float)) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 45: implement type check for this methods, what if other will be an instnace of Distance ? it will not work correctly

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __eq__(self, other: [Distance | float | int]) -> bool:
+        return self.km == other
+
+    def __le__(self, other: [Distance | float | int]) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: [Distance | float | int]) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 45: add type checking for all of this methods, other can be a Distance instance

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #405, opened 2023-02-08
solutionis

skmiwa1:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return f"Distance: {str(self.km)} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
oigod, line 19: u dont need use isinstance twice

oigod (changes requested):

oigod (approved):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {str(self.km)} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
YuriiKindrat, line 18: no need to use else here

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: [int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
YuriiKindrat, line 27: u can use only one return here

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __mul__(self, other: [int, float, Distance]) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: [int, float, Distance]) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 33: other can't be a instance of distance class in this methods

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km > other
+
+    def __eq__(self, other: [int, float, Distance]) -> bool:
+        return self.km == other
+
+    def __le__(self, other: [int, float, Distance]) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: [int, float, Distance]) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 48: will it work with if other will be an instance of distance?
image

YuriiKindrat (changes requested):

app/main.py
skmiwa1: usualy no, but isn`t it an ability of Magic metods to do that (compare instances to a int(ex.)???

skmiwa1: so, i concider it will work!

app/main.py
YuriiKindrat: fix type annotation here, Distance can't be an other

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: [int, float]) -> bool:
+        return self.km == other
+
+    def __le__(self, other: [int, float]) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: [int, float]) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 46: this methods should work with Distnace instance, so add a check for instance class

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: [int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: [int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
Nattalli, line 56: why did you remove `Distance` from type annotations?

Nattalli, line 56: implement these methods with usage of `__lt__` and `__gt__` methods

app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: [float, int]) -> any:
```
Nattalli, line 6: please, don't use `any` yupe annotation. Here returns nothing: `-> None`. Fix it everywhere -- remove `any` type annotation

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {str(self.km)} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [int, float]) -> Distance:
```
Nattalli, line 15: `other` can be a `Distance` as well - add it in type annotations

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: [int, float]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Nattalli, line 25: don't repeat yourself: you can use `return` statement at once for this ,ethod

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [int, float]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: [int, float]) -> Distance:
```
Nattalli, line 20: add `Distance` to type annotations

Nattalli (changes requested):

app/main.py
skmiwa1: я не розумію як це зробити...
            self.km += other
        else:
            self.km += other.km
        return self
    (якщо так то я не змінюватиму. попередні коментарі були про "позбудься else"
    якщо інакше то хотілося б посидання на ідею або приклад! дякую

Nattalli: hm, Okay, agree, can skip this comment

app/main.py
skmiwa1: going to qna to solve this(out of ideas)
or do you mean (self.__lt__(other) or self.__eq__(other)???

wQuelS (approved):
GJ

PR #410, opened 2023-02-09
Solution

myroslavatrschk:


myroslavatrschk (commented):
Removed second return in __iadd__ method and added isinstance check in comparing magic methods.

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
manjustice, line 30: Same

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __gt__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
```
manjustice, line 42: same

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __lt__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
manjustice, line 48: same

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
manjustice, line 54: same

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
```
manjustice, line 60: same

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
```
manjustice, line 27: Here you must check if other is an int or a float.

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
```
manjustice, line 36: redundant else

manjustice (changes requested):

manjustice (approved):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: [int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
Nattalli, line 14: `other` variable can be a float or int as well. PLease, fix type annotations here and everywhere!

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
Nattalli, line 57: make these methods using previously implemented methods `__lt__` and `__gt__`

Nattalli (changes requested):

wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self.km == other.km
+        return self.km == other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
```
Serhii-Leonenko, line 45: What if `other` is of type `Distance` here?

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
Serhii-Leonenko, line 51: it's ok but you can try to make the same more simplify )

Serhii-Leonenko (changes requested):

myroslavatrschk (commented):
I don't know how to simplify code in __ge__ and __le__. Using magic methods __gt__/__lt__ or __eq__ was simple enough.

wQuelS (approved):
You solution is okay, but simplify means that you could use just `>` when reusing `__gt__` magic method, since `>` will call this magic method under the hood

PR #413, opened 2023-02-10
Solution

tarasavage:


app/main.py
```diff
@@ -1,3 +1,44 @@
...
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    @staticmethod
+    def get_km(distance: Distance | int | float) -> int | float:
+        return distance.km if isinstance(distance, Distance) else distance
```
vasylhnatiuk, line 10: this is interesting solution

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return Distance.get_km(self) < Distance.get_km(other)
```
vasylhnatiuk, line 32: you must use `self.km` instead of `Distance.get_km(self)` and to use `get_km()` form instance: `self..get_km(other)` instead of `Distance.get_km(other)`

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        return Distance.get_km(self) > Distance.get_km(other)
```
vasylhnatiuk, line 35: the same here and in all of next methods

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Well done!

PR #417, opened 2023-02-13
Solution

V-Shkrobatskyi:


Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
     # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 6: float is possible

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: Not fixed

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi:
In all bool methods float also possible

katevasilenko (approved):
Well done

PR #420, opened 2023-02-22
Solution

YegorT:


app/main.py
```diff
@@ -1,3 +1,69 @@
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
ArsenPidhoretskyi, line 5: int also possible

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        self.km = km
+
+    def __str__(self) -> Any:
```
ArsenPidhoretskyi, line 8: No need to use Any if it is return correct type
Fix in all places

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+    def __lt__(self, other: Any) -> Any:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif type(other) is float or type(other) is int:
+            return self.km < other
+
+    def __gt__(self, other: Any) -> Any:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif type(other) is float or type(other) is int:
```
ArsenPidhoretskyi, line 50: no need to this elif

ArsenPidhoretskyi: the same in other places

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Any:
+        return Distance(
+            km=self.km * other
+        )
+
+    def __truediv__(self, other: int) -> Any:
```
ArsenPidhoretskyi, line 36: float also

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+            )
+
+    def __iadd__(self, other: Any) -> Any:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif type(other) is float or type(other) is int:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Any:
```
ArsenPidhoretskyi, line 31: float also

ArsenPidhoretskyi (changes requested):
Any type is not accepted for this task!

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(
+                km=self.km + other.km
+            )
+        if type(other) is float or type(other) is int:
```
ArsenPidhoretskyi, line 20: No need to check in other cases
We already know that if it is not Distance it must be int of float
Remove this check!

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+    def __truediv__(self, other: Union[int, float]) -> Distance:
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: Union[int, float, Distance]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
ArsenPidhoretskyi, line 46: else: not needed 
```suggestion
        return self.km < other
```

ArsenPidhoretskyi: The same in other places

ArsenPidhoretskyi (changes requested):

olena-hul:
Good job!

PR #422, opened 2023-02-25
Solution

AngelinaRemark:


app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
```
ArsenPidhoretskyi, line 6: int possible also

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
```
ArsenPidhoretskyi, line 15: int possible also
the same in other places

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float) -> float:
```
ArsenPidhoretskyi, line 20: It returns Distance

ArsenPidhoretskyi (changes requested):

Abnormaltype (approved):

PR #423, opened 2023-02-26
Solution

jinjo0222988:


app/main.py
```diff
@@ -1,3 +1,66 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: any) -> None:
```
Abnormaltype, line 5: put the correct type annotation

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: any) -> Distance:
```
Abnormaltype, line 14: put the correct type annotation

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: any) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
Abnormaltype, line 66: put the correct type annotation

Abnormaltype (changes requested):

app/main.py
jinjo0222988: I dont'n understand why the annotation type is wrong?

jinjo0222988: I don't understand why annotations type is wrong?

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __le__(self, other: any) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: any) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
```
vsmutok, line 65: remove redundant  else

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+            return self.km == other
+
+    def __le__(self, other: any) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: any) -> bool:
```
vsmutok, line 62:
```suggestion
    def __ge__(self, other: Distance | int | float) -> bool:
```

vsmutok: fix everywhere

app/main.py
vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
ArsenPidhoretskyi: not fixed it can be float | int

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+            self.km += other
+            return self
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 28: move return after if-else

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
```
ArsenPidhoretskyi, line 15: add must create new instance instead of change current

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        self.km = self.km * other
```
ArsenPidhoretskyi, line 31: mul must create new instance instead of change current

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        return self
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        self.km = round(self.km / other, 2)
```
ArsenPidhoretskyi, line 35: truediv must create new instance instead of change current

app/main.py
ArsenPidhoretskyi: not fixed

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(km)
+        km = self.km + other
+        return Distance(km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Yurnerroo, line 26: make 1 return

Yurnerroo (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            km = self.km + other.km
+            return Distance(km)
+        km = self.km + other
+        return Distance(km)
```
Yurnerroo, line 19: make 1 return here as well

Yurnerroo (changes requested):
checks didn't pass

Yurnerroo (approved):

PR #424, opened 2023-02-26
'Solution'

sasha-semenenko:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance) -> bool:
```
Bohdan-Salamakha, line 53: according to ur logic other can be not only Distance but also int and float
add related annotations

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
+
+    def __gt__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other: Distance) -> bool:
```
Bohdan-Salamakha, line 43: according to ur logic other can be not only Distance but also int and float
add related annotations

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __truediv__(self, number: int | float) -> Distance:
+        return Distance(round((self.km / number), 2))
+
+    def __lt__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
+
+    def __gt__(self, other: Distance) -> bool:
```
Bohdan-Salamakha, line 38: according to ur logic other can be not only Distance but also int and float
add related annotations

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other.km
+        return self
+
+    def __mul__(self, number: int | float) -> Distance:
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: int | float) -> Distance:
+        return Distance(round((self.km / number), 2))
+
+    def __lt__(self, other: Distance) -> bool:
```
Bohdan-Salamakha, line 33: according to ur logic other can be not only Distance but also int and float
add related annotations

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #429, opened 2023-02-28
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1,3 +1,42 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> object:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> object:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> object:
+        return self.km >= other
```
Grayder0152, line 42: What if `other` is `Distance`?

Grayder0152 (changes requested):

app/main.py
HalynaPetrova: Add a condition "if isinstance" as in __add__ and __iadd__ .  At first I had it in all the code, but then I decided to make the code more dry, pytest was satisfactory.

Grayder0152 (approved):

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> object:
```
Grayder0152, line 11:
```suggestion
    def __add__(self, other: int | float | 'Distance') -> object:
```

Grayder0152 (approved):

PR #430, opened 2023-03-01
solution

antonpiddubnyi:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        elif isinstance(other, (int, float)):
+            return Distance(km=self.km + other)
+
+    def __iadd__(self, other: (Distance, float)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (float, int)):
+            self.km += other
+            return self
```
Vasyl-Smutok, line 26: `DRY` 
here and 23 line

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(km=round((self.km / other), 2))
+
+    def __lt__(self, other: int) -> bool:
```
Vasyl-Smutok, line 34: `other` also can be `float` or `Distance`

please fix it here and other places

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: (int, float, Distance)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: (int, float, Distance)) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: (int, float, Distance)) -> bool:
+        return self.km >= other
```
Vasyl-Smutok, line 46: please add the same check as in `__add__`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, Distance)) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        elif isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 17: remove this check it is no needed

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            return Distance(km=self.km + other)
+
+    def __iadd__(self, other: (int, float, Distance)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (float, int)):
```
ArsenPidhoretskyi, line 23: just else without condition

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: (int, float, Distance)) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(km=self.km * other)
+
+    def __truediv__(self, other: (int, float, Distance)) -> Distance:
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 32: truediv must only work with int and float

ArsenPidhoretskyi: the same with mul
than no need to check for isinstance in this cases

ArsenPidhoretskyi (changes requested):

app/main.py
antonpiddubnyi: Should I put else: on line 17 instead of elif?

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, Distance)) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        else:
```
katevasilenko, line 17: Please remove redundant else

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: (int, float, Distance)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: (int, float, Distance)) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: (int, float, Distance)) -> bool:
+        return self.km >= other
```
katevasilenko, line 46: Please add check as in __add__ method

katevasilenko (changes requested):

app/main.py
antonpiddubnyi: @katevasilenko do you mean to add _if isinstance(other, Distance)_

ArsenPidhoretskyi (approved):
Well done!
but in this cases:
```
if something:
    return 1
else:
    return 2

```
else no need to use

PR #449, opened 2023-03-04
Solution

Kavbian:


app/main.py
```diff
@@ -1,3 +1,51 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
ArsenPidhoretskyi, line 5: int also possible

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance) -> bool:
```
ArsenPidhoretskyi, line 38: other also can be int or float
BTW, no need to change code inside function, just change typing
The same in other boolean method compare

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return self.km > other
```
Vasyl-Smutok, line 42: if other has type is Distance you need write other.km

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        final = None
+
+        if isinstance(other, Distance):
+            final = self.km > other.km
+        else:
+            final = self.km > other
+
+        return final
```
Vasyl-Smutok, line 49: Please fix it here and other places 

```suggestion
       if isinstance(other, Distance):
             return self.km > other.km
         return self.km > other
```

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #453, opened 2023-03-08
Solution

AlexandrZhezheria:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __init__(self, km: float) -> float:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" or float) -> "Distance":
```
ArsenPidhoretskyi, line 11: 1. Add import and remove quotes in all places
2. It also accepts int
3. Not need to use or.. Use Union[comma separated types] or | separed(int | float | str)

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 14: No need to check this. Because code must work only in two cases:

and second is another case when it other types. No need to check for int and float

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance" or float) -> "Distance":
```
ArsenPidhoretskyi, line 17: int possible

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> "Distance":
```
ArsenPidhoretskyi, line 24: float possible

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> "Distance":
```
ArsenPidhoretskyi, line 27: float possible

app/main.py
ArsenPidhoretskyi: The same in all boolean methods

ArsenPidhoretskyi (changes requested):

Grayder0152 (changes requested):
pls resolve comments

app/main.py
ArsenPidhoretskyi: @AlexandrZhezheria not fixed

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
```
ArsenPidhoretskyi, line 37: No need elif here use just return
The same in other places

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
katevasilenko, line 14:
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
katevasilenko, line 18:
```suggestion
        if isinstance(other, Distance):
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```

katevasilenko (changes requested):

app/main.py
AlexandrZhezheria: "Distance" без кавичок визива помилку, оставив поки що так як було

katevasilenko: Use `from __future__ import annotations ` to avoid this problem

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: float) -> Distance:
```
ArsenPidhoretskyi, line 28: int also possible

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> Distance:
```
ArsenPidhoretskyi, line 31: int also possible

Grayder0152 (approved):

PR #455, opened 2023-03-11
Solution

nickkozlov90:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int or float or Distance) -> Distance:
```
olena-hul, line 14:
```suggestion
    def __add__(self, other: Union[int, float, Distance]) -> Distance:
```

olena-hul, line 14: or
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if type(other) == int or type(other) == float:
```
olena-hul, line 15: Use isinstance(other, <type>) instead of type(other)

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
+from __future__ import annotations
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[int, float]) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float]) -> Distance:
+        if other != 0:
+            self.km = round(self.km / other, 2)
+            return self
```
ArsenPidhoretskyi, line 34: truediv no need to affect self, it must return new instance

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #457, opened 2023-03-12
Develop

sbilikepy:


app/main.py
```diff
@@ -1,3 +1,64 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self: "Distance", km: Union[int, float]) -> None:
```
vasylhnatiuk, line 6: don't write `self: "Distance"` . it is redundant

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        self.km = km
+
+    def __str__(self: "Distance") -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self: "Distance") -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]
```
vasylhnatiuk, line 15: write `Distance` type without `""`

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+                ) -> [int, float, "Distance"]:
```
vasylhnatiuk, line 16: your function always returns only Distance type. Change this annotation

vasylhnatiuk (changes requested):
Follow my advice and check every line of code.

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other: Union[int, float, Distance]
+                 ) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: Distance) -> Distance:
```
ArsenPidhoretskyi, line 31: other is not distance the same with truediv

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #462, opened 2023-03-15
'Solution'

kirzenko:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(other * self.km)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
vasylhnatiuk, line 35: what if `other` will be Distance type? add a check for this and all of functions

vasylhnatiuk (changes requested):

PR #463, opened 2023-03-15
Solution

bezhevets:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
```
vasylhnatiuk, line 32: what if `other` will be Distance type? add a check for this and all of functions below

vasylhnatiuk (changes requested):

app/main.py
bezhevets: Code works for Distance type and others types (int and float).

Grayder0152 (approved):

PR #467, opened 2023-03-16
Solution

DmytroBondariev:


app/main.py
```diff
@@ -1,3 +1,57 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Vasyl-Smutok, line 5: `float` possible

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Vasyl-Smutok, line 24: `DRY`
22 and 24 lines

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(
+            km=self.km * other
+        )
```
Vasyl-Smutok, line 29:
```suggestion
        return Distance(km=self.km * other)
```

Vasyl-Smutok (changes requested):

app/main.py
Dmytry95: if i make method look like this, i don`t pass several tests   
```
            self.km += other.km
        self.km += other
        return self
```

Grayder0152 (approved):

PR #471, opened 2023-03-21
solution

VadymRabotiuk:


app/main.py
```diff
@@ -1,3 +1,66 @@
-class Distance:
-    # Write your code here
-    pass
+class Distance:
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __add__(self, other: callable) -> callable:
```
vasylhnatiuk, line 5: other is not callable, other can be int, float or Distance, and you return Distance. Change here and other places

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+
+        if not isinstance(other, Distance):
+            return Distance(self.km + other.real)
+
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: callable) -> callable:
+
+        if not isinstance(other, Distance):
+            self.km += other.real
```
vasylhnatiuk, line 15: Why do you use real?

vasylhnatiuk (changes requested):

app/main.py
VadymRabotiuk: Because I think that we use value as a complex number and use first 'real' part of this number. Now I changed that and use only "other", thanks

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+
+        if not isinstance(other, Distance):
+            self.km += other
+            return self
+
+        self.km += other.km
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
ArsenPidhoretskyi, line 24: float also possible

ArsenPidhoretskyi (approved):
Good job!

PR #475, opened 2023-03-23
distance_class

Vperehrest:


app/main.py
```diff
@@ -1,3 +1,44 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
katevasilenko, line 2: Km can be int as well

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> float:
```
katevasilenko, line 11: Other can be int or Distance, not only float. And method returns object of Distance class, not float

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        if isinstance(other, Distance) is True:
+            return Distance(km=self.km + other.km)
+        else:
```
katevasilenko, line 14: Else is redundant

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+            self.km += other
+            return self
+
+    def __mul__(self, other: float) -> float:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: float) -> float:
+        return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: float) -> float:
```
katevasilenko, line 31: Other can be int or Distance here too, and method returns boolean value

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km < other
+
+    def __gt__(self, other: float) -> float:
```
katevasilenko, line 34: Other can be int or Distance here too, and method returns boolean value

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km > other
+
+    def __eq__(self, other: float) -> float:
```
katevasilenko, line 37: Other can be int or Distance here too, and method returns boolean value

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km == other
+
+    def __le__(self, other: float) -> float:
```
katevasilenko, line 40: Other can be int or Distance here too, and method returns boolean value

katevasilenko (changes requested):

app/main.py
Vperehrest: як тут без else написати?

PR #478, opened 2023-03-26
Solution

Munkoor:


app/main.py
```diff
@@ -1,3 +1,71 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(
+                km=self.km + other
+            )
+        else:
```
ArsenPidhoretskyi, line 19: else not needed here

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+        else:
+            self.km += other.km
+
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 33: No need check here. You say that your code will only work with int and float. It is accepted

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        else:
```
ArsenPidhoretskyi, line 46: else: not needed here
Check in other places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #481, opened 2023-03-29
Solution distance class

vborovets0:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+
+        self.km += other
+        return self
```
olena-hul, line 26: Use `else` here. This way you'll avoid using `return` 2 times

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError
```
Bohdan-Salamakha, line 29: u don't need to raise TypeError
maybe return Distance(self.km * other.km)?

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError
```
Bohdan-Salamakha, line 34: u don't need to raise TypeError
maybe return Distance(self.km * other.km)?

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #484, opened 2023-04-07
Distance solution

Filllick:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, callable]) -> callable:
```
```suggestion
    def __add__(self, other: Union[int, Distance]) -> Distance:
```

vasylhnatiuk: change here and all of the code if method returns Distance instance

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job!

PR #486, opened 2023-04-08
'Solution'

denysbelei9:


README.md
```diff
@@ -7,7 +7,7 @@ Every day you have morning running. You want to store your result.
 For easier calculating, comparison and printing your result write class
 `Distance`.
 
-`Distance`'s `__init__` method takes only
+`Distance`"s `__init__` method takes only
```
ArsenPidhoretskyi, line 10: Undo changes here

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
ArsenPidhoretskyi, line 11: other float and int is possible

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            return NotImplemented
```
ArsenPidhoretskyi, line 17: No need this. Because we specify that function must work with int and float if others type is not our problem
```suggestion
        return Distance(self.km + other)
```

ArsenPidhoretskyi: Fix in other places

app/main.py
ArsenPidhoretskyi: Fix typing in other places

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            return NotImplemented
+        return self
+
+    def __mul__(self, other: "Distance") -> "Distance":
```
ArsenPidhoretskyi, line 28: other can not be Distance
the same in truediv

tests/test_main.py
ArsenPidhoretskyi, line 1: Restore this file no need to change this

ArsenPidhoretskyi (changes requested):
Please no need to change readme and test_main. Also fix type annotations

PR #487, opened 2023-04-08
Create initial implementation of the module

barbequeue:


app/main.py
```diff
@@ -1,3 +1,76 @@
...
+    def __truediv__(self, distance2: int | float) -> Distance:
+        return Distance(round(self.km / distance2, 2))
+
+    def __eq__(self, distance2: Distance | int | float) -> bool:
+        if isinstance(distance2, (int, float)):
+            return self.km == distance2
+        elif isinstance(distance2, Distance):
+            return self.km == distance2.km
+        else:
+            raise TypeError
```
barbequeue, line 44: For some reason, when i define `__cmp__` method, i'm getting:
```
TypeError: '<' not supported between instances of 'Number' and 'Number'
```

There is so much repetitive code here, which could be done by just defining the `__cmp__` method

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}: {self.km} kilometers."
```
ArsenPidhoretskyi, line 9: No need to do like this just write this word: Distance

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(km={self.km})"
+
+    def __add__(self, distance2: Distance | int | float) -> Distance:
```
ArsenPidhoretskyi, line 14:
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

ArsenPidhoretskyi: Because it not always distance

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+        if isinstance(distance2, (int, float)):
+            return Distance(self.km + distance2)
+        elif isinstance(distance2, Distance):
+            return Distance(self.km + distance2.km)
+        else:
+            raise TypeError
```
ArsenPidhoretskyi, line 20: This cheks are not needed. You specify that your code works only with int, float and Distance
```suggestion
        return Distance(self.km + distance2.km)
```

ArsenPidhoretskyi: Fin in all places

app/main.py
ArsenPidhoretskyi: Rename in all places

app/main.py
ArsenPidhoretskyi: It is okay, define here this method which specified by task

ArsenPidhoretskyi (changes requested):

app/main.py
barbequeue: But isn't it a good practice to add some error handling? I can still pass a string or a list as an argument and get an error, like:
```
TypeError: unsupported operand type(s) for +: 'float' and 'str'
```

And the cause of this problem is not really obvious

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+        elif isinstance(other, Distance):
+            self.km += other.km
+        else:
+            raise TypeError
```
Bohdan-Salamakha, line 25: u don't need to raise a TypeError by yourself

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #495, opened 2023-04-20
Solution

OlehOryshchuk:


app/main.py
```diff
@@ -1,3 +1,66 @@
+from __future__ import annotations
+from __future__ import division
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    Validator = Union["Distance", int]
```
EdAlekseiev, line 7: `Validator` is not a clear name for this Union. Also, variable can be float

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    @staticmethod
+    def is_a_instance(other: Validator) -> None:
+        if isinstance(other, Distance):
+            return None
```
EdAlekseiev, line 33: What does do this method? It always returns `None`.

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+
+    def __mul__(self, other: Validator) -> Distance:
```
EdAlekseiev, line 35: Other can't be Distance

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        self.is_a_instance(other)
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: Validator) -> Distance | None:
```
EdAlekseiev, line 39: Other can't be Distance

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #500, opened 2023-04-25
'Solution'

MarinaKhrunina:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: ("Distance", int | float)) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: ("Distance", int | float)) -> "Distance":
```
vasylhnatiuk, line 23:
```suggestion
    def __mul__(self, other: (int | float)) -> "Distance":
```

vasylhnatiuk: `__mul__` can not get a Distance instance, delete from annotation

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return Distance((self.km * other))
+
+    def __truediv__(self, other: ("Distance", int | float)) -> "Distance":
```
vasylhnatiuk, line 26: `__truediv__` can not get a Distance instance, delete from annotation

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #501, opened 2023-04-25
Solution

vkrasnovyd:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        return Distance(self.km + other.km) \
+            if isinstance(other, Distance) \
+            else Distance(self.km + other)
```
olena-hul, line 17: It's better to use standard `if` instead of ternary operator if it doesn't fit into 1 line

Fix everywhere

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #505, opened 2023-04-30
'Solution'

ilizh:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> None:
```
olena-hul, line 11: Fix annotations everywhere: `other` can be Distance, int or float

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
olena-hul, line 14: `else` is not needed, you can just leave `return` here

olena-hul: Fix everywhere

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[float, int, "Distance"]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[float, int]) -> "Distance":
```
Bohdan-Salamakha, line 26: other can be Distance too

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[float, int]) -> "Distance":
```
Bohdan-Salamakha, line 29: other can be Distance too

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #507, opened 2023-05-02
Solution

z1k01ff:


app/main.py
```diff
@@ -1,3 +1,42 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: object) -> None:
```
ArsenPidhoretskyi, line 2: km is not object it can be int or float

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> object:
```
ArsenPidhoretskyi, line 11: `other: object` and `-> object` 
Is not accepted as type annotations in this task. Please specify the correct type annotations

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #516, opened 2023-05-11
Created class Distance with some magic methods

N1khto:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            return self.km > other
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
ArsenPidhoretskyi, line 54: You can just do 
```suggestion
        return self > other
```
Debug how it is works. It works so cool)
The same with le

vasylhnatiuk (approved):

PR #517, opened 2023-05-13
solution

uzlss:
I forgot how to create short-in annotation for Distance, int, float and mate.academy/learn was dead and I have no clue how to google this request.... may be im just freaking out over....

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,57 @@
+from __future__ import annotations
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        return Distance(
+            self.km + self.number_return(other)
+        )
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km += self.number_return(other)
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(
+                self.km * self.number_return(other)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(
+                self.km * self.number_return(other)
+            )
+
+    def __truediv__(self, other: float | int) -> Distance:
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 32: This check is not needed
The same in mul
Because we already say that we expect only float or int

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return self.km == self.number_return(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= self.number_return(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= self.number_return(other)
+
+    @staticmethod
+    def number_return(number: Any) -> int | float:
```
ArsenPidhoretskyi, line 53: It also can return None
when if return false and elif return false

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi:
You can do
OtherInputType = Union[float, int, Distance]
And then just write
other: OtherInputType

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    @staticmethod
+    def number_return(number: Any) -> int | float | None:
+        if isinstance(number, Distance):
+            return number.km
+        elif isinstance(number, (int, float)):
+            return number
+
+
+OtherInputType = Union[float, int, Distance]
```
vasylhnatiuk, line 59: write it before Distance, not after

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Good 👍

PR #518, opened 2023-05-14
solution

IvanKorshunovE:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
Vasyl-Smutok, line 14: fix it and other places 

Article about this https://stackoverflow.com/questions/44664040/type-hints-with-user-defined-classes
```suggestion
    def __add__(self, other: Union[int, float, Distance]) -> Type[Distance]:
```

Vasyl-Smutok (changes requested):

app/main.py
IvanKorshunovE: Screenshot 2023-05-14 at 20 46 51
Hi. While following your instructions I have PyCharm error: Expected type 'Type[Distance]', got 'Distance' instead in the __add__ function. 
Can I use the method described below instead? 

Vasyl-Smutok (approved):
Great 👍

PR #522, opened 2023-05-17
Distance class implementation

thorium-zero:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __radd__(self, rho: Any) -> Distance:
+        return self.__add__(rho)
+
+    def __add__(self, rho: Any) -> Distance:
+        return Distance(self.km + float(rho))
+
+    def __iadd__(self, rho: Any) -> Distance:
+        self.km += float(rho)
+        return self
```
YuriiKindrat, line 26: are u sure about Any for type annotation ? could we get float f.e. from dict ?) 
also what is "rho" ?) fix this everywhere, please

YuriiKindrat (changes requested):

app/main.py
thorium-zero: rho stands for 'right-hand object", it's a common way of parameter naming when overriding a binary operator ( [left-hand object] [operator] [right-hand object] ). So, when it takes making a method (C# or python for example) to override an operator, you have 'this' ('self'/'me') and 'rho', where 'this' contains the left-hand object and 'rho' - right-hand object. For languages like C++,  when you have to use a friendly function to override most operators, you have both lho (for left-hand object) and rho parameters

thorium-zero: Will fix annotations in the next commit

YuriiKindrat (approved):

PR #528, opened 2023-05-20
implemented class Distance to calculate distance

Oknametra:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        if isinstance(other, Distance):
+            distance3 = self.km + other.km
+        else:
+            distance3 = self.km + other
+        return Distance(distance3)
+
+    def __iadd__(self, other: int | float) -> Distance:
```
vasylhnatiuk, line 21:
```suggestion
    def __iadd__(self, other: int | float | Distance) -> Distance:
```

vasylhnatiuk (changes requested):
__lt__ , __gt__ , __eq__ , __le__, __ge__  can also get Distance. add code for this case

app/main.py
Oknametra: thanks, i'll keep that in mind

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
vasylhnatiuk, line 34:
```suggestion
    def __lt__(self, other: int | float | Distance) -> bool:
            if isinstance(other, Distance):
              return self.km < other.km
```

vasylhnatiuk: Add also for all of the bellow functions if they get Distance as other

vasylhnatiuk (changes requested):
you didn't do it `lt , gt , eq , le, ge can also get Distance. add code for this case`

app/main.py
Oknametra: ah, ok, thx

Vasyl-Smutok (approved):
Great 👍

PR #530, opened 2023-05-22
Solution

prykhodkom:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 45: You must implement logic when other is Distance

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
```
EdAlekseiev, line 26: other can be Distance too

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
```
EdAlekseiev, line 29: other can be Distance too

EdAlekseiev (changes requested):

app/main.py
prykhodkom: image

EdAlekseiev: Sorry, It was my bad. Your solution is correct.

EdAlekseiev (approved):
Good job

PR #531, opened 2023-05-24
done

chipapapo:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: object or int or float) -> object:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 21: use else 
```suggestion
     else:
```

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 23:
```suggestion
        return self
```

app/main.py
```diff
@@ -1,3 +1,59 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int or float) -> None:
```
ArsenPidhoretskyi, line 2: `int or float`
Is not correct use or Union[int, float] or int | float
Fix in all places

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object or int or float) -> object:
```
ArsenPidhoretskyi, line 11: It returns Distance, not object

vasylhnatiuk (approved):
Good job

PR #533, opened 2023-05-25
'Solution'

olmatiash:
Hello!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,62 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: ["Distance", float]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: [int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: [int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: ["Distance", float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+
+    def __gt__(self, other: ["Distance", float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
ArsenPidhoretskyi, line 11: other also can be int and float

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 14: elif is redundant just do return

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __iadd__(self, other: ["Distance", float]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: [int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 25: No need this check. Due to type annotations code only expect int or float

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: ["Distance", float]) -> "Distance":
```
ArsenPidhoretskyi, line 17:
```suggestion
    def __iadd__(self, other: Union["Distance", float]) -> "Distance":
```
or
```suggestion
    def __iadd__(self, other: "Distance" | float) -> "Distance":
```
Fix in all places

app/main.py
ArsenPidhoretskyi: `__truediv__` the same

app/main.py
ArsenPidhoretskyi: Remove in all places this check

app/main.py
olmatiash: done

app/main.py
olmatiash: done

app/main.py
```diff
@@ -1,3 +1,55 @@
+from typing import Union, Any
+
```
vasylhnatiuk, line 2:
```suggestion
from future import __annotations__
```

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", float, int]) -> Any:
```
```suggestion
    def __add__(self, other: Union[Distance, float, int]) -> Any:
```

vasylhnatiuk: change here and all of the code

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #534, opened 2023-05-26
Solution

gleblitvinenko:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 14: No need to use elif. Due to type annotations in this step it will be always int or float

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
ArsenPidhoretskyi, line 11: int and float also possible

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
ArsenPidhoretskyi, line 17: int and float also possible

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 20: just else:

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 29: This check is not needed. Due to type annotations it always true
The same with mul

app/main.py
ArsenPidhoretskyi: fix in all places

app/main.py
ArsenPidhoretskyi: fix in all places

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __mul__(self, other: int | float) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
vsmutok, line 35: Redundant `else`

fix here and in other places

vsmutok (changes requested):

vsmutok (approved):
Great 👍

Abnormaltype (approved):

ihor-jpeg (approved):

ihor-jpeg (approved):

PR #535, opened 2023-05-29
'solved'

oleksiikolii:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, new: int | float | Distance) -> Distance:
+        if isinstance(new, (int, float)):
+            return Distance(self.km + new)
+        elif isinstance(new, Distance):
+            return Distance(self.km + new.km)
```
YuriiKindrat, line 18: no need to use isinstance twice here, if there will be an unsupported type, it should raise an error

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __iadd__(self, new: int | float | Distance) -> None:
+        if isinstance(new, (int, float)):
+            self.km += new
+        elif isinstance(new, Distance):
+            self.km += new.km
+        return self
```
YuriiKindrat, line 25: same here

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km < new
+
+    def __gt__(self, new: int) -> bool:
+        return self.km > new
+
+    def __le__(self, new: int) -> bool:
+        return self.km <= new
+
+    def __ge__(self, new: int) -> bool:
+        return self.km >= new
```
YuriiKindrat, line 47: image
What if `new` is of type `Distance` here and in all the following methods?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, new: int | float | Distance) -> Distance:
+        if isinstance(new, (int, float)):
+            return Distance(self.km + new)
+        else:
```
YuriiKindrat, line 17: do u need `else` here?)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __iadd__(self, new: int | float | Distance) -> None:
+        if isinstance(new, (int, float)):
+            self.km += new
+        else:
+            self.km += new.km
+        return self
+
+    def __mul__(self, new: int) -> Distance:
+        res = self.km * new
```
YuriiKindrat, line 28: no need to create variable here)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return Distance(res)
+
+    def __truediv__(self, new: int) -> Distance:
+        return Distance(round(self.km / new, 2))
+
+    def __eq__(self, new: int | Distance) -> bool:
+        if isinstance(new, (int, float)):
+            return self.km == new
+        else:
```
YuriiKindrat, line 37: do u need else here?)

YuriiKindrat: fix this everywhere

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(new, (int, float)):
+            return self.km <= new
+        else:
+            return self.km <= new.km
+
+    def __ge__(self, new: int | Distance) -> bool:
+        if isinstance(new, (int, float)):
+            return self.km >= new
+        else:
+            return self.km >= new.km
```
YuriiKindrat, line 62: You could use previously created magic method here to simplify your code

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self
+
+    def __mul__(self, new: int) -> Distance:
+        return Distance(self.km * new)
+
+    def __truediv__(self, new: int) -> Distance:
+        return Distance(round(self.km / new, 2))
+
+    def __eq__(self, new: int | Distance) -> bool:
+        return self.km == new
```
Oleg-s-s, line 33: What if new is of type Distance here and in all the following methods?
Yurii already commented you about this

Oleg-s-s (changes requested):

app/main.py
oleksiikolii: I can make it more complex by checking `isinstance(new, Distance)` but why?

oleksiikolii: or maybe make another method to convert in one type but still why

YuriiKindrat (approved):

PR #537, opened 2023-05-29
Solution

DmytroTsehelnyk:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 46: image
What if `other` is of type `Distance` here and in all the following methods?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: int | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 56: try to simplify this, using previously created methods
hint: `>= is not <`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __eq__(self, other: int | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | Distance) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __ge__(self, other: int | Distance) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
YuriiKindrat, line 52: u can use backward condition here, to simplify ur code, also no need to call methods like this, u can use operators instead

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job)

PR #538, opened 2023-05-29
Solution

AntonShpakovych:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 32: are u sure about type annotation here?

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
```
YuriiKindrat, line 57: try to simplify this, using previously created methods
hint: `>= is not <`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return not self > other and not self == other
```
Oleg-s-s, line 45: You can simplify this too, using previously created methods

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self > other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self > other or self == other
```
Oleg-s-s, line 51: Same here

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km > other
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other) and not self.__eq__(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
Oleg-s-s, line 51: no need to call methods like this, u can use operators
also try to use backward condition here

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(round(self.km / other, 2))
```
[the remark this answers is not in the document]
AntonShpakovych: fixed

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return not self > other and not self == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self > other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self < other
```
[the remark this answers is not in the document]
AntonShpakovych: Yes, of course, I can create two methods that cover the remaining comparison methods, but the description of this task stated "implement all of this methods: ....methods," and in my opinion, implementing all these methods means writing them from scratch.  But ok I will del

Oleg-s-s (approved):
Good job)

PR #541, opened 2023-05-29
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+
+        if isinstance(other, int) or isinstance(other, float):
+            self.km += other
+            return self
```
YuriiKindrat, line 21: no need to use isinstance twice, u can just use `else` here
also instead of double isinstance u can write like this `isinstance(other, (int, float))`
also u are returning the same thing for 2 times, could u use only one return here?

YuriiKindrat: fix this everywhere

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            is_less_than = self.km < other.km
+        else:
+            is_less_than = self.km < other
+        return is_less_than
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        return not self.__le__(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
Oleg-s-s, line 48: Try to use the same style as in __ge__()

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Good job

PR #542, opened 2023-05-29
Solution

volodymyr-komarnyckyi:


YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            other = other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __ge__(self, other: Distance | int) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
YuriiKindrat, line 50: u could use operant here, no need to call methods like this

YuriiKindrat: also u can simplify ur solution using backward conditions

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job

PR #543, opened 2023-05-29
done

maccarets:


app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        if isinstance(dist, Distance):
+            return Distance(self.km + dist.km)
+        return Distance(self.km + dist)
+
+    def __iadd__(self, dist: Union["Distance", float, int]) -> "Distance":
+        if isinstance(dist, Distance):
+            self.km += dist.km
+            return self
+        self.km += dist
+        return self
```
YuriiKindrat, line 24: it is better to use else instead of 2 returns )

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __eq__(self, dist: Union["Distance", float, int]) -> bool:
+        if isinstance(dist, Distance):
+            return self.km == dist.km
+        return self.km == dist
+
+    def __ge__(self, dist: Union["Distance", float, int]) -> bool:
+        return not self.__lt__(dist)
+
+    def __le__(self, dist: Union["Distance", float, int]) -> bool:
+        return not self.__gt__(dist)
```
YuriiKindrat, line 51: gj, but u can use operants to call this methods, no need to call this methods like this

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #544, opened 2023-05-29
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
```
YuriiKindrat, line 23: u can use else here, no need to use condition here)

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 56: try to simplify this, using previously created methods 
hint: `>= is not <`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not (self > other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not (self < other)
```
YuriiKindrat, line 51: unnecessary brackets here

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job

PR #545, opened 2023-05-29
solution

Tarasidze:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        else:
```

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
YuriiKindrat, line 26: try to use only one return here

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            km = round((self.km / other), 2)
+            return Distance(km)
```
YuriiKindrat, line 34: why do u check for other type in truediv method, but didn't check this in mul method ? 
also, no need to create variable here, u can write this directly to distance class

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
YuriiKindrat, line 40: try to simplify this solution - u can use only one return here) same for other methods

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
YuriiKindrat, line 64: You should use previously created magic method here to simplify your code

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round((self.km / other), 2))
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == other.km if isinstance(other, Distance)\
```
LiudmylaKulinchenko, line 35: Please, don't split using `\` as a line break. Rewrite without ternary operator

LiudmylaKulinchenko (changes requested):

PR #546, opened 2023-05-30
Solution

erikagayan:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError("Unsupported operand type for +")
```
YuriiKindrat, line 17: won't it raise an error by default if i pass an unsuposed type ?
no need to use double check, fix this everywhere

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        else:
+            raise TypeError("Unsupported operand type for *")
+
+    def __truediv__(self, other: int) -> float:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError("Unsupported operand type for /")
```
YuriiKindrat, line 38: do u need else here?)

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        else:
+            raise TypeError("Unsupported operand type for <=")
+
+    def __ge__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        else:
+            raise TypeError("Unsupported operand type for >=")
```
YuriiKindrat, line 78: u can use previously created methods here)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
```
LiudmylaKulinchenko, line 11: You have wrong type annotations, your `other` could have also `Disntance` or `float` type

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 14: You can skip this check and just return

LiudmylaKulinchenko (changes requested):
Mind comment throughout the code

app/main.py
LiudmylaKulinchenko: You can skip `elif` check and make return right away

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance" | float:
```
LiudmylaKulinchenko, line 14: You have already imported annotations from __future__, so use `Distance` without quotes
Change annotations for `other`, it could be not just integer value & function can return just `Distance` instance

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: int) -> int:
+        return self.km <= other
+
+    def __ge__(self, other: int) -> int:
+        return self.__gt__(other) or self.__eq__(other)
```
LiudmylaKulinchenko, line 57: You can call these magic methods using operands

LiudmylaKulinchenko (changes requested):
Mind comments throughout the code

app/main.py
erikagayan: But when I skip elif in __add__, an error appears (AssertionError)

erikagayan: When I run it like this, it gives an error:  
    def __add__(self, other: int | Distance) -> Distance:

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
YuriiKindrat, line 17: why do u need this else here?

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> "Distance":
```
YuriiKindrat, line 20: wrong type annotation, fix this like in add method

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
YuriiKindrat, line 23: no need to use elif here, u can just make else here

app/main.py
YuriiKindrat: fix this everywhere

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: int) -> int:
+        return self.km <= other
+
+    def __ge__(self, other: int) -> int:
+        return self.km >= other
```
YuriiKindrat, line 57: does this methods return int ? 
type annotation for other is wrong
no need to make check for instance twice 
in le/ge methods other could be a Distance instance

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | Distance | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | Distance | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | Distance | float) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 47: what about distance instance ? will this work correctly ?

YuriiKindrat (changes requested):

app/main.py
erikagayan: Yes, all the tests pass

LiudmylaKulinchenko (approved):
Well done 😎

PR #547, opened 2023-05-30
Solution

YevheniiIovchu:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
```
Oleg-s-s, line 29: `other` can not be of type `Distance` in this method so check is unnecessary

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
Oleg-s-s, line 33: `other` can not be of type `Distance` in this method so check is unnecessary

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
Oleg-s-s, line 18: You can skip this check here and directly return from this point
Fix through all code

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return False
```
Oleg-s-s, line 40: Redundant return

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
```
YuriiKindrat, line 22: no need to make this check here, it is better to use else here)

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 55: try to use methods implemented before to simplify ur code
hint: `>= is not <`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km == other
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return not Distance.__eq__(self, other)\
+            and not Distance.__lt__(self, other)
```
YuriiKindrat, line 44: avoid using linebreakers in ur code
also no need to call methods like this, u can use operators for this

YuriiKindrat: fix everywhere

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km < other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return not (self == other) and not (self < other)
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return not (self > other)
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return not (self < other)
```
Oleg-s-s, line 49: unnecessary brackets here

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Good job)

PR #548, opened 2023-05-30
'Solution'

RedCatHot:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+        return self.km == other
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+
+        return self.km <= other
```
LiudmylaKulinchenko, line 57: You should use previously created magic method here to simplify your code (call methods using operands)

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🚀

PR #549, opened 2023-05-30
solution

vlodek69:
I noticed that all comparison methods work just fine like that (see attached file).
What is the best practice?
Untitled


vasylhnatiuk (approved):
Good job, you used good option

PR #550, opened 2023-05-30
Solution

Kirontiko:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 60: gj, but try to simplify this, using previously created methods
hint: `>= is not <`

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 35: are u sure about type annotation here?

YuriiKindrat (changes requested):

app/main.py
Kirontiko: Oh, I understand. I will fix it

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self.__lt__(other)
```
YuriiKindrat, line 56: gj, but u can use operators instead of calling method like this

YuriiKindrat (changes requested):

PR #552, opened 2023-05-30
Python magic methods

volholovko:


YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: [int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [int, float]) -> [int, float]:
```
YukihiroSM, line 12: I meant `Distance` without `"`. 
```suggestion
    def __add__(self, other: Distance) -> Distance:
```

YukihiroSM (changes requested):
Please, take a look into comment. Anything else looks good! Nice job!

app/main.py
volholovko: yes, but when I do without " I get an error:
app\main.py:12: in Distance
E   NameError: name 'Distance' is not defined
=============================================== short test summary info ================================================ 
ERROR tests/test_main.py - NameError: name 'Distance' is not defined

YukihiroSM: Yeah, sorry, my bad. You can change back to `"`. Sorry for that. Change back to what it was before and I will approve it.

YukihiroSM (changes requested):

app/main.py
volholovko: no problem)

YukihiroSM (approved):

PR #553, opened 2023-05-30
implemented py-distance-class

mariiahorbova:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return self.km > other
+        return self.km > other.km
+
+    def __lt__(self, other: Any) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
+
+    def __ge__(self, other: Any) -> bool:
+        if isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 58: You should use previously created magic method here to simplify your code (call methods using operands)

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> Any:
+        if isinstance(other, (int, float)):
+            distance = Distance(self.km + other)
```
LiudmylaKulinchenko, line 16: The additional variable is redundant here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> Any:
```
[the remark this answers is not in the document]
mariiahorbova: done

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, (int, float)):
```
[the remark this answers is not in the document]
mariiahorbova: done

app/main.py
mariiahorbova: done

app/main.py
mariiahorbova: don't quite understand the comment

mariiahorbova: done

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            self.km = self.km + other
+            return self
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km = self.km + other
+            return self
+        else:
+            self.km += other.km
+            return self
```
YuriiKindrat, line 27: u are returning the same thing every time, could u use only one return here?

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __truediv__(self, number: float | int) -> Distance:
+        return Distance(round(self.km / number, 2))
+
+    def __eq__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return self.km == other.km
+
+    def __ne__(self, other: Distance) -> bool:
+        return not self.km.__eq__(other.km)
```
YuriiKindrat, line 41: no need to call method in that way, u can use operator for this `=`

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __gt__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return not self.km.__eq__(other) and not self.km.__lt__(other)
+        return not self.km.__eq__(other.km) and not self.km.__lt__(other.km)
```
YuriiKindrat, line 46: same here

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __ge__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return not self.km.__lt__(other)
+        return not self.km.__lt__(other.km)
+
+    def __le__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km.__eq__(other) or self.km.__lt__(other)
+        return self.km.__eq__(other.km) or self.km.__lt__(other.km)
```
YuriiKindrat, line 61: same

YuriiKindrat: u can implement a le/ge mthods in simpler way
hint: `>= is not <`

YuriiKindrat (changes requested):
also, check ur type annotations for other in eq, gt, lt, etc. methods, is other will always be a Distance instance ?

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
LiudmylaKulinchenko, line 14: `other` could also be `float` instance. Fix annotations, please 🌺

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return not self.km < other
+        return not self.km < other.km
```
YuriiKindrat, line 54: no need to make check for instance here, u are already check this in lt method, that u are using here

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km == other or self.km < other
+        return self.km == other.km or self.km < other.km
```
YuriiKindrat, line 59: same is here, also u can make a backward condition as for ge method

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #554, opened 2023-05-30
Solution

imelnyk007:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+
+        return self.km < other
+
+    def __gt__(self, other: float | Distance) -> bool:
+        return not self.__lt__(other) and not self.__eq__(other)
```
YuriiKindrat, line 41: no need to call methods like this, u can use operants for this

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+        return self.km == other
+
+    def __le__(self, other: float | Distance) -> bool:
+        return self.__eq__(other) or self.__lt__(other)
+
+    def __ge__(self, other: float | Distance) -> bool:
+        return self.__eq__(other) or not self.__lt__(other)
```
YuriiKindrat, line 53: try to simplify this, using previously created methods
hint: `>= is not <`

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):
also all tests should pass, fix ur flake8 errors)

Oleg-s-s (approved):
Good job)

PR #556, opened 2023-05-30
Solution

Gvilion:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if type(other) in (int, float):
```
YuriiKindrat, line 15: try to use isinstance instead) fix this everywhere

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return Distance(km=self.km + other)
+        return Distance(km=self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if type(other) in (int, float):
+            self.km = self.km + other
+            return self
+        self.km = self.km + other.km
+        return self
```
YuriiKindrat, line 24: it is better to use else instead of 2 returns

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if type(other) in (int, float):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if type(other) in (int, float):
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 55: try to use priviously created methods here
hint: `>= is not <`

YuriiKindrat (changes requested):

app/main.py
LiudmylaKulinchenko: Make corrections according to this comment

LiudmylaKulinchenko: Make corrections according to this comment

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other: isinstance | int | float) -> bool:
+        if type(other) in (int, float):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: isinstance | int | float) -> bool:
+        return not self.__gt__(other)
```

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: isinstance | int | float) -> isinstance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+        self.km = self.km + other
+        return self
```
YuriiKindrat, line 24: it is better to use else instead of 2 returns here)

LiudmylaKulinchenko (approved):
Good job!! 🚀

PR #557, opened 2023-05-30
solution

12-deadblue:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
YuriiKindrat, line 14: no need to make this check here

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
YuriiKindrat, line 20: u can use else instead )

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
YuriiKindrat, line 61: no need to make double check in this methods

YuriiKindrat, line 61: try to simplify le/ge methods, using previously created methods
hint: `>= is not <`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __eq__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
```
LiudmylaKulinchenko, line 47: You can skip `else` and just make return

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
LiudmylaKulinchenko, line 11: Are you sure `other` could have only `Distance` type? Fix annotations according to the task description

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __mul__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: "Distance") -> bool:
```
LiudmylaKulinchenko, line 32: You can call these magic methods using operands

LiudmylaKulinchenko (changes requested):
Mind all comments throughout the code

app/main.py
12-deadblue: I don't know how to do it.

YuriiKindrat (approved):

PR #558, opened 2023-05-30
Solution

Po-SvinSkipodla:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+
+        self.km = self.km + other
+        return self
```
YuriiKindrat, line 29: it is better to use else instead of 2 returns here

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+        return self.km > other
+
+    def __ge__(self, other: int) -> bool:
+
+        return self.km >= other
+
+    def __eq__(self, other: int) -> bool:
+
+        return self.km == other
```
YuriiKindrat, line 57: other also could be a Distance type, fix this

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = self.km + other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 37: why do u need blank lines here?

YuriiKindrat (changes requested):
fix this

app/main.py
Po-SvinSkipodla: Why not?

app/main.py
Po-SvinSkipodla: For more readable code, i think

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
LiudmylaKulinchenko, line 15: `other` could also have a `float` type. Fix annotations, please

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km <= other
+
+    def __gt__(self, other: int | float) -> bool:
+        return self.km > other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
```
YuriiKindrat, line 47: u didn't fix that, other also could be an instance of Disctance class in this methods
image

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        else:
+            self.km = self.km + other
+
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 32: this 2 methods should not work with Distance instance

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km <= other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return self.km > other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.km >= other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        return self.km == other
```
YuriiKindrat, line 47: will this work correctly with Distance ?

YuriiKindrat (changes requested):

Oleg-s-s (approved):

PR #559, opened 2023-05-30
My solution

kovaliskoveronika:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if type(other) == int or type(other) == float:
```
YuriiKindrat, line 15: try to use `isinstance` method) fix everywhere

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: u didn't fix that (

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: u can use isintance like that f.e.`isinstance(other, (int, float))`

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __eq__(self, other: Distance | float) -> bool:
+        if isinstance(other, float) or isinstance(other, int):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: Distance | float) -> bool:
+        return not (self > other)
+
+    def __ge__(self, other: Distance | float) -> bool:
+        return not (self < other)
```
Oleg-s-s, line 51: Unnecessary brackets here

Oleg-s-s (changes requested):

PR #561, opened 2023-05-30
Solution

Andriy-Sydorenko:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[Distance, int, float]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
YuriiKindrat, line 18: no need to check for int/float type here, the same is for other methods except mul/thruediv

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: Union[Distance, int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
YuriiKindrat, line 64: try to simplify le/ge methods, using previously created methods
hint: `>= is not <`

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job

PR #562, opened 2023-05-30
"Solution"

HDKESL:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km > self.check_instance_type(other)
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == self.check_instance_type(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= self.check_instance_type(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= self.check_instance_type(other)
```
YuriiKindrat, line 47: try to make solution smaller here)
hint: u can use methods created before `>= is not <`

YuriiKindrat (changes requested):

PR #563, opened 2023-05-31
solution

Kasatello:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            mul_distance = self.km * other
+            return Distance(mul_distance)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            div_distance = round(self.km / other, 2)
```
YuriiKindrat, line 34: no need to create additional var here

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 50: this methods also should work with Distance instance

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
Oleg-s-s, line 57: try to use priviously created methods here
hint: ` >= is not <`

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
Oleg-s-s, line 32: Why you put that check there?)

Oleg-s-s (changes requested):

app/main.py
Kasatello: to ensure that other is an integer or float? i dont need it there?

Oleg-s-s: `other` can not be of type `Distance` in this method and you wrote that in annotation so if it will be wrong type it will automatically call error

Kasatello: got it, ty

PR #567, opened 2023-06-01
Solution

anastmishchuk:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[float, int, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(km=self.km + other)
```
Grayder0152, line 19:
```suggestion
        return Distance(km=self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,57 @@
+from __future__ import annotations
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
```
Grayder0152, line 6: Why do you use `|` here but in 15 line use Union ? choose one thing

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return self
+
+    def __mul__(self, other: Union[float, int]) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: Union[float, int]) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union[float, int, Distance]) -> bool:
+        if self.km < other:
```
vasylhnatiuk, line 34: you must check if other is Distance and add another option like 

            return self.km < other.km

vasylhnatiuk: add here and all of bellow functions

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __mul__(self, other: Union[float, int]) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: Union[float, int]) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union[float, int, Distance]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif self.km < other:
```
vasylhnatiuk, line 36: you can write only this because `self.km < other` returns `True` or `False` like in above code
```suggestion
        return self.km < other
```

vasylhnatiuk: change here and all of the code

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #573, opened 2023-06-06
Implemented Distance-Class task

OleksiiKiva:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return self.km == self.return_number(other)
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self.km <= self.return_number(other)
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.km >= self.return_number(other)
+
+    @staticmethod
+    def return_number(number: int) -> int | float:
```
YukihiroSM, line 50: You have the case, when number might be an instance of Distance.  So, number: int is not correct in this case, because number is Distance. Fix the type hint.

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return Distance(
+            self.km + self.return_number(other)
+        )
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        self.km += self.return_number(other)
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km *= other
```
YukihiroSM, line 25: How do you think is it a good idea to change the input object? `self.km *= other` will change the object itself. 
Also, `other` might be Distance, not only int or float.

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return Distance(self.km)
+
+    def __truediv__(self, other: int | float) -> Distance:
```
YukihiroSM, line 28: Check comment for `__mul__`

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km == self.return_number(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= self.return_number(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= self.return_number(other)
+
+    @staticmethod
+    def return_number(number: Distance | int | float) -> int | float:
```
olena-hul, line 48: No need to use word `return` in function because functions usually return something

Rename it to `get_value` instead

olena-hul (changes requested):

app/main.py
OleksiiKiva: Indeed, for some reason I didn't guess)) Thanks

olena-hul (approved):

PR #574, opened 2023-06-06
Solution

Anastasia-Su:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        )
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance | None:
```
ArsenPidhoretskyi, line 28: mul does not accepts Distance. And it always return distance, because check for isintance is not needed inside
The same with truediv

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __truediv__(self, other: int | float | Distance) -> Distance | None:
+        if isinstance(other, (int, float)):
+            return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        else:
+            return self.km < other.km
```
ArsenPidhoretskyi, line 40:
```suggestion
        return self.km < other.km
```
The same fix in other places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #577, opened 2023-06-10
solution

vkleshko:


app/main.py
```diff
@@ -1,3 +1,81 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
ArsenPidhoretskyi, line 14: other can also be int or float

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            return NotImplemented
```
ArsenPidhoretskyi, line 20: Why you need return Exception Type? In is not raise exception. It just return

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+    def __mul__(self, other: Union[int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        else:
+            return NotImplemented
+
+    def __truediv__(self, other: Union[int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        else:
```
ArsenPidhoretskyi, line 40: else: not needed fix in all places

ArsenPidhoretskyi (changes requested):

PR #587, opened 2023-06-20
Develop

NazarKravchenk0:


Bohdan-Salamakha (changes requested):
1
don't change flake8 settings
2
add typehint for "other"

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
EdAlekseiev, line 11: Other can be int and float too. Fix everywhere
```suggestion
    def __add__(self, other: "Distance" | int | float) -> "Distance":
```

EdAlekseiev (changes requested):

app/main.py
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
vsmutok, line 16: `else` redundant here

vsmutok (changes requested):

PR #588, opened 2023-06-21
Solution

pnakongit:


app/main.py
```diff
@@ -1,3 +1,134 @@
...
+    def __truediv__(self, other: int | float) -> Distance:
+
+        self._distance_types_validator(
+            obj=other,
+            valid_types=(int, float),
+            operator="/"
+        )
+
+        km = self.km / other
+        round_km = int(km) if int(km) == km else float("{:.2f}".format(km))
```
ArsenPidhoretskyi, line 72: round_km = round(km, 2)

ArsenPidhoretskyi (changes requested):

app/main.py
pnakongit: Thx a lot

vsmutok (approved):
Good Job 👍

PR #597, opened 2023-06-27
py-distance-class

StepanShevchuk:


app/main.py
```diff
@@ -1,3 +1,60 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 6: float also possible, fix in all places

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Looks Good 👍

PR #602, opened 2023-06-27
All tests passed, please review my code.

bkirichenko:


app/main.py
```diff
@@ -1,3 +1,61 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 5: float also possible, fix in all places

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return str(f"Distance: {self.km} kilometers.")
+
+    def __repr__(self) -> str:
+        return str(f"Distance(km={self.km})")
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
ArsenPidhoretskyi, line 18: else: is not needed
```suggestion
        return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
ArsenPidhoretskyi, line 37:
```suggestion
        return self.km < other
```
The same with in other boolean methods

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+        else:
+            self.km = self.km + other
+        return self
+
+    def __mul__(self, other: Distance | int) -> Distance:
```
ArsenPidhoretskyi, line 27: `__mul__` does not accepts Distance, only float and int

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
well done

PR #607, opened 2023-07-03
Solution

hrabovii:


app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 5: float also possible

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 25: Move the return after if-else statement. Due to duplicate ocde

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __mul__(self, other: (int, float)) -> Distance:
```
ArsenPidhoretskyi, line 27: Use int | float instead of (int, float)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: (int, float)) -> Distance:
```
ArsenPidhoretskyi, line 29:
```suggestion
    def __truediv__(self, other: int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
ArsenPidhoretskyi, line 24:
```suggestion
        else:
            self.km += other
        return self
```

ArsenPidhoretskyi (approved):
Good

PR #608, opened 2023-07-03
Develop

lubart2517:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = distance
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if type(other) == Distance:
```
Dimosphen1, line 15: Although there are no errors in logic, use `isinstance` over `type` for this case and others, you can read the reasoning for that [here](https://blog.devgenius.io/what-is-better-type-vs-isinstance-3b948d1b81b5) and [here](https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance).

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if type(other) == Distance:
+            self.km = other.km + self.km
```
Dimosphen1, line 21: Use `+=` here as on line 23.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #610, opened 2023-07-06
Solution

EvgenMilokumov:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
vsmutok, line 14: fix everywhere 
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vsmutok (changes requested):

app/main.py
EvgenMilokumov: done

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            Distance(self.km + other.km)
+            if isinstance(other, Distance)
+            else Distance(self.km + other)
+        )
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km += other.km if isinstance(other, Distance) else other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
ArsenPidhoretskyi, line 25: In mul and truediv Distance is imposible
Remove Distance only for this method

ArsenPidhoretskyi (changes requested):

app/main.py
EvgenMilokumov: done

vasylhnatiuk (approved):

PR #611, opened 2023-07-07
solution

vkleshko:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 38: elif is not needed. We check all types due to annotations. Remove this check 
Fix in all places

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float]) -> "Distance":
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 32: This check is not needed

ArsenPidhoretskyi: Check in all places

ArsenPidhoretskyi (changes requested):

PR #620, opened 2023-07-13
Solution

SuliganLP:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return self.km < other
```
vasylhnatiuk, line 36: if `other `is `Distance `instance you must add condition self.km < other.km. add here and in all of bellow functions

vasylhnatiuk (changes requested):

app/main.py
SuliganLP: Done!

Grayder0152 (approved):

PR #636, opened 2023-07-28
Test (py-distance-class)

baranovr:


app/main.py
```diff
@@ -1,3 +1,73 @@
+from typing import Union    # we use "Union" to indicate that the parameter can receive different data types
+
+
```
Polyakiv-Andrey, line 3: You have done a great job implementing the Distance class and its various comparison and arithmetic methods. However, for this specific task, we only need the class definition without the comments. The code you provided is sufficient, and the comments might not be necessary for the current context.

Polyakiv-Andrey (changes requested):

app/main.py
baranovr: It's clear)

PR #640, opened 2023-07-31
py-distance-class

stamaksim:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            return ValueError
```
vsmutok, line 20: remove redundant `else`

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
```suggestion
    def __add__(self, other: Distance | float | int ) -> Distance:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            return None
```
Grayder0152, line 32: why None ?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
Grayder0152, line 18: else raise exception

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return NotImplemented
+        elif isinstance(other, (int, float)):
+            distance = self.km / other
+        return Distance(round(distance, 2))
+
+    def __lt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
```
ArsenPidhoretskyi, line 43: YOu always must return bool
You can raise exception or just do if-else
```
            return self.km < other.km
        else:
            return self.km < other
```


Fix in all places

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            raise ValueError("Unsupported operand type")
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance) -> Distance:
```
ArsenPidhoretskyi, line 29: other can be only int or float
The same with truediv

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise ValueError("Unsupported operand type")
+
+    def __iadd__(self, other: Distance) -> Distance:
```
ArsenPidhoretskyi, line 22: other also can be int/float

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            return NotImplemented
+        elif isinstance(other, (int, float)):
+            distance = self.km / other
+        return Distance(round(distance, 2))
+
+    def __lt__(self, other: Distance) -> bool:
```
ArsenPidhoretskyi, line 39: other can be also int and float

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: @stamaksim not fixed

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, Distance):
+            return NotImplemented
+        else:
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
ArsenPidhoretskyi, line 36: lt also can accept  Distance as other. Fix for all bool methods

ArsenPidhoretskyi: For boolean methods other can be: Distance, int, float

ArsenPidhoretskyi (changes requested):

app/main.py
stamaksim: Thanks

ArsenPidhoretskyi (approved):
Well done!

PR #643, opened 2023-08-03
Implemented solution

RPodolets:


nyanplague (approved):

app/main.py
```diff
@@ -1,3 +1,49 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float | int) -> None:
```
Kroll410, line 5: In most cases it's better to use typing annotations (Union here)

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return "Distance: {} kilometers.".format(self.km)
+
+    def __repr__(self) -> str:
+        return "Distance(km={})".format(self.km)
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        value = other if isinstance(other, (int, float)) else other.km
+        return Distance(self.km + value)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
```
Kroll410, line 18: You can omit duplication here with calling __add__ in __iadd__ :)

Kroll410 (changes requested):

PR #644, opened 2023-08-03
Solution

nyanplague:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
```
Kroll410, line 14: Use type annotations from the `typing` library (`Union` here)

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+                km=self.km + other
+            )
+        else:
+            return Distance(
+                km=self.km + other.km
+            )
+
+    def __iadd__(self, other: Distance | float) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
```
Kroll410, line 26: You can try to omit condition duplicating here, with implementing `__add__` and swapping the parameter in __iadd__, calling the __add__ method, a common practice

Kroll410 (changes requested):

app/main.py
nyanplague: looks good, updated

app/main.py
nyanplague: sure, thank you, changed

Kroll410 (approved):

PR #645, opened 2023-08-03
Done

DimaPr7:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __add__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
```
Grayder0152, line 24:
```suggestion
            self.km += other.km
        else:
            self.km += other
```

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
```
Grayder0152, line 21: add the same for all methods

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 26: Move return after if-else statement. No need to duplicate code.
```suggestion
        else:
            self.km += other
        return self
```

ArsenPidhoretskyi (changes requested):
Small fix required

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Union[int, float, Distance]) -> Distance:
+        return self.km == other
+
+    def __le__(self, other: Union[int, float, Distance]) -> Distance:
+        return self.km <= other
+
+    def __ge__(self, other: Union[int, float, Distance]) -> Distance:
+        return self.km >= other
```
ArsenPidhoretskyi, line 47: They all return bool, not Distance

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #646, opened 2023-08-04
Implemented magic methods as requested

sberdianskyi:


app/main.py
```diff
@@ -1,3 +1,49 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    km: int
+
+    def __init__(self, km: int) -> None:
```

app/main.py
```diff
@@ -1,3 +1,49 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    km: int
```
ArsenPidhoretskyi, line 5: km is instance attribute. Remove it from here

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance) is True:
```
ArsenPidhoretskyi, line 17: is True is redundant
```suggestion
        if isinstance(other, Distance):
```

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
ArsenPidhoretskyi, line 20: else: is not needed
```suggestion
        return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance) -> Distance:
```
ArsenPidhoretskyi, line 16: float also possible fix in all places

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance) is True:
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 28: Duplicate of code. Move return self after if-else statement

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __mul__(self, other: int) -> Distance:
```
ArsenPidhoretskyi, line 30: float also possible

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> bool:
```
ArsenPidhoretskyi, line 36: float and Distance also possible for all boolean methods

ArsenPidhoretskyi (changes requested):

app/main.py
sberdianskyi: Done

app/main.py
sberdianskyi: done

app/main.py
sberdianskyi: Done

app/main.py
sberdianskyi: Done

app/main.py
sberdianskyi: Done

app/main.py
sberdianskyi: Done

app/main.py
sberdianskyi: Done

app/main.py
sberdianskyi: Done

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance) is True:
```
```suggestion
        if isinstance(other, Distance):
```
This is redundant

ArsenPidhoretskyi (approved):
Good job

PR #647, opened 2023-08-05
methods for Distance class implemented, test passed

maxkatkalov:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __add__(self, other: Distance | int) -> Distance | int:
+        if isinstance(other, Distance):
+            return Distance(
+                self.km + other.km
+            )
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance | int:
```
Barsh4ec, line 25: I don't think that __iadd__ and __add__ can return int

Barsh4ec (changes requested):

Barsh4ec (approved):

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(
+                self.km + other.km
+            )
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
```
Kroll410, line 25: You can call __add__ within __iadd__, a common practice

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(
+            round(self.km / other, 2)
+        )
+
+    def __eq__(self, other: Distance | (int, float)) -> bool:
+        if not isinstance(other, Distance):
+            return self.km == other
```
Kroll410, line 42: This will be comparing Distance to any object, not `int` precisely

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if not isinstance(other, Distance):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance) -> bool:
+        if not isinstance(other, Distance):
+            return self.km >= other
+        return self.km >= other.km
+
+    def __ne__(self, other: Distance) -> bool:
```
Kroll410, line 65: != operator return booleans

Kroll410: You cam simplify

Kroll410 (changes requested):

app/main.py
maxkatkalov: okey, done

thx u

app/main.py
```diff
@@ -1,3 +1,65 @@
+from __future__ import annotations
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(
+            self,
+            km: int
```
Y-Havryliv, line 8: it could be float as well

Y-Havryliv: fix everywhere please

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(
+                self.km + other.km
+            )
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        self.km = (self.__add__(other)).km
```
Y-Havryliv, line 27: don't call magic methods directly

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if not isinstance(other, Distance):
+            return self.km < other
+        return self.km < other.km
+
+    def __gt__(self, other: Distance | (int, float)) -> bool:
+        if not isinstance(other, Distance):
+            return self.km > other
+        return self.km > other.km
+
+    def __le__(self, other: Distance | (int, float)) -> bool:
```
Y-Havryliv, line 54: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if not isinstance(other, Distance):
+            return self.km == other
+        return self.km == other.km
+
+    def __lt__(self, other: Distance) -> bool:
+        if not isinstance(other, Distance):
+            return self.km < other
+        return self.km < other.km
+
+    def __gt__(self, other: Distance | (int, float)) -> bool:
```
Y-Havryliv, line 49:
```suggestion
    def __gt__(self, other: Distance | int | float) -> bool:
```
it's not the right annotation. YOu suggesting that it could be either Distance or tuple

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return Distance(
+            round(self.km / other, 2)
+        )
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Distance):
+            return self.km == other
+        return self.km == other.km
+
+    def __lt__(self, other: Distance) -> bool:
```
Y-Havryliv, line 44: it could be int and float as well

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(
+            round(self.km / other, 2)
+        )
+
+    def __eq__(self, other: Any) -> bool:
```
Y-Havryliv, line 39: Any means literally anything, so it's misleading since it could be only DIstance, int or float

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #651, opened 2023-08-07
'Solution'

lgSilay:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
KristianYe, line 19: Would be better without redundant else
```suggestion
        if isinstance(other, Distance):
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```

KristianYe (approved):

app/main.py
lgSilay: Thx for your Comment, I'll correct this asap)

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
```
Kroll410, line 20: You can call __add__ here

Kroll410 (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
Y-Havryliv, line 45: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (approved):

PR #652, opened 2023-08-07
Implemented class Distance and magic methods in it

KristianYe:


innyshka (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
```
Kroll410, line 19: Call __add__ here

Kroll410 (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = self.__add__(other).km
```
Y-Havryliv, line 20: we don't call magic methods directly, right?)

PR #655, opened 2023-08-07
Implemented py-distance-class

katryana:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, Distance]) -> Distance:
```
ORENYT, line 15: Union[int, float, Distance]

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[int, Distance]) -> Distance:
```
ORENYT, line 20: Union[int, float, Distance]

ORENYT (approved):
Clear solution. But you add not only a int or Distance object. In tests there are a float numbers sometimes: 
```
    'kilometers1,kilometers2,result',
    [
        (50, 15, 65),
        (100, 30, 130),
        (300, 100, 400),
        (15.5, 14.5, 30.0),
        (20.6, 40.8, 61.4)
    ]
```

app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Y-Havryliv, line 6: you use `float` here, and `int` in other methods. Fix it please, km could be either `int` or `float`

app/main.py
Y-Havryliv: I agree

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union[int, Distance]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __le__(self, other: Union[int, Distance]) -> bool:
```
Y-Havryliv, line 38: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (approved):

PR #656, opened 2023-08-07
Solution

eduardhabryd:


anna-lybid (approved):

app/main.py
```diff
@@ -1,3 +1,83 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Y-Havryliv, line 5: it could be int as well

app/main.py
```diff
@@ -1,3 +1,83 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
Y-Havryliv, line 14: `elif isinstance(other, (int, float)):`
it could be not only DIstance type

app/main.py
```diff
@@ -1,3 +1,83 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise ValueError("Unsupported operand type for +")
```
Y-Havryliv, line 20: nice

app/main.py
```diff
@@ -1,3 +1,83 @@
...
+    def __iadd__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise ValueError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, factor: int) -> Distance:
```
Y-Havryliv, line 31: here you have `int`, but in truediv you have `float`

app/main.py
```diff
@@ -1,3 +1,83 @@
...
+
+    def __eq__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        else:
+            raise ValueError("Unsupported operand type for ==")
+
+    def __le__(self, other: Distance) -> bool:
```
Y-Havryliv, line 69: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (changes requested):

PR #659, opened 2023-08-08
Solution py-distance-class

anna-lybid:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" or int | float) -> "Distance":
```
denys-source, line 11: it is better to use `Union["Distance", int, float]`

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, int | float):
```
denys-source, line 14: you can keep one `isinstance` call

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance" or int | float) -> "Distance":
```
denys-source, line 17: call `__add__` from this method to remove duplicate code

app/main.py
anna-lybid: As far as I remember, It's ok to use  |  operator from Python 3.11.  Maybe "or" doesn't fit enough. Thanks for your review

anna-lybid: > it is better to use `Union["Distance", int, float]`

As far as I remember, It's ok to use | operator from Python 3.11. Maybe "or" doesn't fit enough. Thanks for your review

denys-source: you are right. it is even recommended in the docs:
> To define a union, use e.g. Union[int, str] or the shorthand int | str. Using that shorthand is recommended.

but using `"Distance" | int | float` will produce an error. to avoid any errors you can wrap this expression in double quotes `"Distance | int | float"`

app/main.py
anna-lybid: > you can keep only one `isinstance` call

Do you mean "else" statement istead of the second isinstance ?

denys-source: Just `return` without second `isinstance`

CaCtUs1k (approved):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        result = self.__add__(other)
```
Y-Havryliv, line 20: you don't add two numbers using `__add__`, right? You use `+` and then __add__ is triggered

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
```
Y-Havryliv, line 23: seems like it's duplicated, you can return it only once

Y-Havryliv (changes requested):

app/main.py
anna-lybid: > seems like it's duplicated, you can return it only once

Thanks

Y-Havryliv (approved):
GJ!

PR #660, opened 2023-08-08
Solution

OlegatorLE:


OleksandrMoroz24 (approved):
good

app/main.py
```diff
@@ -1,3 +1,54 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Y-Havryliv, line 5: it could be `int` as well

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
Y-Havryliv, line 15:
```suggestion
        if isinstance(other, Distance):
```
seems more simple to me) but it's up to you

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        self.km += other.km
```
Y-Havryliv, line 23: imo, using if-else with 1 `return self` would be more readable and simple

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        return self.km >= other.km
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        return self.km < other.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
Y-Havryliv, line 51: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #661, opened 2023-08-08
Solution

CaCtUs1k:


Illia-Kononenko (approved):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> int:
```
Y-Havryliv, line 14: `if isinstance(other, int | float):`
`other` can be not only DIstance. fix everywhere please

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance) -> int:
+        if isinstance(other, int | float):
+            self.km += other
+            return self
```
Y-Havryliv, line 22: you can use `if-else` and return self only once

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, int | float):
+            return self.km < other
+        return self.km < other.km
+
+    def __eq__(self, other: Distance) -> bool:
+        if isinstance(other, int | float):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: Distance) -> bool:
```
Y-Havryliv, line 42: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (changes requested):

app/main.py
CaCtUs1k: I'm sorry, but I don't quite understand what you mean.

Y-Havryliv: e.g. here in `le` we can use previously created magic methods. `return not self > other` if it's not bigger then it's lower or equal, and you don't need to check types, because `gt` will be triggered

PR #663, opened 2023-08-08
Solution

OleksandrMoroz24:


9rosLove (approved):

app/main.py
```diff
@@ -1,3 +1,63 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Y-Havryliv, line 5: `km` could be float as well

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
Y-Havryliv, line 14:
```suggestion
    def __add__(self, other: int | float| Distance) -> Distance:
```
you've checked whether it's a DIstance,but don't have it in annotation

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #665, opened 2023-08-08
'Solution'

grungezz:


Kurta97 (approved):

app/main.py
```diff
@@ -1,3 +1,61 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Y-Havryliv, line 2: it could be `int` as well

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
Y-Havryliv, line 11: `other` can be not only Distance type

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #666, opened 2023-08-08
Created Distance class and its methods

MarianKovalyshyn:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return True if self.km < other.km else False
```
ochernous, line 35: Maybe the ternary operator is redundant?

ochernous (approved):

app/main.py
```diff
@@ -1,3 +1,61 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Y-Havryliv, line 5: `km` could be float as well

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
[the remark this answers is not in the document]
MarianKovalyshyn: Agree. But for me, it is just easier to read that`s why I decided to put 'else'.

Y-Havryliv (approved):
nice

PR #667, opened 2023-08-08
Solution

9rosLove:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
denys-source, line 17: you can remove `else` if there is a `return` statement

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
```
denys-source, line 20: use `__add__` in `__iadd__` method to get rid of duplicate code

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km = self.__add__(other)
```
denys-source, line 20: `self.__add__(other)` returns `Distance` object. you have to assign a value of `km` attribute

PR #669, opened 2023-08-08
'Solution'

ochernous:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
Y-Havryliv, line 17: `else` is not necessary if you have return above

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
```
Y-Havryliv, line 27:
```suggestion
    def __mul__(self, other: int | float) -> Distance:
```
it seems like other can be only int or float based on your code)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
```
Y-Havryliv, line 52: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
+from __future__ import annotations, division
```
Y-Havryliv, line 1: btw, do you use `division`?

app/main.py
ochernous: I agree. Thank you!

app/main.py
ochernous: I agree. Thank you!

app/main.py
ochernous: Good point!

app/main.py
ochernous: I had some problems passing tests with this method. After reading about __truediv__ for some reason I decided I had to use it. But now after reading about __truediv__ again, I see that the division import is used for compatibility with Python 2.

PR #671, opened 2023-08-09
Solution

VyacheslavFomenko:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
```
vasylhnatiuk, line 33: what if other is Distance instance? add if condition for this here and in all of below functions

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
vsmutok, line 35: Remove redundant `else`

vsmutok: fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #672, opened 2023-08-09
py-distance-class

yaroslav-mazurkevych:


app/main.py
```diff
@@ -1,3 +1,67 @@
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
```
Y-Havryliv, line 6: `km` could be `int` as well

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> "Distance":
```
Y-Havryliv, line 15: `Any` means literally any. So can `other` be list or dict? It's misleading a bit

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
Y-Havryliv, line 18: `else` is not necessary if you have return above

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+        return self
+
+    def __mul__(self, other: Any) -> "Distance":
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Any) -> "Distance":
+        if other == 0:
+            raise ZeroDivisionError
```
Y-Havryliv, line 35: what's the point of doing so? Python will raise the same error if other will be 0

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        else:
+            return self.km > other
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: Any) -> bool:
```
Y-Havryliv, line 57: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #673, opened 2023-08-09
Solution

Ihor-MA:


app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+from typing import Callable
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Y-Havryliv, line 6: it can be float as well

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        self.km = km
+
+    def __str__(self) -> None:
```
Y-Havryliv, line 9: does it really return None?

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> None:
```
Y-Havryliv, line 12: same here

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Callable) -> "Distance":
```
Y-Havryliv, line 15:
```suggestion
    def __add__(self, other: int | float | Distance) -> "Distance":
```
imo, since other can be either Distance or number, it's better to use Distance annotation

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(
+                km=self.km + other
+            )
+        else:
```
Y-Havryliv, line 20: `else` is not necessary if you have return above

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            return Distance(
+                km=self.km + other.km
+            )
+
+    def __iadd__(self, other: int | float | Callable) -> "Distance":
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
```
Y-Havryliv, line 28: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        else:
+            self.km += other.km
+            return self
+
+    def __mul__(self, other: int | float | Callable) -> "Distance":
```
Y-Havryliv, line 33: it seems like other can be only int or float based on your code)

Y-Havryliv: same with truediv

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(
+            km=self.km * other
+        )
+
+    def __truediv__(self, other: int | float | Callable) -> "Distance":
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: int | float | Callable) -> bool:
```
Y-Havryliv, line 43: what if `other` would be Distance here?

Y-Havryliv (changes requested):

PR #674, opened 2023-08-09
Solution

ZaichenkoViktoriia:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
Y-Havryliv, line 47: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
Y-Havryliv, line 26: if other would be Distance here, your code will break, same in truediv

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #676, opened 2023-08-10
'Solution'

WonnaKoo:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
```
Kroll410, line 21: use __add__ here instead of duplicating code

Kroll410 (changes requested):
All methods without else block will return None, shouldn't you handle that?

Kroll410 (approved):

PR #681, opened 2023-08-12
py-distance-class

Vladyslav-Pliashechnyk:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
```
Kroll410, line 20: You can call `__add__` here

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(km=round(self.km / other, 2))
```
Kroll410, line 30: what if we divide on zero?

Kroll410 (changes requested):

PR #685, opened 2023-08-17
task completed

DaniilHinzburh:
task completed

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,61 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> object:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: object) -> object:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, factor: float) -> object:
+        if isinstance(factor, (int, float)):
+            return Distance(self.km * factor)
+
+    def __truediv__(self, divisor: float) -> object:
+        if isinstance(divisor, (int, float)):
+            result = self.km / divisor
+            return Distance(round(result, 2))
+
+    def __lt__(self, other: object) -> object:
+        if isinstance(other, Distance):
+            return self.km < other.km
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> object:
```
ArsenPidhoretskyi, line 11: other is not object. It can be int, float, Distance
And return it Distance

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
ArsenPidhoretskyi, line 15: elif is redundant just do like this:
```suggestion
        return Distance(self.km + other)
```

app/main.py
ArsenPidhoretskyi: Fix annotations in all places

app/main.py
ArsenPidhoretskyi: Fix in all places

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __mul__(self, factor: float) -> object:
+        if isinstance(factor, (int, float)):
+            return Distance(self.km * factor)
+
+    def __truediv__(self, divisor: float) -> object:
+        if isinstance(divisor, (int, float)):
+            result = self.km / divisor
+            return Distance(round(result, 2))
+
+    def __lt__(self, other: object) -> object:
```
ArsenPidhoretskyi, line 33: Boolean methods returns bool

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, factor: float) -> Distance:
```
vasylhnatiuk, line 26:
```suggestion
    def __mul__(self, factor: int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(self.km * factor)
+
+    def __truediv__(self, divisor: float) -> Distance:
```
vasylhnatiuk, line 29:
```suggestion
    def __truediv__(self, divisor: int | float) -> Distance:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #686, opened 2023-08-17
Solution

VasylTurok:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+
+        self.km += other.km
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
pavlejviki, line 28: __mul__ should expect  other only as int | float

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
```
pavlejviki, line 32: truediv should expect other only as int | float

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, (int, float)):
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
pavlejviki, line 52: You can use previously created magic method here(and in __ge__ as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+
+        self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
pavlejviki, line 29: now this check is redundant

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
pavlejviki, line 33: now this check is redundant

app/main.py
pavlejviki: is not fixed yet.

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #687, opened 2023-08-18
Solution

prochigor:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
LiudmylaKulinchenko, line 26: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            return self.km > other
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
```
LiudmylaKulinchenko, line 52: You should use previously created magic method here to simplify your code (call methods using operands)

LiudmylaKulinchenko (changes requested):

app/main.py
prochigor: I use it in __ge__, __le__ and __gt__ methods

app/main.py
prochigor: Done

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int) -> bool:
```
fsocie7y, line 33: again about else

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
+
+    def __gt__(self, other: Distance | int) -> bool:
+        return not self < other and not self == other
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
```
fsocie7y, line 43: can u rewrite this one too?)

fsocie7y (changes requested):

app/main.py
prochigor: I rewrite this with others magic methods)

fsocie7y (approved):
GJ 👍

PR #689, opened 2023-08-20
Solution

obohatov:


app/main.py
```diff
@@ -1,3 +1,65 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self,
+                 km: int) -> None:
```
Y-Havryliv, line 6: it could be float as well, fix everywhere please

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+                other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self,
+                 other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
Y-Havryliv, line 25: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self,
+               other: int | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self,
```
Y-Havryliv, line 55: You can use previously created magic method here(and in `ge` as well) to simplify your code. e.g. `return not self > other`

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

PR #691, opened 2023-08-21
Solutions

tkachuk2291:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
```
```suggestion
    def __add__(self, other: int) -> Distance:
```

vasylhnatiuk: from __future__ import annotations use

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> object:
```
vasylhnatiuk, line 19:
```suggestion
    def __iadd__(self, other: int) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> bool:
```
vasylhnatiuk, line 32:
```suggestion
    def __lt__(self, other: int | float | Distance) -> bool:
```

vasylhnatiuk: change here and in all of below code

vasylhnatiuk (changes requested):

tkachuk2291:
fixed

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance:
```
ArsenPidhoretskyi, line 14: other can be float and Distance

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> object:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
ArsenPidhoretskyi, line 26: other can be float also

app/main.py
ArsenPidhoretskyi: Other can be float and Distance also

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #694, opened 2023-08-22
Solution

Daniil-Pankieiev:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @classmethod
+    def checkvalue(cls, other: "Distance") -> float:
```
LiudmylaKulinchenko, line 12: You can just use type annotations and don't check for other types. It is up to developer call methods with the correct data types if there are annotations
So you can remove this method

LiudmylaKulinchenko (changes requested):
Fix type annotations in all methods according to my comment

Familenko (approved):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 51: cool, but one more point
You can call these magic methods using operands
ex: `<` instead `self.__lt__(other)`

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #695, opened 2023-08-23
Implemented Distance class

AndriiHomeniuk1:


app/main.py
```diff
@@ -1,3 +1,82 @@
...
+    def __iadd__(self, other: Union["Distance", int, float]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, other: Union["Distance", int, float]) -> "Distance":
```
ArsenPidhoretskyi, line 32: mul does not accept Distance
The same with truediv

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiHomeniuk1: Fixed annotation to mul and truediv methods

app/main.py
```diff
@@ -1,3 +1,82 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

vasylhnatiuk: use `annotations from __futures__`

vasylhnatiuk (changes requested):

app/main.py
AndriiHomeniuk1: Oh, cool idea, because I didn't understand how to fix the annotation of the output, python indicated that there was an error
Now I understand
thank you

app/main.py
```diff
@@ -1,3 +1,83 @@
+from __future__ import annotations
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int | float) -> None:
```
vasylhnatiuk, line 7:
```suggestion
    def __init__(self, km: Union[int | float]) -> None:
```

vasylhnatiuk (changes requested):

app/main.py
AndriiHomeniuk1: Yes) fixed it

PR #702, opened 2023-08-28
Solution

Depaton:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __add__(self, other: float) -> float:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
pavlejviki, line 9: else is redundant

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __iadd__(self, other: float) -> float:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
pavlejviki, line 24: remove else or one return

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __mul__(self, other: float) -> float:
```
pavlejviki, line 26: other can be integer as well and in some of the methods an instance of Distance class, please correct the type annotation in all your methods

and the return type annotation is wrong as well

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> float:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
pavlejviki, line 35: else is redundant

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return self.km < other
+
+    def __gt__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
```
pavlejviki, line 41: else is redundant

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return self.km > other
+
+    def __eq__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
```
pavlejviki, line 47: else is redundant

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return self.km == other
+
+    def __le__(self, other: float) -> bool:
```
pavlejviki, line 50: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #703, opened 2023-08-29
created Distance class with magic methods

kartoshka2309:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
pavlejviki, line 14: other can be a float as well, please correct type annotation in all your methods

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
pavlejviki, line 17: else is redundant, as you use return. please remove it in all other methods in similar cases

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return self.km > other
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
```
pavlejviki, line 51: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #707, opened 2023-08-30
solution class Distance

lLiftoffl:


app/main.py
```diff
@@ -1,3 +1,68 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
YukihiroSM, line 5: km can be int as well

YukihiroSM (changes requested):
The code is good, but fix one tiny comment, please.

app/main.py
```diff
@@ -1,3 +1,68 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> int:
```
olena-hul, line 5: not fixed

olena-hul (changes requested):
Good job!

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 17: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 23: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: float | int) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 38: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return self.km < other
+        return False
+
+    def __gt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 45: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        return False
+
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 66: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        return False
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 59: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+        return False
+
+    def __eq__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
```
Bohdan-Salamakha, line 52: u don't need the second check because if other is not Distance - then it's number for sure

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            return self.km <= other
+        return False
+
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        return False
```
Bohdan-Salamakha, line 68: u don't need to return False here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        return False
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        return False
```
Bohdan-Salamakha, line 61: u don't need to return False here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km > other
+        return False
+
+    def __eq__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        return False
```
Bohdan-Salamakha, line 54: u don't need to return False here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        return False
+
+    def __gt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+        return False
```
Bohdan-Salamakha, line 47: u don't need to return False here

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __truediv__(self, other: float | int) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        return False
```
Bohdan-Salamakha, line 40: u don't need to return False here

Bohdan-Salamakha (changes requested):

PR #710, opened 2023-08-31
Solution

Artemh59:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __le__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
```
Dimosphen1, line 62: Remove the `else` statement here and other methods where applicable, the code will not execute after the `return` in the `if` block.

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: Union[float, int]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
Dimosphen1, line 15: You could use `|` syntaxis for type annotation instead of `Union` and indicate the return type as `Distance` without `""` here and in other methods. Replace type annotation to utilize the latest language features.

Dimosphen1 (changes requested):

app/main.py
Artemh59: NameError: name 'Distance' is not defined 
I get this error when use Distance without ""

Artemh59: and this 
TypeError: unsupported operand type(s) for |: 'str' and 'type'
where I use "|" with " 'Distance' "

ArsenPidhoretskyi (approved):
Well done!

PR #711, opened 2023-09-01
solution: distance class

sacrosanctry:


app/main.py
```diff
@@ -1,3 +1,55 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
pavlejviki, line 5: km can be integer as well

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
```
pavlejviki, line 14: other can be integer as well, please fix the type annotation in all your methods

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float) -> bool:
```
pavlejviki, line 47: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #712, opened 2023-09-01
Solution

loplicat:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        other_dist = Distance(other)
```
pavlejviki, line 18: you don't need to create an additional instance of the Distance class, you can simply add int or float to self.km

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return Distance(self.km + other_dist.km)
```
pavlejviki, line 19:
```suggestion
        return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        other_dist = Distance(other)
+        self.km += other_dist.km
```
pavlejviki, line 26: see comment above

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        other_dist = Distance(other)
```
pavlejviki, line 38: no need to create a new instance

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return self.km < other_dist.km
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        other_dist = Distance(other)
```
pavlejviki, line 44: no need to create a new instance

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return self.km > other_dist.km
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        other_dist = Distance(other)
+        return self.km == other_dist.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
pavlejviki, line 53: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

loplicat:
Thank you for your comments, I've already fixed the code. Could you please check it once more?

Dimosphen1 (approved):
LGTM, well done!

PR #719, opened 2023-09-09
'Solution'

nsdNite:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __truediv__(self, other: int) -> Distance:
+        div_result = round((self.km / other), 2)
+        return Distance(km=div_result)
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance) and self.km < other.km:
+            return True
+        if self.km < other:
+            return True
+        return False
```
pavlejviki, line 39: to simplify the code, you can directly return the result of the comparison:

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
```
pavlejviki, line 40:
```suggestion
return self.km < other
```

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __mul__(self, other: int) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        div_result = round((self.km / other), 2)
+        return Distance(km=div_result)
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance) and self.km < other.km:
+            return True
```
pavlejviki, line 36:
```suggestion
return self.km < other.km
```

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        if self.km < other:
+            return True
+        return False
+
+    def __gt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance) and self.km > other.km:
+            return True
+        if self.km > other:
+            return True
+        return False
```
pavlejviki, line 46: see comment above

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance) and self.km == other.km:
+            return True
+        if self.km == other:
+            return True
+        return False
```
pavlejviki, line 53: see comment above

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __le__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance) and self.km <= other.km:
```
pavlejviki, line 56: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

app/main.py
nsdNite: wow, very interesting solution, many thanks

app/main.py
nsdNite: done

app/main.py
nsdNite: done

app/main.py
nsdNite: done. thanks for the comment

app/main.py
nsdNite: done

app/main.py
nsdNite: done

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
Polyakiv-Andrey, line 23: use else here and just one return

Polyakiv-Andrey (changes requested):

app/main.py
nsdNite: hey, thanks for comment
you mean i should place retuen at previous "if" indent?

smth like this?

def __iadd__(self, other: Distance | int) -> Distance:
            self.km += other.km
    return self

nsdNite: done, thanks for comment

viktoria-rybenchuk:
GJ!

viktoria-rybenchuk (approved):

PR #722, opened 2023-09-11
Solution

valikdeb:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ["Distance", int, float]) -> object:
```
pavlejviki, line 11: wrong type annotation. You can use  Union[ ] or `|` symbol to indicate  that a type could be one of multiple options

and the return type annotation should be more specific

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        else:
```
pavlejviki, line 14: else is redundant

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return Distance(km=self.km + other)
+
+    def __iadd__(self, other: ["Distance", int, float]) -> object:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+        else:
+            self.km = self.km + other
+            return self
```
pavlejviki, line 23: you can simplify and use just one return

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __truediv__(self, other: [int, float]) -> object:
+        if other != 0:
+            res_km = round((self.km / other), 2)
+            return Distance(km=res_km)
+
+    def __lt__(self, other: ["Distance", float, int]) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+        else:
```
pavlejviki, line 36: else is redundant, please review all your methods

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            return self.km > other.km
+
+    def __eq__(self, other: ["Distance", int, float]) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km == other
+        else:
+            return self.km == other.km
+
+    def __le__(self, other: ["Distance", int, float]) -> bool:
```
pavlejviki, line 51: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #723, opened 2023-09-12
Solution distance class

Shapoval1992:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        km = other if isinstance(other, (int, float)) else other.km
+        return Distance(km=self.km + km)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+        else:
+            self.km = self.km + other
+            return self
```
pavlejviki, line 26: you can use just one return here

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        km = other if isinstance(other, (int, float, Distance)) else other.km
```
pavlejviki, line 29: if its an instance of Distance class km should be other.km

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(km=self.km * km)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        km = other if isinstance(other, (int, float, Distance)) else other.km
```
pavlejviki, line 33: if its an instance of Distance class km should be other.km

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __gt__(self, other: int | float | Distance) -> bool:
+        km = other if isinstance(other, (int, float)) else other.km
+        return self.km > km
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        km = other if isinstance(other, (int, float)) else other.km
+        return self.km == km
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        km = other if isinstance(other, (int, float)) else other.km
```
pavlejviki, line 50: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return self.km == km
+
+    # def __le__(self, other: int | float | Distance) -> bool:
+    #     km = other if isinstance(other, (int, float)) else other.km
+    #     return self.km <= km
+    #
+    # def __ge__(self, other: int | float | Distance) -> bool:
+    #     km = other if isinstance(other, (int, float)) else other.km
+    #     return self.km >= km
+
```
viktoria-rybenchuk, line 56: remove comments

viktoria-rybenchuk (changes requested):
you have failed tests. Please fix it

Polyakiv-Andrey (approved):
Good job!

PR #724, opened 2023-09-12
Solution

MatsalakO:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __truediv__(self, divider: int) -> None:
+        return Distance(round(self.km / divider, 2))
+
+    def __lt__(self, kilometers: Distance | int) -> bool:
+        return self.km < kilometers
+
+    def __gt__(self, kilometers: Distance | int) -> bool:
+        return self.km > kilometers
+
+    def __eq__(self, kilometers: Distance | int) -> bool:
```
pavlejviki, line 38: you need to handle the case when kilometers is an instance of Distance class. Review the rest of the methods

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(kilometers, Distance):
+            return self.km > kilometers.km
+        return self.km > kilometers
+
+    def __eq__(self, kilometers: Distance | int) -> bool:
+        if isinstance(kilometers, Distance):
+            return self.km == kilometers.km
+        return self.km == kilometers
+
+    def __le__(self, kilometers: Distance | int) -> bool:
```
pavlejviki, line 47: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __mul__(self, multiplier: int) -> None:
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, divider: int) -> None:
```
pavlejviki, line 29: wrong return type annotation

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(round(self.km / divider, 2))
+
+    def __lt__(self, kilometers: Distance | int) -> bool:
```
pavlejviki, line 32: kilometers: can be float as well

app/main.py
MatsalakO: > You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

i don't understand what i need to do in this case

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, kilometers: Distance | int) -> bool:
+        if isinstance(kilometers, Distance):
+            return self.km <= kilometers.km
+        return self.km <= kilometers
+
+    def __ge__(self, kilometers: Distance | int) -> bool:
+        if isinstance(kilometers, Distance):
+            return self.km >= kilometers.km
+        return self.km >= kilometers
```
Dimosphen1, line 55: You could implement these methods by utilizing the previously written methods, e.g. for `__ge__` you could rely on your implementation of the `__lt__`, thus writing that the object is greater or equal when it is not less than the compared object. In the code, it would be just one line `return not self < kilometers` for this case. This requirement is not mandatory, but you could experiment for yourself with this implementation.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(kilometers, Distance):
+            return self.km == kilometers.km
+        return self.km == kilometers
+
+    def __le__(self, kilometers: Distance | int) -> bool:
+        if isinstance(kilometers, Distance):
+            return self.km <= kilometers.km
+        return self.km <= kilometers
+
+    def __ge__(self, kilometers: Distance | int) -> bool:
```
Dimosphen1, line 52: `kilometers` is not the best argument name for methods, it would be better to use `other` to be more general with arguments that are passed here.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __mul__(self, multiplier: int) -> Distance:
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, divider: int) -> Distance:
+        return Distance(round(self.km / divider, 2))
+
+    def __lt__(self, kilometers: Distance | (int, float)) -> bool:
```
Dimosphen1, line 32: You could write this type annotation as the following without the use of `()` here and in other methods as well:
```suggestion
    def __lt__(self, kilometers: Distance | int | float) -> bool:
```

Dimosphen1 (approved):
LGTM, several comments are left for consideration.

app/main.py
MatsalakO: > You could implement these methods by utilizing the previously written methods, e.g. for `__ge__` you could rely on your implementation of the `__lt__`, thus writing that the object is greater or equal when it is not less than the compared object. In the code, it would be just one line `return not self < kilometers` for this case. This requirement is not mandatory, but you could experiment for yourself with this implementation.

thanks for the answer

PR #726, opened 2023-09-12
implemented py-distance-class task

HlovaOleh:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
```
olena-hul, line 14: other can be int too

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(other + self.km)
+        return Distance(km=self.km + other.km)
+
+    def __iadd__(self, other: Distance) -> Distance:
```
olena-hul, line 19: other can be int and float too

olena-hul (changes requested):

PR #727, opened 2023-09-12
Solution_py_distance_class

VladKrit-commits:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
```
Dimosphen1, line 57: You could rely on the implementation of other methods such as `__gt__` here and in other methods as well.

Dimosphen1 (changes requested):

VladKrit-commits:
I made it(i hope so:)) in the "...(changed)" commit

viktoria-rybenchuk (commented):
Good job!

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, int | float):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, int | float):
+            return Distance(round(self.km / other, 2))
+        elif isinstance:
```
pavlejviki, line 33:
```suggestion
        elif isinstance(other, Distance):
```

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            raise TypeError(
+                "'__truediv__' method shouldn't accept Distance instance"
+            )
+        return Distance(self.km / other.km)
```
pavlejviki, line 37: other can't be Distance instance so this return is redundant

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, int | float):
+            return self.km > other
+        return self.km > other.km
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, int | float):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: int | float | Distance) -> bool:
```
pavlejviki, line 54: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

pavlejviki (changes requested):

VladKrit-commits:
Done in solution(changed_2)

PR #729, opened 2023-09-14
'Solution'

marinaua13:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, "Distance"]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[int, Distance]) -> Distance:
```

vasylhnatiuk: Change everywhere, add `from __future__ import annotations ` in the file beginning

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, Distance]) -> Distance:
```
```suggestion
    def __add__(self, other: Union[int, float,Distance]) -> Distance:
```

vasylhnatiuk: And other can be float as well

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #730, opened 2023-09-15
Develop

oleksashcherbakov:
pleasr check

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -75,8 +75,3 @@ class Distance:
         if self.km >= other.km:
             return True
         return False
-
-var_1 = Distance(10.1)
-var_2 = Distance(10.1)
-
-print(var_1 == 10.2)
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,82 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float):
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> object:
+        if not isinstance(other, Distance):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int | float) -> object:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> object:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> object:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | object) -> bool:
+        if isinstance(other, int) or isinstance(other, float):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,77 @@
...
+            if self.km < other:
+                return True
+            return False
+
+        if self.km < other.km:
+            return True
+        return False
+
+    def __gt__(self, other: int | float | object) -> bool:
+        if isinstance(other, int) or isinstance(other, float):
```
pavlejviki, line 40: You can use isinstance with a tuple of types if you want to check whether a variable is an instance of one of several types.

pavlejviki, line 40:
```suggestion
        if isinstance(other, (int, float)):
```

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> object:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> object:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | object) -> bool:
```
pavlejviki, line 29: the type annotation should be more specific, replace object with Distance in all your methods

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+        if isinstance(other, int) or isinstance(other, float):
+            if self.km < other:
```
pavlejviki, line 31: you can directly return the comparison result.( it will be True/False)

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+                return True
+            return False
+
```
pavlejviki, line 34:
```suggestion
    return self.km < other
```

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+        if isinstance(other, int) or isinstance(other, float):
+            if self.km == other:
+                return True
+            return False
+
+        if self.km == other.km:
+            return True
+        return False
+
+    def __le__(self, other: int | float | object) -> bool:
```
pavlejviki, line 59: You can use previously created magic method here(and in ge as well) to simplify your code. e.g. return not self > other

app/main.py
```diff
@@ -1,3 +1,77 @@
...
+    def __init__(self, km: int | float):
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> object:
```
pavlejviki, line 11: other can be an instance of Distance class as well

pavlejviki (changes requested):

pavlejviki (changes requested):
comments haven't been fixed

pavlejviki (changes requested):
Please fix all previous comments

oleksashcherbakov:
эта ветка уже удалена.
исправленное решение висит на ветке develop_2
ссылка тут - https://github.com/mate-academy/py-distance-class/pull/732/files
у меня к нему доступ только такой

PR #731, opened 2023-09-15
Fixed the branch from master to develop

raychw:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: int | float) -> None:
+        self. km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
Dimosphen1, line 15: You should indicate here that the `other` could be a `Distance` class instance as well, your code encounters that.

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float) -> Distance:
```
Dimosphen1, line 21: You should indicate here that the `other` could be a `Distance` class instance as well, your code encounters that.

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
```
Dimosphen1, line 29: You should process the `Distance` class instance as well; add it to the code and remove the raise of the `TypeError`.

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            raise TypeError("Unsupported operand type for\
+             /: Distance and non-numeric")
+        else:
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
```
Dimosphen1, line 36: You should process the `Distance` class instance as well; add it to the code and remove the raise of the `TypeError`.

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
Dimosphen1, line 56: You should process the `Distance` class instance as well for these methods as mentioned in the task description.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
```
pavlejviki, line 29: wrong type annotation for other. in __mul__ and __truediv__ methods it can be only int or float

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
```
pavlejviki, line 37: you need to handle the case when other is an instance of Distance class. in this and following methods

pavlejviki (changes requested):

app/main.py
Bohdan-Salamakha: implement Distance case instead of 'pass'

app/main.py
Bohdan-Salamakha: implement Distance case instead of 'pass'

app/main.py
Bohdan-Salamakha: not fixed

Bohdan-Salamakha (changes requested):
fix all comments

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
```
Dimosphen1, line 32: You should remove the `Distance` class instance from the type annotation here and in the `__mul__`.

Dimosphen1 (approved):
PR is approved, take into consideration the comment that is left regarding type annotation.

PR #735, opened 2023-09-20
Solution

sixsixsixsamurai:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: "Distance" | int) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: "Distance" | int | float) -> "Distance":
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: int) -> bool:
```
pavlejviki, line 32: handle cases when other is an instance of Distance class. in the following methods as well

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance" | int) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: "Distance" | int) -> "Distance":
```
pavlejviki, line 26: in mul and truediv methods other is not supposed to be a Distance instance

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
pavlejviki, line 14:
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

pavlejviki: fix in all methods

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
```
pavlejviki, line 52: Use previously created magic methods
<= is the same as not >

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
```
pavlejviki, line 57: Use previously created magic methods

pavlejviki (changes requested):

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __gt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
```
pavlopro, line 46: Please change le and ge method to explicit comparison rather than relying on the opposite comparison's result

```
        return self.km <= other.km
    return self.km <= other

        return self.km >= other.km
    return self.km >= other
```

pavlopro (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Dimosphen1, line 5: Do not forget to indicate all acceptable types in the type annotation in the future
```suggestion
    def __init__(self, km: int | float) -> None:
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #739, opened 2023-09-23
Solution

kravetb:


app/main.py
```diff
@@ -1,3 +1,42 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int | float) -> None:
```
olena-hul, line 3: The implementation is not correct and tests should be failing locally. Make sure that flake8 and pytest are passing locally and send the task again

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __truediv__(self, num: int) -> "Distance":
+        length = round(self.km / num, 2)
+        return Distance(length)
+
+    def __lt__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
olena-hul, line 41:
```suggestion
        return self.km < other
```

Fix in other places too

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #743, opened 2023-09-26
'Solution'

Illya-Maznitskiy:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> "Distance":
```
Dimosphen1, line 11: `other` argument could be `int` or `Distance` class instance as well. Besides that, write return type without the use of `""`. Change your type annotation here and in other methods as well to encounter that. In case you will face annotation errors, place `from __future__ import annotations` at the beginning of the file.

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: float) -> bool:
```
Dimosphen1, line 57: You could implement some of the methods relying on the logic of others, e.g. `other` would be greater or equal to `self` in case `self` would be not less than `other`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> Distance:
```
Dimosphen1, line 14: `other `argument could be `int` or `Distance` class instance as well, improve your type annotation to encounter that where applicable.
```suggestion
    def __add__(self, other: float | int | Distance) -> Distance:
```

Dimosphen1 (changes requested):
Several changes in type annotation were requested.

Polyakiv-Andrey (approved):
Good job!

PR #745, opened 2023-09-27
feat: solution

sashokpilipok7:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if (isinstance(other, Distance)):
+            return Distance(self.km + other.km)
+        else:
```
vasylhnatiuk, line 17: else is redundant here

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if (isinstance(other, Distance)):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
vasylhnatiuk, line 26: use return only one time

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
vasylhnatiuk, line 14: other can be float also

vasylhnatiuk (changes requested):

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: OtherType) -> Distance:
+        value = other.km if isinstance(other, Distance) else other
+        return Distance(self.km + value)
+
+    def __iadd__(self, other: OtherType) -> Distance:
+        value = other.km if isinstance(other, Distance) else other
+        self.km += value
```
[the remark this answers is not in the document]
sashokpilipok7: > what happened if `other` type str?

regarding our criterias, it cannot be string, as i understand

app/main.py
```diff
@@ -1,3 +1,54 @@
+from __future__ import annotations
+
+OtherType = object | int | float
```
ArsenPidhoretskyi, line 3:
```suggestion
OtherType = Distance | int | float
```

ArsenPidhoretskyi (changes requested):
Small fix required

PR #747, opened 2023-09-30
Solution 30.09

Viva1-ua:


app/main.py
```diff
@@ -1,3 +1,58 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 5: It can accept float

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, int | float):
+            return Distance(self.km * other)
```
ArsenPidhoretskyi, line 28: Do the same as truediv

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Yurnerroo, line 24: make only one `return self`

Yurnerroo (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
Yurnerroo, line 25: you can return `self` only once at the end of the function

Yurnerroo (changes requested):

app/main.py
Viva1-ua: I tried it but some tests don't seem to work with one return. I wonder if there is any other mistake in my code that affects the current one but I am not sure

vsmutok (approved):
Looks Good

PR #748, opened 2023-09-30
Solution

Naz-iv:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        return Distance(
+            self.km + other.km
+            if isinstance(other, Distance)
+            else self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
```
fsocie7y, line 21: Please, don't split using \ as a line break. 
same in all similar cases

fsocie7y (changes requested):

YuliaHladyshkevych (approved):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __gt__(self, other: Distance | int) -> bool:
+        return (self.km > other.km
+                if isinstance(other, Distance)
+                else self.km > other)
+
+    def __eq__(self, other: Distance | int) -> bool:
+        return not self.__gt__(other) and not self.__lt__(other)
+
+    def __le__(self, other: Distance | int) -> bool:
+        return not self.__gt__(other)
```
LiudmylaKulinchenko, line 49: You can call these magic methods using operands

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        return Distance(
+            self.km + other.km
+            if isinstance(other, Distance)
+            else self.km + other)
```
LiudmylaKulinchenko, line 19: Fix formatting: brackets must be located in one of two options: open and start with the text or have a line break between the text

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
LiudmylaKulinchenko, line 15: What if `other` will be float instance?

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        self.km = self.km + other.km \
+            if isinstance(other, Distance) \
+            else self.km + other
+        return self
+
+    def __mul__(self, multiplier: int) -> Distance:
+        self.km *= multiplier
+        return self
```
LiudmylaKulinchenko, line 29: Create new Distance object here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __add__(self, other: Distance | int) -> Distance:
+        return Distance(
+            self.km + other.km
+            if isinstance(other, Distance)
+            else self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        self.km = self.km + other.km \
+            if isinstance(other, Distance) \
+            else self.km + other
```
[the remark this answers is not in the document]
Naz-iv: changed everywhere apart from line 16, 17 since I didn't want to create temporary variables.

app/main.py
Naz-iv: thanks for hint, updated in code

app/main.py
Naz-iv: fixed

app/main.py
Naz-iv: since type of km parameter of distance class  was not specified I used int since all examples had integers and tests worked fine with it. Since arguments annotations don't effect program execution if other is float program will still work same way it did with int. But Distance instance  will get parameter km of type float instead of int.

app/main.py
Naz-iv: done

oleksandr-pupchenko (approved):

app/main.py
LiudmylaKulinchenko: Yes! But you should specify in function that it can work also with floats. This is how mathematical operations should work, without additional warnings that function gets only integer values

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: gotcha, updated type annotations to avoid having warning if float used instead of int

Katherine-Greg (approved):
Well done! 👍

LiudmylaKulinchenko (approved):
Well done! 🍬

PR #749, opened 2023-10-01
Solution

vshvanska:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
fsocie7y, line 56: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __eq__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float) -> bool:
+        return not Distance.__gt__(self, other)
+
+    def __ge__(self, other: Distance | float) -> bool:
+        return not Distance.__lt__(self, other)
```
fsocie7y, line 52: u are close)
You can call these magic methods using operands `< or > ` 
instead of `Distance.__lt__(self, other) or Distance.__gt__(self, other)`

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return Distance(km=round(self.km / number, 2))
+
+    def __lt__(self, other: Distance | float | int) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: Distance | float) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: Distance | float) -> bool:
+        return self.km == other
```
fsocie7y, line 40: add check on Distance object in this methods
use isinstance() 
like here 
image

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return not self.km > other.km
+        return not self.km > other
+
+    def __ge__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return not self.km < other.km
+        return not self.km < other
```
LiudmylaKulinchenko, line 56: You don't need a check if you anyway do the same action

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job!! 🎈🌼😊

PR #751, opened 2023-10-02
Solution_py-distance-class

VladKrit-commits:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, number: int) -> "Distance":
```
viktoria-rybenchuk, line 27:
```suggestion
    def __mul__(self, number: int) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(km=self.km * number)
+
+    def __truediv__(self, number: int) -> "Distance":
```
viktoria-rybenchuk, line 30: number could be float and Distance as well. Fix in all place

viktoria-rybenchuk (changes requested):

app/main.py
VladKrit-commits: Done it in "...class2"

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
viktoria-rybenchuk, line 25: you can make it with one return

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):

PR #752, opened 2023-10-02
Develop

D1st3f:
Added solution!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,82 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError("Unsupported operand type for +")
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, factor: float) -> Distance:
+        if isinstance(factor, (int, float)):
+            return Distance(self.km * factor)
+        else:
+            raise TypeError("Unsupported operand type for *")
+
+    def __truediv__(self, divisor: float) -> Distance:
+        if isinstance(divisor, (int, float)):
+            result = self.km / divisor
+            return Distance(round(result, 2))
+        else:
+            raise TypeError("Unsupported operand type for /")
+
+    def __lt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        else:
+            raise TypeError("Unsupported operand type for <")
+
+    def __gt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+        else:
+            raise TypeError("Unsupported operand type for >")
+
+    def __eq__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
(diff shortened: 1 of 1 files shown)


DenPrislipskyi (approved):

app/main.py
```diff
@@ -1,3 +1,82 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
LiudmylaKulinchenko, line 14: Fix annotations. You work not only with Distance object here

app/main.py
```diff
@@ -1,3 +1,82 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError("Unsupported operand type for +")
+
```
LiudmylaKulinchenko, line 21: You don't need it, you can just specify types in the function annotations. It will be up to a developer to use function properly

app/main.py
```diff
@@ -1,3 +1,82 @@
...
+            raise TypeError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, factor: float) -> Distance:
+        if isinstance(factor, (int, float)):
+            return Distance(self.km * factor)
+        else:
+            raise TypeError("Unsupported operand type for *")
+
+    def __truediv__(self, divisor: float) -> Distance:
```
LiudmylaKulinchenko, line 37: Don't change parameters names here. Let it be `other` everywhere to have the same style throughout the code

LiudmylaKulinchenko (changes requested):
Mind comments everywhere throughout the code

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
```
Oxbay, line 38: Hello! Very interesting solution, I'm impressed! However, I believe the 'elif' block here and below can be removed; in my opinion, it's unnecessary.

Oxbay (changes requested):
"Hello! Very interesting solution, I'm impressed! However, I believe the 'elif' block here and below can be removed; in my opinion, it's unnecessary."

app/main.py
D1st3f: elif is needed because when you use Distance you call the .km attribute, and when you get a numeric value in int you cannot call .km

Oxbay: u can use like simple return after, so if it`s not a "Distance" - return the other thing.

Oxbay (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 28: Why do you need this method to accept Distance object for `other`?
You can just remove Distance from annotation and then this check will be unneeded

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round((self.km / other), 2))
```
LiudmylaKulinchenko, line 33: Same here. No need to add annotation of types you don't want to work with in function or method

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
LiudmylaKulinchenko, line 56: Almost there 🎈
try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #753, opened 2023-10-02
Solution distance class

DenPrislipskyi:


app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int) -> bool:
+        return self.km >= other
```
mykolamateichuk, line 51: `other` can also be a `Distance` instance.

mykolamateichuk (changes requested):
Good job!

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
```
fsocie7y, line 17: You can skip this check here and directly return from this point

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            result_distance = self.km * other
+            return Distance(result_distance)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
```
fsocie7y, line 32: `other` can not be of type `Distance` in this method

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
```
fsocie7y, line 27: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+            return Distance(result_km)
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == other
```
fsocie7y, line 44: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | Distance) -> bool:
+        return self.km >= other
```
fsocie7y, line 50: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: int | Distance) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 50: u are close)
but it can be written simpler
use operand `<` instead of `self.__lt__(other)` etc.

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+            return Distance(result_km)
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        return self.km == other
```
fsocie7y, line 43: add check on Distance object in this methods
use isinstance()

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 27: You don't need this check here. According to your type annotations, `other` can have just int or float type
It is up to developer to use your function properly. You don't need to handle this

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self < other or self == other
```
LiudmylaKulinchenko, line 52: <= is the same as not >

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self < other or self.__eq__(other)
```
LiudmylaKulinchenko, line 51:
```suggestion
        return not self > other
```
try that

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __ge__(self, other: int | Distance) -> bool:
+        return self > other or self == other
```
LiudmylaKulinchenko, line 54:
```suggestion
        return not self < other
```

LiudmylaKulinchenko (approved):
Good job 🦋😸

PR #754, opened 2023-10-02
Solution

mykolamateichuk:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object | int) -> "Distance":
```
kostomeister, line 11: Need to change type hints, it's better to use Distance instead of object, because you can only get either Distance instance or int

kostomeister (changes requested):
Need to change type hints

kostomeister (approved):
Well done!

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __le__(self, other: object | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: object | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
fsocie7y, line 52: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y, line 52: fix type annotation, use `Distance` instead the `object`

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, number: int) -> "Distance":
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: int) -> "Distance":
+        return Distance(round(self.km / number, 2))
```
fsocie7y, line 27: u can use `Distance` without quotes for type annotations

fsocie7y (changes requested):

app/main.py
mykolamateichuk: It won't let me do this. I get an error saying: "Unresolved reference Distance", and it suggests creating class Distance in main.py.

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
+        return not self.__gt__(other)
```

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
LiudmylaKulinchenko, line 14: What if `other` will be float?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🐣🎈

PR #755, opened 2023-10-02
Solution

kostomeister:


Wonsky1 (approved):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+
+        return self.km >= other
```
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+        return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 56: You are close, but it can be wroten sipmlier)
You can call these magic methods using operands

fsocie7y (changes requested):

fsocie7y (approved):
well done 🚀

PR #756, opened 2023-10-03
Solution

DenVia:


RVChornyy (approved):
Seems to be good!

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, Distance)) -> Distance:
```
LiudmylaKulinchenko, line 15: Use `|`
`int | float | Distance`

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+    def __le__(self, other: (int, float, Distance)) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: (int, float, Distance)) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+
+        return self.km >= other.km
```
LiudmylaKulinchenko, line 58: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as `not >`

LiudmylaKulinchenko (changes requested):

app/main.py
DenVia: Should I use "|" every time I have multiple arguments or when there are more than 2?
Should I use it every time instead of a coma?
I hope I don't need to redo this for all my homeworks)
Should I use it in all parts of the methods/functions?
image

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __eq__(self, other: (int | float | Distance)) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return self.km == other.km
+
+    def __le__(self, other: (int | float | Distance)) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: (int | float | Distance)) -> bool:
+        return not self.__lt__(other)
```
LiudmylaKulinchenko, line 53: Almost there) U can use operands to call magic methods instead of calling them directly

LiudmylaKulinchenko (changes requested):
When you annotate a variable with (int, float), it indicates that the variable should be a tuple where the first element is an int and the second element is a float

LiudmylaKulinchenko (approved):
Well done! 🔥

PR #759, opened 2023-10-03
Solution

petrykivd:


RVChornyy (commented):
Probably you might reduce quantity of redundant "if" and "elif" in your class methods.
Mentor would ask you to do so for sure.

YuliaHladyshkevych (changes requested):
Almost all methods (except __str__ and __repr__)can take not only int.

katyakoziar (approved):

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+            if self.km > other:
+                return True
+        return False
+
+    def __eq__(self, other : Any) -> bool:
+        if isinstance(other, Distance) and self.km == other:
+            return True
+        if isinstance(other, (int, float)) and self.km == other:
+            return True
+        return False
```
fsocie7y, line 59: can we make solution a little bit simplier?

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+    def __ge__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            if self.km >= other.km:
+                return True
+            return False
+        elif isinstance(other, (int, float)):
+            if self.km >= other:
+                return True
+            return False
```
fsocie7y, line 79: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
petrykivd: @fsocie7y wow, very good advice! Thank you👍

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> Distance:
```
fsocie7y, line 16: You can skip this check here and directly return from this point
in other similar cases too

fsocie7y, line 16: and please fix type annotations for func arguments everywhere) 
`Any` its not the best here
try smtn like this `Distance | int | float` etc.

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return self.km == other
+        if isinstance(other, (int, float)):
+            return self.km == other
+        return False
+
+    def __le__(self, other: Any) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Any) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 62: and here u can just use operands
`<` instead of `self.__lt__(other)` etc.

fsocie7y (changes requested):
u are close)
lets go fix it 🚀

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
LiudmylaKulinchenko, line 25: try to rewrite without having return duplicated

LiudmylaKulinchenko (changes requested):

app/main.py
petrykivd: Thanks, fixed😄

LiudmylaKulinchenko (approved):
Good job! 🌼🎈😸

PR #761, opened 2023-10-03
Solution

katyakoziar:


sakhaline (approved):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: "Distance") -> bool:
+        return self.km < other
+
+    def __gt__(self, other: "Distance") -> bool:
+        return self.km > other
+
+    def __eq__(self, other: "Distance") -> bool:
+        return self.km == other
```
fsocie7y, line 41: fix type annotations and use `isinstance`

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            self.km += other
+        elif isinstance(other, Distance):
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> "Distance":
+        return Distance(round(self.km / other, 2))
```
fsocie7y, line 32: fix return type annotations

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km > other
+
+    def __eq__(self, other: "Distance") -> bool:
+        return self.km == other
+
+    def __le__(self, other: "Distance") -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: "Distance") -> bool:
+        return self.km >= other
```
fsocie7y, line 47: Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
```
fsocie7y, line 50: u can do it without else)

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | int) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 56: u are close)
but it can be written simpler

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        elif isinstance(other, Distance):
+            result = self.km + other.km
+        return Distance(result)
```
LiudmylaKulinchenko, line 19: No need to use elif here, your function gets only Distance | int | float types

LiudmylaKulinchenko: fix everywhere

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km += (isinstance(other, Distance) and other.km) or other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
LiudmylaKulinchenko, line 23: What if other is float? You should fix annotations everywhere

LiudmylaKulinchenko (approved):
Good job, but pay attention to annotations!!!

PR #762, opened 2023-10-03
Implement pi-distance-class

SevKrok:


RabenkoYevhenii (approved):
But everything else is perfect ;)

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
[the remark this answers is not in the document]
SevKrok: I'm playing it safe)))

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
fsocie7y, line 63: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
```
fsocie7y, line 51: You can skip second check here and directly return from this point
in one case above too

app/main.py
fsocie7y, line 17: here, as Yevhenii gives advice

fsocie7y, line 17: here, as Yevhenii gives advice

fsocie7y, line 17: here, as Yevhenii gives advice

fsocie7y, line 17: here, as Yevhenii gives advice

fsocie7y (changes requested):

fsocie7y (approved):
GJ 🥇

PR #763, opened 2023-10-03
Solution

sakhaline:


MiskoRuslan (approved):
I don't see any errors, great code

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+
+        return self.km >= other
```
Hint: <= is the same as not >

fsocie7y (changes requested):

SelFisher1488 (approved):
great solution and fix, nice!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
LiudmylaKulinchenko, line 28: What if other is float? fix everywhere

LiudmylaKulinchenko (changes requested):

PR #766, opened 2023-10-03
solution

SelFisher1488:


roffi37 (approved):
s kaifom 👍

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: Union[int, "Distance"]) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Union[int, "Distance"]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
Hint: <= is the same as not >

fsocie7y (changes requested):
u can use Distance in type annotations without quotes)

LiudmylaKulinchenko (approved):
Well done! 😸

PR #768, opened 2023-10-03
Solution

maxvell4:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> isinstance:
+        if isinstance(other, Distance):
+            res = other.km + self.km
```
viktoria-rybenchuk, line 13: use descriptive variable name instead of `res`

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            res = other + self.km
+        return Distance(res)
+
+    def __iadd__(self, other: int) -> isinstance:
```
viktoria-rybenchuk, line 18: extend type annotations for other, it could be float  and `Distance` as well

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
```
viktoria-rybenchuk, line 22: else is redundant,  fix in all places

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            self.km += other
+            return self
+
+    def __mul__(self, other: int) -> isinstance:
```
viktoria-rybenchuk, line 26: The return type annotation should be specified, fix in all places

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (changes requested):
fix comments above

viktoria-rybenchuk (changes requested):
several changes required

app/main.py
maxvell4: Don`t understand how to work without Else

viktoria-rybenchuk: When one condition is true, we return a value and do not proceed to the other block of code. You can just remove else or make this code with one  return

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+
+        if isinstance(other, Distance):
+            return Distance(other.km + self.km)
+        return Distance(other + self.km)
```
Dimosphen1, line 18: It is a better practise to place an instance attribute in the first place and the upcoming argument in the second position during different operations.

Dimosphen1 (approved):
LGTM, well one! One comment is left for consideration.

app/main.py
maxvell4: If you swap places, the code doesn’t work, I don’t understand why?

maxvell4:
Corrected

PR #769, opened 2023-10-03
Solution

MiskoRuslan:
Code can have problem in __add__    and    __iadd_ magic methods.
becouse i use Python 3.6 for solving this task and can`t import __future__ and use "ClassName" for future anotation, but i hope i update my python version today

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,52 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other):
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: float) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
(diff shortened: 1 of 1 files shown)


MiskoRuslan:
I know that i can use "ClassName" or import annotations from "__future__", but my version of Python (3.6) dos`t allow me to do this :( I`m sorry, my bad, i fix it soon, when I can update the python version in my IDE

Shatrovskyi (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __eq__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: float) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 49: u are close)
but it can be written simpler

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __truediv__(self, other: float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float) -> bool:
+        return self.__lt__(other.km) if isinstance(other, Distance) \
+            else self.km < other
+
+    def __gt__(self, other: float) -> bool:
+        return self.__gt__(other.km) if isinstance(other, Distance) \
+            else self.km > other
```
fsocie7y, line 38: Try to avoid using linebreakers `\` in your code

fsocie7y (changes requested):

app/main.py
MiskoRuslan: kk  😶‍🌫️

LiudmylaKulinchenko (approved):
Well done! 😸

PR #770, opened 2023-10-03
'Solution'

fedorov-s-od:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" | int | float) -> Optional["Distance"]:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
Andrii-Silenko, line 20:
```suggestion
        if isinstance(other, Distance):
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```
Maybe here and in next situations we can write code withoute second if, because we already have anotations of parameters

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        if isinstance(other, (int, float)):
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int) -> Optional["Distance"]:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Optional["Distance"]:
+        return Distance(round(self.km / other, 2))
```
Andrii-Silenko, line 38: Maybe here we can make annotation of function as -> Distance, because I don't see the way how we can get None return.

Andrii-Silenko (approved):

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" | int | float) -> Optional["Distance"]:
```
fsocie7y, line 15: dont duplicate conditions with `isinstance` everywhere it can be possible

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+
+        if isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: "Distance" | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+
+        if isinstance(other, (int, float)):
+            return self.km >= other
```
fsocie7y, line 73: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
fedorov-s-od: Good advice, thanks )

app/main.py
fedorov-s-od: Noted! )

app/main.py
fedorov-s-od: Nice idea! Thank you for the good advice )

fsocie7y (approved):
well done 🚀

PR #772, opened 2023-10-03
Solution

skoval7:


fedorov-s-od (approved):
Excellent job! Clean and clear code )

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float | int) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
fsocie7y, line 14: You can skip this check here and directly return from this point
in other similar cases too

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: float | int) -> "Distance":
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
fsocie7y, line 60: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __iadd__(self, other: float | int) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: float | int) -> "Distance":
+        if isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 24: You don't need this check. Your function gets only float and integers, according to the annotations. So this is not necessary
Fix everywhere

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float | int) -> "Distance":
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other: float | int) -> "Distance":
+        if isinstance(other, Distance):
```
LiudmylaKulinchenko, line 37: In your annotations you have only float and integer. If you want to work with Distance object here, add it to your annotatoins

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: float | int) -> "Distance":
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: float | int) -> "Distance":
+        return not self.__gt__(other)
```
LiudmylaKulinchenko, line 47: Use operands instead

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __eq__(self, other: Union["Distance", float, int]) -> "Distance":
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Union["Distance", float, int]) -> "Distance":
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __ge__(self, other: Union["Distance", float, int]) -> "Distance":
+        return self.__gt__(other) or self.__eq__(other)
```
fsocie7y, line 53: u are close)
but it can be written simpler

fsocie7y (changes requested):

app/main.py
LiudmylaKulinchenko: You can skip using quotes for annotations here
From python 3.10 you can also skip using Union, `|` instead

app/main.py
LiudmylaKulinchenko: You should remove this check here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __eq__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float | int) -> Distance:
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float | int) -> Distance:
+        return self.km >= other
```
fsocie7y, line 52: use self not > other for __le__ and self not < other for __ge__

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
fsocie7y, line 56: please fix that according to my advice

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float | int) -> Distance:
+        return not (self > other)
```
LiudmylaKulinchenko, line 49: You can skip using parentheses here

LiudmylaKulinchenko (approved):
Good job! 🍬😸

PR #773, opened 2023-10-03
Solution)))

Ded31:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: Distance | float) -> Distance | None:
```
ArsenPidhoretskyi, line 27: other can not be Distance

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if not isinstance(other, Distance):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float) -> Distance:
```
ArsenPidhoretskyi, line 31: other can not be Distance

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float) -> Distance:
```
YukihiroSM, line 20: other might be int as well

YukihiroSM: Fix it anywhere it's needed

YukihiroSM (changes requested):

YukihiroSM (approved):
Looks good!

PR #774, opened 2023-10-04
Solution

DanSheremeta:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: float | int) -> Distance:
+        return Distance(km=round(self.km / other, 2))
+
+    def __eq__(self, other: float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, (int, float)):
+            return self.km == other
```
ArsenPidhoretskyi, line 37:
```suggestion
        return self.km == other
```
This check is redundant fix in all boolean methods

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self.km < other
+
+    def __le__(self, other: float | int) -> bool:
+        return self.km <= other
+
+    def __gt__(self, other: float | int) -> bool:
+        return self.km > other
+
+    def __ge__(self, other: float | int) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 46: `other` can be Distance. You must keep in mind this case

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(km=self.km + other)
```
EdAlekseiev, line 18: This if-statement is redundant

EdAlekseiev, line 18:
```suggestion
        return Distance(km=self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: float | int) -> Distance:
+        return Distance(km=round(self.km / other, 2))
+
+    def __eq__(self, other: float | int | Distance) -> bool:
+        return self.km == other.km if isinstance(other, Distance) \
```
EdAlekseiev, line 34: Use `()` instead of `\` for separating line

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km <= other.km if isinstance(other, Distance) \
+            else self.km <= other
+
+    def __gt__(self, other: float | int | Distance) -> bool:
+        return self.km > other.km if isinstance(other, Distance) \
+            else self.km > other
+
+    def __ge__(self, other: float | int | Distance) -> bool:
+        return self.km >= other.km if isinstance(other, Distance) \
+            else self.km >= other
```
EdAlekseiev, line 51: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #775, opened 2023-10-04
Solution

Shatrovskyi:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
```
mykolamateichuk, line 25: On line 21 you check if other is instance of Distance class, but you don't type hint it in function declaration.

mykolamateichuk (changes requested):

app/main.py
Shatrovskyi: Got it!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
fsocie7y, line 56: u are close)
try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return (self.__lt__(other.km) if isinstance(other, Distance)
```
LiudmylaKulinchenko, line 34: Avoid using ternary operator to increase readability

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+                else self.km < other)
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return (self.__gt__(other.km) if isinstance(other, Distance)
```
LiudmylaKulinchenko, line 38: use mathematical operands instead of self.__gt__
Fix everywhere

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎈🌼

PR #776, opened 2023-10-04
Solution

Andrii-Silenko:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
LiudmylaKulinchenko, line 55: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
```

LiudmylaKulinchenko (changes requested):

app/main.py
Andrii-Silenko: ok, will try to rewrite

LiudmylaKulinchenko (approved):
Great! ✨🌺

PR #777, opened 2023-10-04
Solution

Lyutillis:


OlehShumov (approved):

app/main.py
```diff
@@ -1,3 +1,62 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: Union[int, float]) -> None:
```
fsocie7y, line 6: u can use | instead of Union

fsocie7y: in all cases)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
fsocie7y, line 15: dont use quotes in type annotations
in each other too

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
fsocie7y, line 18: u can do case like this without else

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Union[int, float, "Distance"]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
```
fsocie7y, line 49: here u also can avoid else statement

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return not self > other
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
```
fsocie7y, line 56: rewrite this in the same way as __ge__ method

fsocie7y: after return u can remove redundant statements)

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #778, opened 2023-10-06
'Solution'

Tituha:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self == other
+
+    @param_check
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self <= other
+
+    @param_check
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self >= other
```
fsocie7y, line 61: try to rewrite these two methods. Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @param_check
+    def __add__(self, other: Distance | int | float) -> Distance:
+        return Distance(self + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km = (self + other).km
```
fsocie7y, line 34: i think u need some check here)

fsocie7y (changes requested):

PR #780, opened 2023-10-06
Implemented py-distance-class

Aeros72:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
```
pavlejviki, line 48: Use previously created magic methods
<= is the same as not >

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
```
pavlejviki, line 53: Use previously created magic methods

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #783, opened 2023-10-07
Solution Distance class

Oxbay:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, *args) -> bool:
+        if isinstance(*args, Distance):
+            return self.km <= args[0].km
+        return self.km <= args[0]
+
+    def __ge__(self, *args) -> bool:
+        if isinstance(*args, Distance):
+            return self.km >= args[0].km
+        return self.km >= args[0]
```
fsocie7y, line 57: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
Oxbay: Done

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, *args) -> Distance:
```
LiudmylaKulinchenko, line 14: Use `other` parameter instead of *args. You should use *args if you don't know how many non-keyword arguments will be passed in a function
In this case it makes your code much less readable

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+            return self.km > args[0].km
+        return self.km > args[0]
+
+    def __eq__(self, *args) -> bool:
+        if isinstance(*args, Distance):
+            return self.km == args[0].km
+        return self.km == args[0]
+
+    def __le__(self, *args) -> bool:
+        return args[0] >= self.km
```
LiudmylaKulinchenko, line 50: Use **previously created** magic methods
Hint: <= is the same as not >

LiudmylaKulinchenko (changes requested):

app/main.py
Oxbay: fixed. Now it`s ok?

app/main.py
Oxbay: fix

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __eq__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance) -> bool:
+        return other > self if other > self else other == self
+
+    def __ge__(self, other: Distance) -> bool:
+        return other < self if other < self else other == self
```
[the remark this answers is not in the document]
Oxbay: i don`t understand...

LiudmylaKulinchenko (approved):
Keep it up! 🔥

PR #784, opened 2023-10-09
Solution

Katherine-Greg:


Lyutillis (approved):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __add__(self, other: Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 23: You forgot to add int to annotations. Fix everywhere

LiudmylaKulinchenko, line 23: You can use just else here

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        if isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        if isinstance(other, (int, float)):
+            return self.km >= other
```
LiudmylaKulinchenko, line 63: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as `not >`

app/main.py
LiudmylaKulinchenko: Fix everywhere

LiudmylaKulinchenko (changes requested):

PR #786, opened 2023-10-09
Solution

OleksandrMV:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
+        return Distance(self.km + other.km)\
+            if isinstance(other, Distance)\
+            else Distance(self.km + other)
```
EdAlekseiev, line 17: Use `()` instead of `\` for separating line

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __mul__(self, other: Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, number: int) -> Distance:
+        return Distance(round(self.km / number, 2))
+
+    def __lt__(self, other: Distance) -> bool:
+        if not isinstance(other, Distance):
+            return True if self.km < other else False
+        return True if self.km < other.km else False
```
EdAlekseiev, line 35:
```suggestion
            return self.km < other
        return self.km < other.km
```

EdAlekseiev: Improve code in other methods too

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #787, opened 2023-10-10
Solution

paashkovaaa:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ("Distance", int, float)) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
YukihiroSM, line 15: remove else. You can just return. Fix everywhere.

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: ("Distance", int, float)) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: ("Distance", int, float)) -> "Distance":
+        return Distance(round((self.km / other), 2))
```
YukihiroSM, line 29: Other can't be Distance in this case.

YukihiroSM (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #790, opened 2023-10-10
olution

MykytaPD09:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
viktoria-rybenchuk, line 17: `other` can be of type int or float as well. fix anything that needs correction

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
```
viktoria-rybenchuk, line 35: You can simplify the code by removing the `elif` statement

viktoria-rybenchuk (changes requested):

app/main.py
MykytaPD09: якщо я видаляю та додаю анотації int та float в мене не проходят тести

pavlejviki: simply remove the whole elif statement. if other is not an instanсe of the Distance class you can straight return self.km < other

pavlejviki, line 35:
```suggestion

```

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return self.km < other
```
pavlejviki, line 36:
```suggestion
        return self.km < other
```

pavlejviki (changes requested):

app/main.py
pavlejviki: not fixed

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: None) -> "Distance":
```
Dimosphen1, line 11: Indicate the correct type annotation for `other`, it could be either `int`, `float`, or an instance of the `Distance` class. Besides that, do not place `Distance` inside of `""`, and omit using quotes in the type annotation. In case you face errors, you could add` from __future__ import annotations` to your file. Revise the type annotation throughout all your code to be representative enough.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #792, opened 2023-10-11
'Solution'

yul1valko:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, int | float):
```
Dimosphen1, line 17: No need to use `elif`, it would be the only option in case other would not be an instance of the `Distance` class. Delete it here and in other methods where applicable.

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
Dimosphen1, line 14: Use `|` instead of `Union` and do not use quotes for return type to indicate an instance of what class you are returning, write class without quotes. In case you will face errors, use `from __future__ import annotations` import.

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, int | float):
+            return self.km <= other
+
+    def __ge__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, int | float):
+            return self.km >= other
```
Dimosphen1, line 64: You could rely on the implementation of other methods in some places, e.g. `self` would be greater or equal to `other` when `self` is not less than `other`.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #798, opened 2023-10-14
Develop

CrazyHilly:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
+        if type(other) == Distance:
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
YukihiroSM, line 18:
```suggestion
        if type(other) == Distance:
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```

YukihiroSM: Fix everywhere.

app/main.py
```diff
@@ -1,3 +1,62 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```
YukihiroSM, line 6: km can be int as well

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
```
YukihiroSM, line 14: other can be int as well. Fix everywhere.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if type(other) == Distance:
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if type(other) == Distance:
```
ArsenPidhoretskyi, line 20: Use isinstance instead of type ==

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 25: move return after if-else statement

app/main.py
ArsenPidhoretskyi: fix in all places

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
ArsenPidhoretskyi, line 22: This line is redundant

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #799, opened 2023-10-15
Solution

comnotaDidi:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __str__(self) -> None:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> None:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> str:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other(int, float)):
```
Dimosphen1, line 18: Here is a typo:
```suggestion
        elif isinstance(other, (int, float)):
```

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return Distance(self.km + other.km)
+        else:
+            raise TypeError("Unsupported operand type for +")
+
+    def __iadd__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
```
Dimosphen1, line 24: You should return `self` here.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> None:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> str:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other(int, float)):
+            return Distance(self.km + other.km)
```
Dimosphen1, line 19: In case `other` is `int` or `float`, it does not have the attribute `km`. Add the value directly.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +")
+
+    def __mul__(self, other: int | float) -> None:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> None:
+        return Distance(self.km / other)
```
Dimosphen1, line 35: You should round the value to two symbols in this method.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __lt__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
Dimosphen1, line 40: There is no need for `else`, code will not execute after return. Make changes where applicable.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return self.km == other
+
+    def __le__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: int | float) -> None:
```
Dimosphen1, line 61: You could implement some methods relying on the implementation of others, e.g. `self` is greater or equal to `other` when the `self` is not less than `other`.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #801, opened 2023-10-16
Solution

Artem-user-19:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return Distance(self.km * scalar)
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __truediv__(self, divisor: int | float) -> int | float | object:
+        if isinstance(divisor, (int, float)):
+            result = self.km / divisor
+            return Distance(round(result, 2))
+        else:
+            raise TypeError("Unsupported operand type")
```
viktoria-rybenchuk, line 39: why do you need `else`?

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __mul__(self, scalar: int | float) -> int | float | object:
+        if isinstance(scalar, (int, float)):
+            return Distance(self.km * scalar)
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __truediv__(self, divisor: int | float) -> int | float | object:
+        if isinstance(divisor, (int, float)):
+            result = self.km / divisor
+            return Distance(round(result, 2))
```
viktoria-rybenchuk, line 37: don't create variable

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __lt__(self, other: int | float) -> int | float | object | bool:
+        return self.km < other.km if isinstance(other, Distance)\
+            else self.km < other
```
viktoria-rybenchuk, line 43: You can improve the readability of the code by using parentheses to break the line instead of a backslash. Fix in all places

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> int | float | object:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
viktoria-rybenchuk, line 16: do you need `else` here?

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            raise TypeError("Unsupported operand type")
+
+    def __iadd__(self, other: int | float) -> int | float | object:
```
viktoria-rybenchuk, line 19: fix type annotation. `Other`  can be `Distance` and return `Distance` as well

viktoria-rybenchuk (changes requested):

pavlejviki (changes requested):
Please make changes according to previous comments

app/main.py
Artem-user-19: Can you give a part of code how to fix it, because I tried to change type annotation and writes about mistake

app/main.py
viktoria-rybenchuk: not fixe. Fix in all places

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: float | int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> int | float | object:
```
viktoria-rybenchuk, line 11: other can be `Distance` as well, fix in all places

viktoria-rybenchuk: Are you sure about return annotations?

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __truediv__(self, divisor: int | float) -> Distance:
+        if isinstance(divisor, (int, float)):
+            return Distance(round(self.km / divisor, 2))
+
+    def __lt__(self, other: int | float) -> bool:
+        return self.km < other.km if isinstance(other, Distance)\
+            else self.km < other
```
viktoria-rybenchuk, line 41:
```suggestion
return (
            self.km < other.km if isinstance(other, Distance)
            else self.km < other
        )
```

viktoria-rybenchuk (changes requested):

PR #805, opened 2023-10-20
'Solution'

Hermesvlad:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
```
ArsenPidhoretskyi, line 29: other can not be Distance. Remove it from type annotations
Also in this case no need to add isinstance, it is redundant because due to type annotations it is always true
```suggestion
    def __mul__(self, other: int | float) -> Distance:
        return Distance(self.km * other)
```

ArsenPidhoretskyi: The same with __truediv__

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
```
ArsenPidhoretskyi, line 39: This check is redundant, because it can be Distance, int, float we already check for Distance and that means after this check it is int or float
```suggestion
        return self.km == other
```
Fix in all places

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
ArsenPidhoretskyi, line 18:
```suggestion
        return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 5:
```suggestion
    def __init__(self, km: int | float) -> None:
```
Float also possible

ArsenPidhoretskyi (changes requested):
Small fix required

ArsenPidhoretskyi (approved):
Well done!

PR #806, opened 2023-10-21
Solution 1.0

MykolaMazh:
1. repetition of       
 ``` 
            return Distance(self.km ... other.km)
            return Distance(self.km ... other)
 ```
   confuses me a little but i 'm not not sure it's possible to get rid of it

 2. I assume type_error() is not necessary in this task but ...

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,82 @@
+from __future__ import annotations
+
+
 class Distance:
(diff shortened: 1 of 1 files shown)


viktoria-rybenchuk (approved):
GJ!

PR #807, opened 2023-10-21
Solution

AnyoneClown:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: int | Distance) -> True | False:
```
viktoria-rybenchuk, line 32:
```suggestion
    def __lt__(self, other: int | Distance) ->bool:
```

viktoria-rybenchuk: Fix in all places

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
```
viktoria-rybenchuk, line 19: `other` could be float as well

viktoria-rybenchuk (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #808, opened 2023-10-22
'Solution'

OLENA-KALITSINSKA:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        other_distance = other.km \
+            if isinstance(other, Distance) else other
+        return self.km <= other_distance
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        other_distance = other.km \
+            if isinstance(other, Distance) else other
+        return self.km >= other_distance
```
Dimosphen1, line 58: You could think over the implementation of some methods basing them on the logic of others. E.g. `self` is greater or equal to `other` when `self` is not less than `other`.

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        other_distance = other.km \
+            if isinstance(other, Distance) else other
```
Dimosphen1, line 37: Reformat your code to avoid using line breaking `\`. Without it, one line does not seem to be too long to require line breaking.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done

PR #809, opened 2023-10-22
solution

Andrey1104:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> Any:
```
Dimosphen1, line 15: Be more specific in your type annotations and indicate that `other` could be either an `int`, `float`, or `Distance` instance and that the return type would be always a `Distance` class instance. Make such changes to other methods where applicable.

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Any) -> bool:
```
Dimosphen1, line 55: You could implement some methods based on the implemented functionality of others. E.g. `self` is greater or equal to `other` when `self` is not less than `other`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: (int, float, Distance)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
viktoria-rybenchuk, line 25: write this code with one return

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, Distance)) -> Distance:
```
viktoria-rybenchuk, line 15:
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

viktoria-rybenchuk: fix in all places

viktoria-rybenchuk (changes requested):

PR #810, opened 2023-10-22
Solution

TymurSh:


app/main.py
```diff
@@ -1,3 +1,70 @@
+from __future__ import annotations
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+        self.km = km
+
+    def __str__(self,) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self,) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> Distance:
```
ArsenPidhoretskyi, line 16: Really other can be Any? Is code willl works with str? Or tuple?

ArsenPidhoretskyi: Fix annotations in all place

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
```suggestion
       return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+        if isinstance(other, Distance):
+            pass
+        else:
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
ArsenPidhoretskyi, line 46:
```suggestion
        return self.km < other
```
Fix in other places. else: redundant

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Any) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 28: Move return after if-else statement

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+
+    def __mul__(self, other: Any) -> Distance:
+        if isinstance(other, Distance):
+            pass
+        else:
+            return Distance(self.km * other)
```
ArsenPidhoretskyi, line 34:
```suggestion
        return Distance(self.km * other)
```
Check is redundant
Fix the same into truediv

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #815, opened 2023-10-25
Done

Moonralex:


app/main.py
```diff
@@ -1,3 +1,79 @@
+from typing import TypeVar
+
+#  "I'm not sure if this is the best solution"
+#  " for creating accurate type annotations, "
+#  "but the others are not working for me."
+
+T = TypeVar("T", bound="Distance")
```
viktoria-rybenchuk, line 7: this is not a good idea. You can use `Distance`  for type annotation

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+            return Distance(self.km + other)
+        raise TypeError("Unsupported operand types")
+
+    def __iadd__(self, other: T | int | float) -> T:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
```
viktoria-rybenchuk, line 33: write it with one return

viktoria-rybenchuk (changes requested):

app/main.py
Moonralex: When I attempt to implement type annotations like this, I encounter an error when running pytest, as indicated in the screenshot.

Dimosphen1: Place `from __future__ import annotations` on the first line, it should resolve the issue. In case not, make sure that you have Python 3.11 installed and bound to the IDE to utilize the latest functionality. You should use the plain `Distance` in the type annotation. If all steps described above are not successful, visit one of the Q&A sessions to resolve the issue.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #816, opened 2023-10-25
'Solution'

aleksandra-prognimak:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __gt__(self, other: Distance | int | float) -> bool:
+        return self.km > self.get_km(other)
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == self.get_km(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= self.get_km(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
```
LiudmylaKulinchenko, line 44: You can rewrite these two methods using previously created magic methods
(<= is the same as `not >`)

LiudmylaKulinchenko (approved):
Well done 🐣

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > self.get_km(other)
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        return self.km == self.get_km(other)
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self.__lt__(other)
```
LiudmylaKulinchenko, line 45: U can just use operands here instead of calling method directry

LiudmylaKulinchenko (changes requested):

PR #818, opened 2023-10-26
Solution

eLQeR:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Distance) -> Distance:
```
YukihiroSM, line 14: other can be int or float as well

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            self.km = self.km + other.km
+            return self
+        self.km = self.km + other
+        return self
+
+    def __mul__(self, other: Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
```
YukihiroSM, line 30: other can be only int or float, not Distance

app/main.py
YukihiroSM: Fix typehinting everywhere

YukihiroSM (changes requested):

app/main.py
YukihiroSM: this comment is related to these 2 cases only

NikOneZ1 (approved):
Good job!

PR #820, opened 2023-10-26
'solution'

boris-devs:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __lt__(self, other: int) -> bool:
```
EdAlekseiev, line 15: `other`  can be float and Distance too. Please, rewrite your functions.

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return True if self.km < other else False
```
ArsenPidhoretskyi, line 16: In this case you can do
```suggestion
        return self.km < other
```
Fix in all places

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
```
ArsenPidhoretskyi, line 39: Move return after if-else statement

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
```
NikOneZ1, line 47: You need to check if the other is a distance, or an int, or a float. If it's an instance of distance, then you'll have to operate with other, not just other. Do it in other function too.
```suggestion
    def __mul__(self, other: int | float | Distance) -> Distance:
        if isinstance(other, Distance):
            return Distance(self.km * other.km)
        return Distance(self.km * other)
```

NikOneZ1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+                else Distance(self.km + other))
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        self.km += other.km if self.is_obj(other) else other
+        return self
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
```
EdAlekseiev, line 46: `other` can't be Distance in this function

EdAlekseiev, line 46:
```suggestion
    def __mul__(self, other: int | float) -> Distance:
```

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Good job!

PR #821, opened 2023-10-26
Solution

yanamushka:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __add__(self, other: Union[Distance, int]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[Distance, int]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
ArsenPidhoretskyi, line 24: Move return self after if-else statement

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[Distance, int]) -> Distance:
```
ArsenPidhoretskyi, line 15: float also possible in all places

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __iadd__(self, other: Union[Distance, int]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
+
+    def __mul__(self, other: Union[Distance, int]) -> Distance:
```
ArsenPidhoretskyi, line 29:
```suggestion
    def __mul__(self, other: Union[float, int]) -> Distance:
```
Distance is not possible in __mul__ and __truediv__

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[Distance, int, float]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
vasylhnatiuk, line 18:
```suggestion
        return Distance(self.km + other)
```

vasylhnatiuk: you can avoid useless else, please do it

vasylhnatiuk (changes requested):

vsmutok (approved):
Great  👍

PR #822, opened 2023-10-27
Solution

Marinel444:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
+
```
viktoria-rybenchuk, line 27: Prefer using a single return statement for clarity

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
```
viktoria-rybenchuk, line 59: To simplify the code for the  method, you can use the concise approach by making use of the `not self > other` expression.  The same you can make in `__ge__` method

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+
```
vsmutok, line 40:
```suggestion
    def __lt__(self, other: Distance | int | float) -> bool:
        if isinstance(other, Distance):
            return self.km < other.km
        elif isinstance(other, (int, float)):
            return self.km < other
        else:
            raise TypeError("some message")

```

vsmutok: fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Great  👍

PR #827, opened 2023-11-01
Solution

leoleiden:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type")
```
pavlejviki, line 26: why in some methods you raise TypeError if the operand is not of an expected type but in the others not? like def __add__ or def __truediv__? try to be consistent, but you can avoid raising this error

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return self
+
+    def __mul__(self, scalar: int | float) -> Distance:
+        if isinstance(scalar, (int, float)):
+            return Distance(self.km * scalar)
+        else:
```
pavlejviki, line 32: redundant else

pavlejviki (changes requested):

app/main.py
leoleiden: Fixed!

Polyakiv-Andrey (approved):
Good job!

PR #828, opened 2023-11-01
Solution

Evheniia96:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
YukihiroSM, line 32: Other can't be Distance. Fix typehinting

YukihiroSM (changes requested):

PR #829, opened 2023-11-01
Task done

aleksandr-paslawski:


app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 6: float also possible

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(second, Distance):
+            return Distance(km=self.km + second.km)
+        return Distance(km=self.km + second)
+
+    def __iadd__(self, second: Distance | int | float) -> Distance:
+        if isinstance(second, Distance):
+            self.km = self.km + second.km
+            return self
+        self.km = self.km + second
+        return self
```
Yurnerroo, line 25: make 1 return

Yurnerroo (changes requested):

PR #830, opened 2023-11-02
'Solution'

Chacer-Rider:


app/main.py
```diff
@@ -1,3 +1,37 @@
...
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: float) -> object:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> object:
+        return Distance(round(self.km / other, 2))
+
```
pavlejviki, line 23: in __mul__ and __truediv__ methods other can be int or float, in the rest of the methods int, float or an  instance of the Distance class. 

Add logic to handle cases when other is an  instance of the Distance class and fix type hints.

pavlejviki (changes requested):

app/main.py
Chacer-Rider: I avoided unnecessary logic, as I was advised last time, there were 110 questions in the test, which my solution passed successfully, how can I understand where the logic is excessive and where it is insufficient?

app/main.py
```diff
@@ -1,3 +1,43 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> object:
```
Dimosphen1, line 14: Change your type annotation to use the latest syntaxis. Besides that, be more specific in the return type annotation, in Python everything is an object:
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```
Apply this change to all places where applicable. In case you face incompatibility errors, use `from __future__ import annotation`.

Dimosphen1 (changes requested):

PR #833, opened 2023-11-02
Solution

KwintDimon:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
viktoria-rybenchuk, line 11: write correct type annotation for `other` it could be `int` and `float` as well

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+            raise TypeError("Unsupported operand type")
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type")
+        return self  # Вернем тот же экземпляр
```
viktoria-rybenchuk, line 26: remove redundant comment

app/main.py
viktoria-rybenchuk, line 11:
```suggestion
    def __add__(self, other: Distance) -> Distance:
```

viktoria-rybenchuk (changes requested):
correct type annotation in all places

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type")
+        return self
+
+    def __mul__(self, other: Distance | int) -> Distance:
```
pavlejviki, line 31:
```suggestion
    def __mul__(self, other:float | int) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __truediv__(self, other: Distance | int) -> Distance:
```
pavlejviki, line 37:
```suggestion
    def __truediv__(self, other: float | int) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __gt__(self, other: Distance | int) -> bool:
```
pavlejviki, line 51: other can be float as well. fix in all places

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __le__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
```
pavlejviki, line 71: Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,3 +1,81 @@
...
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
```
pavlejviki, line 76: Use previously created magic methods

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
viktoria-rybenchuk, line 24: use a single return statement

app/main.py
viktoria-rybenchuk: not fixed

viktoria-rybenchuk (changes requested):

app/main.py
KwintDimon: why not ?I not understand not fixed 73-76
            return self.km >= other.km
        return self.km >= other

app/main.py
Dimosphen1: You could use the implementation of other magic methods to create new ones, e.g. `self` is greater or equal to `other` when `self` is not less than `other`.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #835, opened 2023-11-04
solution for py_distance_class

VladyslavBazhyn:


app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+    def __str__(self) -> str:
+
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (Distance, int)) -> Distance:
```
EdAlekseiev, line 18: `other` can be `float` too
```suggestion
    def __add__(self, other: (Distance, int, float)) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        new_km_amount = self.km + other
+
+        return Distance(new_km_amount)
+
+    def __iadd__(self, other: (Distance, int, float))\
+            -> (Distance, int, float):
```
EdAlekseiev, line 28: Invalid type hint for return value
```suggestion
    def __iadd__(self, other: (Distance, int, float)) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        self.km += other
+
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
EdAlekseiev, line 37: `other` can be `float` too

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        new_km_amount = self.km * other
+
+        return Distance(new_km_amount)
+
+    def __truediv__(self, other: (int, float)) -> Distance:
+
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: (Distance, int)) -> bool:
```
EdAlekseiev, line 47: The same comment as above

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        return self.km < other
+
+    def __gt__(self, other: (Distance, int)) -> bool:
```
EdAlekseiev, line 54: The same comment as above

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        return self.km > other
+
+    def __eq__(self, other: (Distance, int)) -> bool:
```
EdAlekseiev, line 61: The same comment as above

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        return self.km == other
+
+    def __le__(self, other: (Distance, int)) -> bool:
```
EdAlekseiev, line 68: The same comment as above

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        return self.km <= other
+
+    def __ge__(self, other: (Distance, int)) -> bool:
```
EdAlekseiev, line 75: The same comment as above

EdAlekseiev (changes requested):

app/main.py
VladyslavBazhyn: Thank you, for your work

app/main.py
```diff
@@ -1,3 +1,79 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 6: km: float also possible

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+    def __str__(self) -> str:
+
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (Distance, int, float)) -> Distance:
```
ArsenPidhoretskyi, line 18:
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```
Use Union or |
Fix in all places

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+        if isinstance(other, Distance):
+            other = other.km
+
+        new_km_amount = self.km + other
+
+        return Distance(new_km_amount)
+
+    def __iadd__(self, other: (Distance, int, float)) -> (Distance):
```
ArsenPidhoretskyi, line 27:
```suggestion
    def __iadd__(self, other: (Distance, int, float)) -> Distance:
```

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Good job!

PR #836, opened 2023-11-05
Solution

Vityk-4W:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, int) or isinstance(other, float):
+            return Distance(self.km + other)
```
YukihiroSM, line 18:
```suggestion
        return Distance(self.km + other)
```

YukihiroSM: Also if you want to check two or more types you can use `isinstance(other, (float, int))`
But in this case it's not needed

YukihiroSM: Fix it everywhere

YukihiroSM (changes requested):

app/main.py
Vityk-4W: fixed everything

Vityk-4W:
Fixed

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 23:
```suggestion
        else:
```

ArsenPidhoretskyi (approved):
Well done!

PR #837, opened 2023-11-06
Solution

vpanchenkoSPS:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Union[Distance, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Union[Distance, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
fsocie7y, line 56: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __eq__(self, other: Union[Distance, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Union[Distance, float]) -> bool:
+        return not self.km > other
+
+    def __ge__(self, other: Union[Distance, float]) -> bool:
+        return self.km >= other
```
fsocie7y, line 52: Change it in the same way as __le__

fsocie7y (changes requested):

fsocie7y (approved):
well done 🥇

PR #839, opened 2023-11-09
Solution

OleksiiMi:


app/main.py
```diff
@@ -1,3 +1,69 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 6: float also possible here

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+    def __add__(
+            self, other: Union[int, float, "Distance"]
+    ) -> Union[float, "Distance"]:
+
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: [int, float]) -> Union[float, "Distance"]:
```
ArsenPidhoretskyi, line 24:
```suggestion
    def __iadd__(self, other: Union[int, float, "Distance"]) -> Union[float, "Distance"]:
```

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: [int, float]) -> Union[float, "Distance"]:
```
ArsenPidhoretskyi, line 31:
```suggestion
    def __mul__(self, other: Union[int, float]) -> Union[float, "Distance"]:
```
Fix in all places

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: [int, float]) -> Union[float, "Distance"]:
+        if other == 0:
+            raise ZeroDivisionError("Division by zero is not possible")
+        elif isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: [int, float]) -> bool:
```
ArsenPidhoretskyi, line 41: Distance also possible
Fix in all boolean method

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(
+            self, other: Union[int, float, "Distance"]
+    ) -> Union[float, "Distance"]:
+
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
ArsenPidhoretskyi, line 22: elif is redundant here because it is always true due to type annotations
```suggestion
        if isinstance(other, Distance):
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```
Please fix in all places(mul, truediv, all boolean methods)

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #845, opened 2023-11-13
Distance class

Ananiev-Vitalii:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance or int or float) -> Distance:
```
EdAlekseiev, line 14: You can not use `or` in type hint. Fix it here and everywhere.
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
EdAlekseiev, line 18: `else` block is redundant, because you call return method inside if block. Fix it here and everywhere.
```suggestion
        return Distance(self.km + other)
```

EdAlekseiev (changes requested):

PR #847, opened 2023-11-14
solution

PyCore23:


app/main.py
```diff
@@ -1,3 +1,98 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"{Distance.__name__}: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"{Distance.__name__}(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]
+                ) -> Union["Distance", TypeError]:
```
viktoria-rybenchuk, line 15:
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

viktoria-rybenchuk: this  approach using `Distance | int | float` is concise, clear, and follows the modern conventions of type hinting in Python, providing a straightforward and expressive way

viktoria-rybenchuk: Fix in all places

app/main.py
```diff
@@ -1,3 +1,98 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
viktoria-rybenchuk, line 20: `else` is redundant

app/main.py
```diff
@@ -1,3 +1,98 @@
...
+                 ) -> Union[None, TypeError]:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
+        else:
+            raise TypeError(
+                f"{other} must be {Distance.__name__} or (int, float)")
```
viktoria-rybenchuk, line 34: write another function for checking value and use it in order to avoid duplication

app/main.py
```diff
@@ -1,3 +1,98 @@
...
+
+    def __truediv__(self, other: Union["Distance", int, float]
+                    ) -> Union["Distance", TypeError]:
+        if isinstance(other, Distance):
+            raise TypeError(
+                f"{other} must be {Distance.__name__} or (int, float)")
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: Union["Distance", int, float]
+               ) -> Union["Distance", TypeError]:
```
viktoria-rybenchuk, line 51: wrong type annotation for return value

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
viktoria-rybenchuk, line 55: Does this functions return `Distance`?

viktoria-rybenchuk (changes requested):
Rollback all changes in `README.md`

PR #851, opened 2023-11-16
ready solution!

job-artem:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, (float, int)):
+            return self.km >= other
+        if isinstance(other, Distance):
+            return self.km >= other.km
+
+
+kilometers1, kilometers2 = 100, 30
+distance = Distance(kilometers1)
+print(distance == kilometers2)
```
viktoria-rybenchuk, line 65: remove this code

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, (float, int)):
+            return self.km <= other
+        if isinstance(other, Distance):
+            return self.km <= other.km
+
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, (float, int)):
+            return self.km >= other
+        if isinstance(other, Distance):
+            return self.km >= other.km
```
viktoria-rybenchuk, line 60: you can write these two methods in one line each. Use previously created magic methods

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, (float, int)):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: Distance | float | int) -> Distance:
```
pavlejviki, line 26: in __mul__ and __truediv__  other is not expected to be an instance of the Distance class, fix the type hints

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float | int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, (float, int)):
+            return self.km < other
+        if isinstance(other, Distance):
```
pavlejviki, line 35: the second if condition is quite redundant as according to the type annotation if other is not float or int, it will be an instance of the Distance class

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return self.km < other.km
+
+    def __gt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, (float, int)):
+            return self.km > other
+        if isinstance(other, Distance):
```
pavlejviki, line 41: see comment above

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return self.km > other.km
+
+    def __eq__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, (float, int)):
+            return self.km == other
+        if isinstance(other, Distance):
```
pavlejviki, line 47: see comment above

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
vsmutok, line 5:
```suggestion
    def __init__(self, km: int | float) -> None:
```

vsmutok (approved):
Great  👍

PR #852, opened 2023-11-17
Distance class correct

Ananiev-Vitalii:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
EdAlekseiev, line 57: These methods return bool. Improve type hint, please

EdAlekseiev (changes requested):

PR #853, opened 2023-11-17
'Solution' py-distance-class

yura-push:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(
+            km=self.km * other
+        )
+
+    def __truediv__(self, other: float) -> Distance:
+        km = round(self.km / other, 2)
```
pavlejviki, line 34: no need to create this variable

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        return Distance(
+            km=km
+        )
+
+    def __lt__(self, other: int) -> bool:
```
pavlejviki, line 39: other can be float or an instance of the Distance class.

Add logic to handle cases when other is an instance of the Distance class

fix it in all the methods below as well

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> repr:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> float | Distance:
```
Dimosphen1, line 14: Only the `Distance` class instance is returned in this method, make changes in your code where applicable to encounter that.

Dimosphen1 (changes requested):
Several changes in type annotations were requested.

Dimosphen1 (approved):
LGTM, well done!

PR #855, opened 2023-11-19
Solution

Grotski:


app/main.py
```diff
@@ -1,3 +1,69 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 3: km can be float

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int():
```
ArsenPidhoretskyi, line 12: other can be float and Distance also

ArsenPidhoretskyi, line 12: Wrong return type annotations it returns Distance
Also, write int() is so bad....

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, Distance):
+            return Distance(
+                self.km + other.km
+            )
+        elif isinstance(other, (int, float)):
+            return Distance(
+                self.km + other
+            )
+
+    def __iadd__(self, other: int) -> int():
```
ArsenPidhoretskyi, line 22: The same as with add

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+
+    def __add__(self, other: int) -> int():
+        if isinstance(other, Distance):
+            return Distance(
+                self.km + other.km
+            )
+        elif isinstance(other, (int, float)):
+            return Distance(
+                self.km + other
+            )
```
```suggestion
        return Distance(
            self.km + other
        )
```

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+
+    def __iadd__(self, other: int) -> int():
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> float():
+        if isinstance(other, (int, float)):
```
ArsenPidhoretskyi, line 30: This check is redundant it is always true due to type annotations

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+            )
+
+    def __iadd__(self, other: int) -> int():
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> float():
```
ArsenPidhoretskyi, line 29: Wrong return type annotations

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(
+                self.km * other
+            )
+
+    def __truediv__(self, other: float | int) -> float():
```
ArsenPidhoretskyi, line 35: The same errors as with mul

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(
+                round(self.km / other, 2)
+            )
+
+    def __lt__(self, other: int) -> bool:
```
ArsenPidhoretskyi, line 41: other can be float and Distance also

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
```
ArsenPidhoretskyi, line 45:
```suggestion
        return self.km < other
```

ArsenPidhoretskyi (changes requested):
Check errors in all places please

app/main.py
Grotski: yep I write Distance but PyCharm give me ERROR

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        return Distance(
+            round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other: float | Distance) -> bool:
```
EdAlekseiev, line 46: `other` can be `int`

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: float | Distance) -> bool:
```
EdAlekseiev, line 51: `other` can be `int`

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: float | Distance) -> bool:
```
EdAlekseiev, line 56: `other` can be `int`

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: float | Distance) -> bool:
```
EdAlekseiev, line 61: `other` can be `int`

EdAlekseiev (approved):
Good job

PR #856, opened 2023-11-19
Solution

brunhildaaaaa:


app/main.py
```diff
@@ -1,3 +1,74 @@
...
+    def __le__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
```
vsmutok, line 73: Redundant  `else`

vsmutok: fix everywhere

app/main.py
```diff
@@ -1,3 +1,74 @@
...
+            return self.km == other
+
+    def __le__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Any) -> bool:
```
vsmutok, line 70:
```suggestion
    def __ge__(self, other: int | float | Distance) -> bool:
```

vsmutok: fix everywhere

app/main.py
```diff
@@ -1,3 +1,74 @@
+from __future__ import annotations
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+    distances = {}
+
+    def __init__(self, run: float) -> None:
```
vsmutok, line 8:
```suggestion
    def __init__(self, run: float | int) -> None:
```

app/main.py
```diff
@@ -1,3 +1,74 @@
...
+        self.km = run
+        Distance.distances[self.km] = self
```
vsmutok, line 10: why you need it?

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            total_distance = self.km + other.km
+        else:
+            total_distance = self.km + other
+        result = Distance(total_distance)
+        return result
```
Dimosphen1, line 21: There is no need for the `result` variable, you should return a `Distance` class instance directly here and in other places to not allocate additional memory to store this instance.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
pavlejviki, line 53: use previously created magic methods to simplify the code

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
```
pavlejviki, line 56: use previously created magic methods to simplify the code

pavlejviki (changes requested):

PR #858, opened 2023-11-19
Solution

gertsovenko:


app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Dimosphen1, line 5: `km` could be `float` as well, add it in the type annotation here and across other methods where `int` is used.

Dimosphen1 (changes requested):
One minor change was requested.

PR #861, opened 2023-11-20
solution

AlisaSkl:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
Dimosphen1, line 14: Indicate that `other` or `number` could be `float` in the type annotation here and in other methods as well.

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, number: int) -> bool:
+        return Distance(km=self.km * number)
+
+    def __truediv__(self, number: int) -> bool:
+        return Distance(km=round(self.km / number, 2))
```
Dimosphen1, line 30: Return type annotation is incorrect for these methods.

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > number
+
+    def __eq__(self, number: int) -> bool:
+        return self.km == number
+
+    def __le__(self, number: int) -> bool:
+        return self.km <= number
+
+    def __ge__(self, number: int) -> bool:
+        return self.km >= number
```
Dimosphen1, line 45: You should make a comparison against `other` which could be `Distance` class instance, `int`, or `float` as mentioned in the task example.

Dimosphen1 (approved):
LGTM, well done!

PR #862, opened 2023-11-21
Solution

Vlad1slavo:


app/main.py
```diff
@@ -1,3 +1,54 @@
 class Distance:
-    # Write your code here
+    def __init__(self, km: int) -> None:
```
ArsenPidhoretskyi, line 2: float also possible

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
```
ArsenPidhoretskyi, line 11: other also can be Distance
Fix annotations, please check all places

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
ArsenPidhoretskyi, line 15:
```suggestion
        return Distance(self.km + other)
```
Due to type annotations it is always true

ArsenPidhoretskyi: Fix the same in other places

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __iadd__(self, other: int | float) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
```suggestion
        else:
```

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
ArsenPidhoretskyi, line 49:
```suggestion
        return self < other or self == other
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #866, opened 2023-11-25
solution

rozovas:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, multiplier: int | float) -> Distance:
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, other: int | float) -> Distance | str:
+        if other != 0:
+            return Distance(round(self.km / other, 2))
+        return "Zero divider"
```
ArsenPidhoretskyi, line 32: Better to raise exception

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
pavlejviki, line 52: Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
```
pavlejviki, line 55: Use previously created magic methods

pavlejviki (changes requested):

app/main.py
rozovas: Thanks, indeed it looks better. Committed the changes.

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
```
pavlejviki, line 50:
```suggestion
        return not self > other
```

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self.__lt__(other)
```
pavlejviki, line 53:
```suggestion
        return not self < other
```

app/main.py
pavlejviki, line 50: You can further simplify the methods by directly using the < (less than) and > (greater than) operators.

pavlejviki (changes requested):

PR #867, opened 2023-11-26
Class magic methods!

itpsychologist:


app/main.py
```diff
@@ -1,3 +1,62 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: None) -> None:
```
ArsenPidhoretskyi, line 3: km is not None

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: None) -> object:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: None) -> object:
+        result = self.km / other
+        if isinstance(other, Distance):
```
ArsenPidhoretskyi, line 30: Due to task requirements it is always false

ArsenPidhoretskyi (changes requested):
other: None is wrong type annotations

app/main.py
```diff
@@ -1,3 +1,60 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> object:
```
Dimosphen1, line 3: `__init__` returns `None`, indicate it in the return type annotation instead of `object`. Besides that, `km` could be `float` as well, add it to the type annotation for `km` using `|`.

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return "Distance(km=" + str(self.km) + ")"
+
+    def __add__(self, other: None) -> object:
```
Dimosphen1, line 12: Indicate in the type annotation that `other` could be `int`, `float`, or an instance of the `Distance` class. Change your return type annotation to indicate that you return an instance of the `Distance` class. Revise all your methods and change your type annotation to display correct incoming and outgoing types.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return "Distance(km=" + str(self.km) + ")"
+
+    def __add__(self, other: int | float | object) -> object:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
viktoria-rybenchuk, line 15: else is redundant

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | object) -> object:
+        result = self.km / other
+        return Distance(round(result, 2))
+
+    def __lt__(self, other: int | float | object) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
viktoria-rybenchuk, line 35: the same here

viktoria-rybenchuk: fix in all places

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #868, opened 2023-11-26
Solution

NataliiaKucher:


app/main.py
```diff
@@ -1,3 +1,86 @@
...
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(km=self.km + other)
+        else:
+            raise TypeError(
+                f"isinstance({type(other)}, (Distance, int, float)) is True")
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
EdAlekseiev, line 20: `other` can be int and float too. Fix it everywhere
```suggestion
    def __iadd__(self, other: int | float | "Distance") -> "Distance":
```

app/main.py
```diff
@@ -1,3 +1,86 @@
...
+                f"isinstance({type(other)}, (int, float)) is True"
+            )
+
+    def __eq__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        else:
+            return False
```
EdAlekseiev, line 54: Why do you return `False` here?

app/main.py
```diff
@@ -1,3 +1,86 @@
...
+
+    def __lt__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        else:
+            return NotImplemented
```
EdAlekseiev, line 62: Why do you raise `TypeError` in other function for the same case?

EdAlekseiev (changes requested):

PR #874, opened 2023-12-04
'solution-distance'

Binkovskyi:


app/main.py
```diff
@@ -1,3 +1,62 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
vsmutok, line 3:
```suggestion
    def __init__(self, km: int | float) -> None:
```

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
vsmutok, line 12:
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+
+    def __le__(self, other: "Distance") -> "Distance":
+        if not isinstance(other, Distance):
+            return self.km <= other
+        if isinstance(other, Distance):
+            return self.km <= other.km
+
+    def __ge__(self, other: "Distance") -> "Distance":
```
vsmutok, line 58:
```suggestion
    def __ge__(self, other: Distance | int | float) -> boll:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
Binkovskyi: ok

Binkovskyi: But i can't use class name as an annotation without ""

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: type["Distance"] | int | float) -> "Distance":
```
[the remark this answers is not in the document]
Binkovskyi: E   TypeError: unsupported operand type(s) for |: 'str' and 'type'

```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vasylhnatiuk (changes requested):

Binkovskyi:
Error

Binkovskyi:
Error

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #875, opened 2023-12-04
distance solution

OleksandrHubar:


app/main.py
```diff
@@ -1,3 +1,71 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Dimosphen1, line 5: Indicate that `km` could be `int` as well in the type annotation.

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", float]) -> "Distance":
```
Dimosphen1, line 14: Do not use `Union` and do not wrap `Distance` in quotes. Besides that, indicate that `other` could `int` as well: 
```suggestion
    def __add__(self, other: Distance | float | int) -> Distance:
```
Apply this change to all type annotations where applicable. In case you face errors related to the type annotation, add `from __future__ import annotations`.

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
Dimosphen1, line 17: There is no need for `else` here since the code will not execute after the `return` statement anyway. Remove `else`.

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[int, float]) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float]) -> "Distance":
+        result = self.km / other
+        return Distance(round(result, 2))
```
Dimosphen1, line 32: You could omit creating a `result` variable here. Pass the expression directly to the `round`.

Dimosphen1 (changes requested):

PR #876, opened 2023-12-04
Solution

yablonskyiv:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> None:
```
EdAlekseiev, line 12: `other` can be float and Distance too. Fix everywhere 
```suggestion
    def __add__(self, other: int | float | Distance) -> None:
```

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __iadd__(self, other: int) -> None:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise ValueError("Unsupported operand type")
+        return self
+
+    def __mul__(self, scalar: None) -> None:
```
EdAlekseiev, line 29: Incorrect type hint for `scalar`

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return Distance(self.km * scalar)
+        else:
+            return NotImplemented
+
+    def __truediv__(self, divisor: int) -> None:
+        if isinstance(divisor, (int, float)):
+            result = self.km / divisor
+            return Distance(round(result, 2))
+        else:
+            return NotImplemented
```
EdAlekseiev, line 40: Why do you raise `NotImplemented` here?

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __lt__(self, other: int) -> None:
+        return self.km < other.km if isinstance(other, Distance) else \
+            self.km < other
```
EdAlekseiev, line 44: Use `()` instead of `\`. Fix everywhere

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> None:
```
EdAlekseiev, line 12: Still not fixed. `other` can be `Distance`. Fix everywhere.
```suggestion
    def __add__(self, other: int | float | Distance) -> None:
```

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+            result = self.km / divisor
+            return Distance(round(result, 2))
+        else:
+            raise TypeError("Unsupported operand type")
+
+    def __lt__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
EdAlekseiev, line 46: `else` block is redundant.
```suggestion
    def __lt__(self, other: int | float) -> None:
        if isinstance(other, Distance):
            return self.km < other.km
        return self.km < other
```

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+
+    def __gt__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
```
EdAlekseiev, line 52: The same comment as above

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+
+    def __eq__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
```
EdAlekseiev, line 58: The same comment as above

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+
+    def __le__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
```
EdAlekseiev, line 64: The same comment as above

app/main.py
```diff
@@ -1,3 +1,70 @@
...
+
+    def __ge__(self, other: int | float) -> None:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
EdAlekseiev, line 70: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #877, opened 2023-12-07
Solution

Olevsky-Ivan:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: "Distance") -> "Distance":
```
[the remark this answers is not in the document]
viktoria-rybenchuk: fix in all places

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return self
```
viktoria-rybenchuk, line 16: why do you need return `self` here?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other
+        elif isinstance(other, int | float):
+            return self.km <= other
+
+    def __ge__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return self.km >= other
+        elif isinstance(other, int | float):
+            return self.km >= other
```
viktoria-rybenchuk, line 65: you can write these two methods in one line each. Use previously created magic methods

PR #880, opened 2023-12-07
Solutin

maxbidun:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
+        if not isinstance(other, Distance):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int | float) -> "Distance":
```
```suggestion
    def __iadd__(self, other: int | float) -> Distance:
```

vasylhnatiuk: change evrywhere

vasylhnatiuk (changes requested):

app/main.py
maxbidun: if i do i get an error NameError: name 'Distance' is not defined. Did you mean: 'isinstance'?

maxbidun: Unreslved reference "Distance"

vasylhnatiuk: add `from __future__ import annotations` in the beginning of file

app/main.py
```diff
@@ -1,3 +1,52 @@
 class Distance:
```
vasylhnatiuk, line 1:
```suggestion
from __future__ import annotations

class Distance:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #881, opened 2023-12-09
completed

Volokitin-Vitaliy:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> "Distance":
```
viktoria-rybenchuk, line 14: `other` cannot be `Any`. It should be `int | float | Distance`

viktoria-rybenchuk: fix in all places

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
viktoria-rybenchuk, line 64: you can write these two methods in one line each. Use previously created magic methods

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, factor: float) -> "Distance":
+        return Distance(self.km * factor)
+
+    def __truediv__(self, divisor: float) -> "Distance":
+        if divisor != 0:
+            return Distance(round(self.km / divisor, 2))
+        else:
```
viktoria-rybenchuk, line 33: `else` is redundant

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (changes requested):
FIx according to the comments

pavlejviki (approved):
Well done!

PR #882, opened 2023-12-09
Solution

dissom:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
```
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

vasylhnatiuk: use without "

vasylhnatiuk: the same everywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            if other != 0:
```
vsmutok, line 64:
```suggestion
        if isinstance(other, (int, float)) and other != 0:
```

vsmutok: it is better to write "and" instead of two "if"

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)) and other != 0:
+                return Distance(round((self.km / other), 2))
```
vsmutok:
```suggestion
        if isinstance(other, (int, float)) and other != 0:
            return Distance(round((self.km / other), 2))
```

vsmutok (changes requested):
please fix tests

vsmutok (approved):
Well Done!

PR #887, opened 2023-12-12
Solution

Zonda001:


app/main.py
```diff
@@ -1,3 +1,72 @@
...
+    def __init__(self, km: float) -> float:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

app/main.py
Zonda001: but, logical test falied

vasylhnatiuk (approved):

PR #889, opened 2023-12-14
Solution

MrsLFG:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance:
```
EdAlekseiev, line 14: `other` can be `float` and `Distance` too. Fix everywhere.
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

EdAlekseiev (changes requested):

PR #890, opened 2023-12-14
Solution

aLEKS-e3:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Callable) -> Callable:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km = Distance(self.km + other)
+        return self
```
NEkropo1, line 24: I thought self.km should be a value, not a Distance instance

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Callable) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __truediv__(self, other: int) -> Callable:
```
NEkropo1, line 54: I guess here could be both int and float

NEkropo1 (changes requested):
Consider this info:
@functools.total_ordering: This decorator from the functools module significantly simplifies the implementation of comparison methods in a class. When applied to a class, you only need to define the __eq__ method and one other comparison method (such as __lt__). The total_ordering decorator then automatically provides implementations for the rest of the comparison methods (__le__, __gt__, __ge__, __ne__). This leads to cleaner, more maintainable code, especially in classes where multiple comparison methods would otherwise need to be manually implemented.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Callable) -> Callable:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
```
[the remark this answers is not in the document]
aLEKS-e3: if I return a specific class then i should return that class, right? Please, say that I got it the right way)

app/main.py
aLEKS-e3: my bad

kabust (approved):
Good job! You can also try return type annotations without double quotes in case of class instance though (Distance instead of "Distance")

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Callable | int | float) -> "Distance":
```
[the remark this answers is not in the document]
aLEKS-e3: For some reason, it was an error when I used just the Distance variable, but I fixed it by importing annotations (even though I thought it was default in Python 3.10 and above).

NEkropo1 (approved):
LGTM!

PR #893, opened 2023-12-16
Solution

Roman-Sokolov-V:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if type(other) == Distance:
+            return True if self.km == other.km else False
+        else:
+            return True if self.km == other else False
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if type(other) == Distance:
+            return True if self.km <= other.km else False
+        else:
```
vsmutok, line 53: remove redundant `else`

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            return True if self.km > other else False
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if type(other) == Distance:
+            return True if self.km == other.km else False
+        else:
+            return True if self.km == other else False
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if type(other) == Distance:
```
vsmutok, line 51: better use is `isinstance`

vsmutok (changes requested):

app/main.py
Roman-Sokolov-V: Of course, I'll redo it, but what would make it better?

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.__eq__(other) or self.__lt__(other)
```
Dimosphen1, line 48: It is possible to make a comparison using `not`, e.g. `self` is less or equal to other in case `self` is not greater than `other`. 
```suggestion
        return not self > other
```

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.__eq__(other) or self.__gt__(other)
```
Dimosphen1, line 51: It is possible to make a comparison using `not` and `<` operator
```suggestion
        return not self < other
```

Dimosphen1 (changes requested):
Several minor changes were requested

PR #894, opened 2023-12-17
task completed

cyb1ove:


app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km > Distance.get_addend(other)
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == Distance.get_addend(other)
+
+    def __le__(self, other: int) -> bool:
+        return self.km <= Distance.get_addend(other)
+
+    def __ge__(self, other: int) -> bool:
+        return self.km >= Distance.get_addend(other)
```
fsocie7y, line 44: Use previously created magic methods
Hint: <= is the same as not >

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km > self.get_addend(other)
+
+    def __eq__(self, other: Distance | int) -> bool:
+        return self.km == self.get_addend(other)
+
+    def __le__(self, other: Distance | int) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int) -> bool:
+        return not self.__lt__(other)
```
fsocie7y, line 44: self.__gt__ the same as >
same about self.__lt__

fsocie7y (changes requested):

app/main.py
cyb1ove: What magic methods should I use?

PR #899, opened 2023-12-22
solution

rtsadm:


app/main.py
```diff
@@ -1,3 +1,63 @@
+from __future__ import annotations
+from typing import Callable, Any
+
+
+def _int_float_types_execption(func: Callable) -> Callable:
+    def inner(self: Any, other: Any) -> Callable:
+        if type(other) in (int, float):
+            return func(self, other=Distance(other))
+        else:
```
vsmutok, line 9: remove redundant `else`

app/main.py
```diff
@@ -1,3 +1,63 @@
+from __future__ import annotations
+from typing import Callable, Any
+
+
+def _int_float_types_execption(func: Callable) -> Callable:
+    def inner(self: Any, other: Any) -> Callable:
```
vsmutok, line 6: add `functools.wraps`

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __mul__(self, coefficient: int | float) -> Distance:
+        return Distance(
+            self.km * coefficient
+        )
+
+    def __truediv__(self, divider: int | float) -> Distance:
+        return Distance(
+            round(self.km / divider, 2)
+        )
```
vsmutok, line 43:
```suggestion
        return Distance(round(self.km / divider, 2))
```

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    @_int_float_types_execption
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km += other.km
+        return self
+
+    def __mul__(self, coefficient: int | float) -> Distance:
+        return Distance(
+            self.km * coefficient
+        )
```
vsmutok, line 38:
```suggestion
        return Distance(self.km * coefficient)
```

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @_int_float_types_execption
+    def __add__(self, other: Distance | int | float) -> Distance:
+        return Distance(
+            self.km + other.km
+        )
```
vsmutok, line 28:
```suggestion
        return Distance(self.km + other.km)
```

vsmutok (changes requested):

app/main.py
rtsadm: The code doesn't work without else

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+from functools import wraps
+
+
+def _int_float_types_execption(func: Callable) -> Callable:
+    @wraps(func)
+    def inner(self: Any, other: Any) -> Callable:
+        if type(other) in (int, float):
+            return func(self, other=Distance(other))
+        else:
+            return func(self, other)
```
vsmutok, line 12:
```suggestion
    def inner(self: Any, other: Any) -> Callable:
        if isinstance(other, (int, float)):
            return func(self, other=Distance(other))
        return func(self, other)
```
also better use `isinstance`

vsmutok (changes requested):

app/main.py
rtsadm: Ahh, I understand what you mean.

vsmutok (approved):
Well Done!

PR #900, opened 2023-12-22
Solution

msymonovych:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    @classmethod
+    def validate(cls, obj: Any) -> bool:
+        return isinstance(obj, cls)
```
NEkropo1, line 11: good, but consider more descriptive naming.

NEkropo1 (changes requested):

PR #901, opened 2023-12-23
Solution

mvsydoruk:


app/main.py
```diff
@@ -1,3 +1,54 @@
 class Distance:
     # Write your code here
```
Dimosphen1, line 2: Remove this comment from the pull request.

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, "Distance")) -> "Distance":
```
Dimosphen1, line 12: Change your type annotation to use `|` and class without quotes in the return type annotation:
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```
Use `from __future__ import annotations` import in case you face errors. Change your type annotation in all places.

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: float) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> bool:
```
Dimosphen1, line 31: Add to all methods that return a `bool` value that `other` could be either float, integer, or an instance of the `Distance` class to fix previous comments.

Dimosphen1 (changes requested):

PR #902, opened 2023-12-24
Solution

Disash:


app/main.py
```diff
@@ -1,3 +1,88 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
EdAlekseiev, line 15: other can be int and float too. Fix it everywhere
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #904, opened 2023-12-26
Solution

Judviii:


app/main.py
```diff
@@ -1,3 +1,59 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: Union[int, float, "Distance"]) -> None:
```
vasylhnatiuk, line 6:
```suggestion
    def __init__(self, km: Union[int, float, Distance]) -> None:
```

vasylhnatiuk: use without "

vasylhnatiuk (changes requested):

app/main.py
Judviii: @vasylhnatiuk  If i use Distance without the ", I get 
`:6:46: F821 undefined name 'Distance'` error
And i cant pass tests without "

vasylhnatiuk (approved):

PR #906, opened 2023-12-27
resolved

DonSosiska:


NazarSheremet (commented):
In all methods, the input distance can be either float or int. Modify it

YuriiSmolii (approved):
Good job.

PR #907, opened 2023-12-27
'Solution'

JohnnyWalker010:


ArturPoltser (approved):

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
NEkropo1, line 18: You can simplify the __add__ method by returning the result of the addition directly without separate checks for int, or float. Python will automatically raise a TypeError if the other object is not of a type that supports addition with the types in your method. This simplifies the code and leverages Python's built-in type checking for arithmetic operations.
While your code will do nothing (return None) in case, for example, where we will try to add string to Distance object, which is unexpected behavior.

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __iadd__(self, other: (Distance, int, float)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: (Distance, int, float)) -> (Distance, None):
+        if isinstance(other, Distance):
+            return None
```
NEkropo1, line 29: Why do we need these lines?

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: (Distance, int, float)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: (Distance, int, float)) -> (Distance, None):
```
NEkropo1, line 27: use Union or `|`
Since for now this method should return a tuple, according to your annotation.

NEkropo1 (changes requested):
Think about changing default python behavior, which we don't want to change, like
```python
3 + None
```
And refactor your code)

app/main.py
JohnnyWalker010: Done!

NEkropo1 (changes requested):
Apply the principles and logic that you refactored in the last pushes to the entire app.

NEkropo1 (approved):
Good job!

PR #908, opened 2023-12-27
Solution

Meksim-fesh:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Distance) -> Distance:
```
vasylhnatiuk, line 15: other can be int and float also

vasylhnatiuk (changes requested):

PR #914, opened 2024-01-03
Solution

rzaliznyi:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> int:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> int:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
```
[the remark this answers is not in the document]
vasylhnatiuk: fix it here and evrywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> int:
```
Dimosphen1, line 9: Change your type annotation to indicate that you return a string, not an integer.

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> int:
```
Dimosphen1, line 12: Change your type annotation to indicate that you return a string, not an integer.

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> (int, float):
```
Dimosphen1, line 15: Change your type annotation to use `|` instead of `Union`, `Distance` instead of `"Distance"` and `Distance` class instance in the return type annotation. 
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```
In case you face errors, add `from __future__ import annotations` import. Fix type annotation in all places.

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Union[int, float, "Distance"]) -> (int, float):
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Union[int, float, "Distance"]) -> (int, float):
```
Dimosphen1, line 56: Pay attention in the return type annotation that in some methods you return `bool`.

Dimosphen1 (changes requested):

PR #919, opened 2024-01-05
Solution

Bombardier256:


app/main.py
```diff
@@ -1,3 +1,71 @@
...
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
vasylhnatiuk, line 33: other can be float here

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        return Distance(
+            km=self.km * other
+        )
+
+    def __truediv__(self, other: int) -> Distance:
```
vasylhnatiuk, line 38: other can be float here

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
vasylhnatiuk, line 15: other can be float and int here also

vasylhnatiuk: fix here and everywhere

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
vasylhnatiuk, line 46: you can avoid else here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> Distance:
```
```suggestion
    def __add__(self, other: float | int) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(
+                km=self.km + other.km
+            )
+        else:
+            return Distance(
+                km=self.km + other
+            )
+
+    def __iadd__(self, other: Distance) -> Distance:
```
vasylhnatiuk, line 25:
```suggestion
    def __iadd__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(
+            km=self.km * other
+        )
+
+    def __truediv__(self, other: float) -> Distance:
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __lt__(self, other: float) -> bool:
```
vasylhnatiuk, line 43: you didn't understand me. I meant that you can avoid only `else`, because return ends function.
The correct variant will be:

```
            return self.km > other.km
        return self.km > other
```

vasylhnatiuk: don't forget avout annotations

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #920, opened 2024-01-06
Solution

VladimirDolhyi:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
pavlejviki, line 14: other can be float as well. add to type hints. fix in all places

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
```
pavlejviki, line 50: you can use previously created magic methods to simplify the code.

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
pavlejviki, line 55: you can use previously created magic methods to simplify the code.

pavlejviki (changes requested):

app/main.py
VladimirDolhyi: right?

app/main.py
VladimirDolhyi: right?

vasylhnatiuk (approved):

PR #922, opened 2024-01-06
Solution

MykhailoIvchenko:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __iadd__(self, distance: int) -> Distance:
+        self.km += distance
+
+        return self
+
+    def __mul__(self, distance: int) -> Distance | None:
+        if isinstance(distance, Distance):
+            return None
+
+        return Distance(self.km * distance)
```
YukihiroSM, line 39: You defined, that distance should be only int, so no need to check if it's Distance. 
```suggestion
        return Distance(self.km * distance)
```

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return func(self, distance)
+
+    return wrapper
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```
YukihiroSM, line 17: km might me float/int

app/main.py
```diff
@@ -1,3 +1,65 @@
+from __future__ import annotations
+from typing import Any
+
+
+def get_operand_value(func: callable) -> callable:
+    def wrapper(self: Distance, distance: Distance | int) -> Any:
```
YukihiroSM, line 6: distance might be float as well

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __add__(self, distance: int) -> Distance:
+        return Distance(self.km + distance)
+
+    @get_operand_value
+    def __iadd__(self, distance: int) -> Distance:
+        self.km += distance
+
+        return self
+
+    def __mul__(self, distance: int) -> Distance | None:
```
YukihiroSM, line 35: distance might be float as well. Fix everywhere.

YukihiroSM (changes requested):

app/main.py
MykhailoIvchenko: Fixed. By type, annotation doesn't prevent passing a wrong argument. But I added this check expression because of the tests. But now I see that tests are passed without this check, so it was redundant.

app/main.py
MykhailoIvchenko: Fixed

app/main.py
MykhailoIvchenko: Fixed

app/main.py
MykhailoIvchenko: Fixed

vasylhnatiuk (approved):

PR #923, opened 2024-01-06
Solution

eugenevsn:


YukihiroSM (changes requested):
Fix flake errors, please. (Add typehints)

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km * scalar)
+
+    def __truediv__(self, divisor: float) -> "Distance":
+        return Distance(round(self.km / divisor, 2))
+
+    def __lt__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
```
EdAlekseiev, line 34:
```suggestion
        if isinstance(other, Distance):
            return self.km < other.km
        return self.km < other
```

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+
+    def __eq__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
```
EdAlekseiev, line 46:
```suggestion
        if isinstance(other, Distance):
            return self.km == other.km
        return self.km == other
```

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[float, int, str]) -> "Distance":
```
YukihiroSM, line 14:
```suggestion
    def __add__(self, other: Union[float, int, str]) -> Distance:
```

Fix it everywhere, please. 
Also, import: 
`from __future__ import annotations` for this to work .

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, scalar: float) -> "Distance":
+        return Distance(self.km * scalar)
+
+    def __truediv__(self, divisor: float) -> "Distance":
+        return Distance(round(self.km / divisor, 2))
```
YukihiroSM, line 30: scalar and divisor might be int as well

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __lt__(self, other: Union[float, int, str]) -> bool:
```
YukihiroSM, line 32: it doesn't accept str, it should accept only float, int and Distance
fix it everywhere

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```
YukihiroSM, line 6:
```suggestion
    def __init__(self, km: float | int) -> None:
        self.km = km
```

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float | int | "Distance") -> Distance:
```
YukihiroSM, line 14:
```suggestion
    def __add__(self, other: float | int | Distance) -> Distance:
```

YukihiroSM: Fix it everywhere

YukihiroSM (changes requested):

app/main.py
NikOneZ1: Fix this, please

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #926, opened 2024-01-07
Solution

Yevhen23tk:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
EdAlekseiev, line 11: `other` can be int and float too. Fix everywhere
```suggestion
    def __add__(self, other: "Distance" | int | float) -> "Distance":
```

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        if isinstance(other, Distance):
+            total_distance = self.km + other.km
+            return Distance(total_distance)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
EdAlekseiev, line 16:
```suggestion
        if isinstance(other, Distance):
            total_distance = self.km + other.km
            return Distance(total_distance)
        return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
```
EdAlekseiev, line 24:
```suggestion
        if isinstance(other, Distance):
            self.km += other.km
            return self
        self.km += other
        return self
```

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 45: `other` can be `Distance`, you should change your methods

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

vasylhnatiuk: use without ""

vasylhnatiuk (changes requested):

app/main.py
YukihiroSM: It's still not fixed (probably you didn't pushed changes.

YukihiroSM (changes requested):

PR #930, opened 2024-01-09
Solution

isikach:


MaxymChyncha (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __iadd__(self,
+                 other: Distance | int | float
+                 ) -> Distance:
+        if type(other) in (int, float):
+            self.km += other
+        else:
```
NEkropo1, line 30: You surely don't need this `else`.

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int) -> bool:
+        return self.km >= other
```
NEkropo1, line 52: what if other will be Distance, not int?)

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self,
+                other: Distance | int | float
+                ) -> Distance:
+        if type(other) in (int, float):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
```
NEkropo1, line 20: consider moving your check to some other static method.

NEkropo1 (changes requested):

app/main.py
isikach: Totaly agree. Thanks for advice)

app/main.py
isikach: Agree...Fixed)

app/main.py
isikach: Change to isinstance) Thanks

PR #931, opened 2024-01-10
Solution

MaxymChyncha:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        return self.km < other.km
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km > other
+
+        return self.km > other.km
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (int, float)):
```
NEkropo1, line 58: you could simplify these methods with not + previously implemented operators. Or more advanced if you use [functools.total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering)

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(
+                km=self.km + other
+            )
+
+        return Distance(
+            km=self.km + other.km
+        )
```
NEkropo1, line 23: You're making this check so much times, consider some static check method, or other ideas.

NEkropo1 (changes requested):

NEkropo1 (approved):
Great!

PR #939, opened 2024-01-11
'solution'

liudmylaperchevska:
I have theoretic question: is it possible to made list of math signs and then looping throw pattern? to avoid similar lines at line33 - line56? I guess wrapping isnt fit here


Polyakiv-Andrey (approved):
Good job!

PR #940, opened 2024-01-11
Solution

Bulleter06:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            total_distance = self.km + other.km
+            return Distance(km=total_distance)
+        elif isinstance(other, (int, float)):
+            total_distance = self.km + other
+            return Distance(km=total_distance)
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
shaiduchyk, line 22: remove elif's

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
```
shaiduchyk, line 12: add @staticmethod for "isinstance" instead of use it at every func.

shaiduchyk (changes requested):

PR #953, opened 2024-01-11
'Solution'

artemlitvenko:


Dobryk12 (approved):
If we have equal, greater and less, there is no  
necessity to use greater equal and less equal

app/main.py
```diff
@@ -1,32 +1,31 @@
...
 class Distance:
 
     def __init__(self, km: int | float) -> None:
         self.km = km
 
+    @staticmethod
+    def check_distance(other: Any) -> int | float:
+        if isinstance(other, Distance):
+            return other.km
+        return other
```
NEkropo1, line 16: I guess you should rename function on something like fetch_distance_metric, since we are not checking distance and it's not intuitive )

app/main.py
```diff
@@ -35,22 +34,22 @@ def __mul__(self, times: int) -> Distance:
...
-        other_km = other.km if isinstance(other, Distance) else other
+    def __eq__(self, other: Any) -> bool:
+        other_km = self.check_distance(other)
         return self.km == other_km
 
-    def __le__(self, other: Distance) -> int | float:
-        other_km = other.km if isinstance(other, Distance) else other
+    def __le__(self, other: Any) -> bool:
+        other_km = self.check_distance(other)
         return self.km <= other_km
```
NEkropo1, line 51: You ignored comment about total_ordering or returning `not self.function()`, fix that

NEkropo1 (changes requested):

PR #956, opened 2024-01-13
Solution

AntonSkyrda:


app/main.py
```diff
@@ -1,3 +1,43 @@
...
+        return self.km < other
+
+    def __le__(self, other: object) -> bool:
+        return self.km <= other
+
+    def __gt__(self, other: object) -> bool:
+        return self.km > other
+
+    def __ge__(self, other: object) -> bool:
+        return self.km >= other
```
MaxymChyncha, line 43: You can decrease count of these methods using [total ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering)

MaxymChyncha (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,39 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, number: int) -> int | float | Distance:
+        return Distance(self.km * number)
+
+    def __truediv__(self, other: int) -> int | float | Distance:
+        return Distance(round((self.km / other), 2))
```
NEkropo1, line 33: You make annotation of returning int | float | Distance, even though in return you only return Distance. How could it be?
think more about args annotations and return annotations. I guess here is the issue.

NEkropo1 (changes requested):

NEkropo1 (approved):
LGTM!

PR #967, opened 2024-01-20
Solution

volry:


NazarSheremet (approved):
Looks good, but In task wasn't requirement to predict and raise error

PR #969, opened 2024-01-20
Solution

VasylRybalka:


app/main.py
```diff
@@ -1,3 +1,52 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
NikOneZ1, line 2: Could be also int

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | float") -> "Distance":
```
NikOneZ1, line 11: Use without "" and also int is possible, fix in all places
```suggestion
    def __add__(self, other: Distance | float | int) -> Distance:
```

NikOneZ1 (changes requested):

PR #977, opened 2024-01-23
solution with 4 mistakes

ggeorgiyy:


EdAlekseiev (changes requested):
Fix tests

ggeorgiyy:
I try to merge my pull request.

app/main.py
```diff
@@ -1,3 +1,65 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: int) -> None:
```
YukihiroSM, line 6: km can be float as well.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
+        if isinstance(other, type(self)):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
YukihiroSM, line 19: `other` can be int / float as well. 
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
        if isinstance(other, type(self)):
            return Distance(self.km + other.km)
        elif isinstance(other, (int, float)):
            return Distance(self.km + other)
```

YukihiroSM: Fix the typehinting in all methods

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            self.km = self.km + other
+            return self
+
+    def __mul__(self, other: Distance) -> int:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance) -> int:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
YukihiroSM, line 35: In these two methods `other` can be only int / float, not Distance
```suggestion
    def __mul__(self, other: int | float) -> int:
        if isinstance(other, (int, float)):
            return Distance(self.km * other)

    def __truediv__(self, other: int | float) -> int:
        if isinstance(other, (int, float)):
            return Distance(round(self.km / other, 2))
```

YukihiroSM (changes requested):

YukihiroSM:
> I try to merge my pull request.

You don't need to merge a PR

app/main.py
vasylhnatiuk: do it

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

EdAlekseiev (approved):
Good job

PR #980, opened 2024-01-24
'Solution'

MarynaBar:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+
+    def __eq__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
```
vsmutok, line 49: DRY 

create a separate method for this to avoid code duplication

app/main.py
```diff
@@ -1,3 +1,61 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
vsmutok, line 2:
```suggestion
    def __init__(self, km: float | int) -> None:
```

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
vsmutok, line 17:
```suggestion
    def __iadd__(self, other: int | float | Distance) -> Distance:
```

vsmutok: fix everywhere

vsmutok (changes requested):

PR #982, opened 2024-01-25
Solution

voidmain69:


app/main.py
```diff
@@ -1,3 +1,84 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
EdAlekseiev, line 14: `other` can be Distance too. Fix everywhere
```suggestion
    def __add__(self, other: int | float) -> Distance:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #983, opened 2024-01-26
implement py-distance-class

oleksiimuhyr:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
YukihiroSM, line 14: other can be int/ fliat as well. Please, fix typehinting everywhere. 
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
YukihiroSM, line 33: in these two methods other can be only int or float, fix the typehint.

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        return self
+
+    def __ge__(self, other: Distance) -> Distance | bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        return self
```
YukihiroSM, line 68: These methods should return only boolean values. 
Please, fix returns and typehinting.

YukihiroSM (changes requested):

NikOneZ1 (approved):

PR #984, opened 2024-01-27
solution

prev1ew:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __eq__(self,
+               other: Distance | float | int
+               ) -> bool:
+        return self.km == (other.km if isinstance(other, Distance) else other)
+
+    def __le__(self,
+               other: Distance | float | int
+               ) -> bool:
+        return self.km <= (other.km if isinstance(other, Distance) else other)
```
vsmutok, line 55: DRY - don't repeat yourself 

 `(other.km if isinstance(other, Distance) else other)` - create a separate method for this

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return self.km > (other.km if isinstance(other, Distance) else other)
+
+    def __eq__(self,
+               other: Distance | float | int
+               ) -> bool:
+        return self.km == (other.km if isinstance(other, Distance) else other)
+
+    def __le__(self,
+               other: Distance | float | int
+               ) -> bool:
```
vsmutok, line 54:
```suggestion
    def __le__(self,other: Distance | float | int) -> bool:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
prev1ew: What is the rule for this? When I should write parameters in different rows and when in the same?

PR #985, opened 2024-01-27
Sollution

nazars03:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
+
```
YukihiroSM, line 31: Other can be only int/float in these two methods
```suggestion
    def __mul__(self, other: int | float) -> Distance:
        return Distance(self.km * other)

    def __truediv__(self, other: int | float) -> Distance:
        return Distance(round(self.km / other, 2))

```

YukihiroSM (changes requested):

app/main.py
nazars03: Done ✅

YukihiroSM (approved):
Good job!

PR #993, opened 2024-01-31
'Solution'

olessh:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            return self.km == other.km
+        return self.km == other
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
```
vsmutok, line 43: I think better to create separate method for this

vsmutok: and use it wherever needed

vsmutok (changes requested):

PR #994, opened 2024-01-31
Solution

TarasFirst:


app/main.py
```diff
@@ -1,3 +1,47 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
```
Dimosphen1, line 3: `km` could be not only `float`, but `int` as well. Change your type annotation here and in the other methods to encounter that.

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ("Distance", float, int)) -> "Distance":
```
Dimosphen1, line 12: Change your type annotation to use an updated syntax
```suggestion
    def __add__(self, other: Distance | float | int) -> Distance:
```
In case you face errors, use `from __future__ import annotations` import.

Dimosphen1 (changes requested):

app/main.py
TarasFirst: Thank you, i did it

Polyakiv-Andrey (approved):
Good job!

PR #1005, opened 2024-02-05
implemented class Distance

RomanHlodann:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
pavlejviki, line 14: other can be float as well

fix the type annotation in all places

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __gt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
pavlejviki, line 52: use previously defined magic method to simplify the code

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
```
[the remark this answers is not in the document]
RomanHlodann: not( self > other or self < other)

should i do like that?

Dimosphen1 (approved):
LGTM, well done!

PR #1006, opened 2024-02-05
solved task

Shantagust:


app/main.py
```diff
@@ -1,3 +1,45 @@
+# Used Python 3.10.11
```
Dimosphen1, line 1: Remove this comment

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+from __future__ import annotations
+
+from _ctypes import Union
```
Dimosphen1, line 4: Do not use `Union`, use `|` in the type annotation instead

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Dimosphen1, line 8: `km` could be `float` as well. Change your type annotation to encounter that.

Dimosphen1 (changes requested):

PR #1009, opened 2024-02-06
Solution

IhorDmytriv:


app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, value: Distance | int) -> Distance:
```
EdAlekseiev, line 14: `value` can be float too. Fix everywhere, please
```suggestion
    def __add__(self, value: Distance | int | float) -> Distance:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #1012, opened 2024-02-07
Implement py-distance-class

anrozhna:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

vasylhnatiuk: use without ""

vasylhnatiuk: add `from __future__ import annotations`

vasylhnatiuk (changes requested):

PR #1013, opened 2024-02-07
add task solution

EugeneSherepa:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, '"Distance"']) -> "Distance":
```
LiudmylaKulinchenko, line 14: from python 3.10 you can avoid using Union in type annotations
You can annotate like this: `int | float | Distance`. It also should work without quotes for a class name

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (float, int)):
+            return self.km <= other
+
+    def __ge__(self, other: Union[int, float, "Distance"]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (float, int)):
+            return self.km >= other
```
LiudmylaKulinchenko, line 64: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >
Also remember to use operands directly: for example `<` instead of calling methods like this `self.__lt__(other)`

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 🌼🌷

PR #1016, opened 2024-02-08
Solution

RomanNest:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, type(self)):
+            self.km += other.km
+            return self
+        if isinstance(other, int | float):
+            self.km += other
+            return self
```
EdAlekseiev, line 27:
```suggestion
        if isinstance(other, type(self)):
            self.km += other.km
        elif isinstance(other, int | float):
            self.km += other
        return self
```

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 48: `other` can be Distance in these methods too

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+            return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float | Distance) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        return self.km < other
```
EdAlekseiev, line 36: This function doesn't work, when other is Distance

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        return self.km > other
```
EdAlekseiev, line 39: The same comment as above

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        return self.km == other
```
EdAlekseiev, line 42: The same comment as above

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return self.km <= other
```
EdAlekseiev, line 45: The same comment as above

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 48: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #1024, opened 2024-02-13
'Solution'

shramkomax:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
```
NazarSheremet, line 29: Mistake in annotation of `other` and checking of type

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+            return Distance(self.km * other)
+        else:
+            raise TypeError("Unsupported operand type for *")
+
+    def __truediv__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
NazarSheremet, line 36: The same, you annotate that `other` is Distance, but check if it is `int, float`

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
NazarSheremet, line 11: Annotation is incorrect 
```suggestion
    def __add__(self, other:  int | float | Distance) -> "Distance":
```

NazarSheremet (changes requested):
Need some changes

NazarSheremet (approved):
Good job!

PR #1026, opened 2024-02-14
'Solution'

donbas777:


app/main.py
```diff
@@ -1,3 +1,48 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Dimosphen1, line 5: `km` could be `int` as well, indicate this in the type annotation

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        self.km += Distance.return_number(other)
+        return self
+
+    def __mul__(self, other: Distance | float | int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float | int) -> Distance:
+        self.km / Distance.return_number(other)
+        print(Distance(self.km / other))
```
Dimosphen1, line 32: There is no need to keep these operations, remove them

Dimosphen1 (changes requested):

PR #1034, opened 2024-02-19
Solution

rosttv:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Dimosphen1, line 24: It would be preferable to add `else` statement here to avoid similar `return` statements with `self`
```suggestion
        if isinstance(other, Distance):
            self.km += other.km
        else:
            self.km += other
        return self
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #1035, opened 2024-02-19
'Solution'

romanlrmk:


app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 44: `other`  can be Distance here too. Fix these methods

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return self.km < other
```
EdAlekseiev, line 32: If `other` can be Distance, you should handler this option inside function. Fix everywhere
```suggestion
        if isinstance(other, Distance):
              return self.km < other.km
        return self.km < other
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #1038, opened 2024-02-21
Solution

VerFox46:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> None:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> None:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> None:
```
NikOneZ1, line 11: Indicate, that other also could be Distance type. Do it in other functions too

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
NikOneZ1, line 15: Else is redundant here
```suggestion
        return Distance(self.km + other)
```

NikOneZ1 (changes requested):

VerFox46:
Corrected

EdAlekseiev (approved):
Good job

PR #1044, opened 2024-02-23
'Solution'

ArtiSH0CKs:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[float, "Distance"]) -> "Distance":
```
vsmutok, line 15:
```suggestion
    def __add__(self, other: Union[float, int, Distance]) -> Distance:
```

vsmutok: fix everywhere

vsmutok (changes requested):

app/main.py
ArtiSH0CKs: > fix everywhere

if i delete quotes i get the next error: "NameError: name 'Distance' is not defined"

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[float, int, "Distance"]) -> "Distance":
+        # if i delete quotes i get the next error:
+        # "NameError: name 'Distance' is not defined"
```
vasylhnatiuk, line 17: use `from __future__ import annotations`. to avoid this error

vasylhnatiuk (changes requested):

PR #1048, opened 2024-02-24
'Solution'

ArtomUmantsev:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
```
NikOneZ1, line 12: Also add Distance type in all places 
```suggestion
    def __add__(self, other: int | Distance) -> "Distance":
```

NikOneZ1 (changes requested):

PR #1056, opened 2024-02-26
Solution

mwellick:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance:
```
EdAlekseiev, line 14: `other` can be `float` and `Distance` too. Fix it everywhere
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
vasylhnatiuk, line 26: other can be float here as well and in all of bellow methods

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #1057, opened 2024-02-27
my solution

kostiukmkalne:
I have a problem with some tests

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,52 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km:.2f} kilometers."
+
+    def __repr__(self):
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __truediv__(self, other):
+        if isinstance(other, (int, float)):
+            result = self.km / other
+            return Distance(round(result, 2))
+        else:
+            raise TypeError("Unsupported operand type for /")
+
+    def __lt__(self, other):
+        return self.km < other
```
NazarSheremet, line 40: Need to add a check if other is an instance of Distance

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __gt__(self, other):
+        return self.km > other
```
NazarSheremet, line 43: The same. Need to add a check if other is an instance of Distance

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+
+    def __eq__(self, other):
+        return self.km == other
+
+    def __le__(self, other):
+        return self.km <= other
+
+    def __ge__(self, other):
+        return self.km >= other
```
NazarSheremet, line 52: The same for all methods

NazarSheremet (changes requested):
Also need to add annotation for all variables
image

app/main.py
kostiukmkalne: Thank You!
I fixed everything

NazarSheremet (approved):
Good job!

PR #1058, opened 2024-02-28
Solution

IjdmLoverI:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
```suggestion
    def __add__(self, other: Distance) -> Distance:
```

vasylhnatiuk: use `from __future__ import annotations`

vasylhnatiuk (changes requested):

PR #1059, opened 2024-02-28
'Solution'

YESosnovska:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
```
NikOneZ1, line 11: Try to not use "object" as type for arguments, because it is not a good practice. Fix it everywhere
```suggestion
    def __add__(self, other: int | float| Distance) -> "Distance":
```

NikOneZ1 (changes requested):

vasylhnatiuk (approved):

NikOneZ1 (approved):

PR #1060, opened 2024-02-28
Develop

Yaschenko1:


app/main.py
```diff
@@ -1,3 +1,59 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Dimosphen1, line 5: `km` could be `float` as well, indicate this in the type annotation

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        elif isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        elif isinstance(other, Distance):
+            self.km += other.km
+            return self
```
Dimosphen1, line 26: There is no need to return `self` twice, write `return` once in the end after all conditions to avoid duplication

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __mul__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float | int) -> Distance:
```
Dimosphen1, line 32: Add processing of the `Distance` class instance here as well in case it is indicated in the type annotation

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, (int, float)):
+            if other != 0:
```
Dimosphen1, line 34: Join these two checks into one to avoid unnecessary nestng

Dimosphen1 (changes requested):

PR #1061, opened 2024-02-29
implement Distance class

Lumideus:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | float) -> Distance:
+        return Distance(round(self.km / other, 2))
+
```
YukihiroSM, line 32: Here other can't be Distance. Fix typehint.

YukihiroSM (changes requested):
In all methods other can be int also. Please, fix it

PR #1066, opened 2024-03-04
Solution

CreepyMo:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
Dimosphen1, line 14: Change your type annotation to avoid using Union and placing class in quotes
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```
In case you face errors, use `from __future__ import annotations` import

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        return (
+            Distance(self.km + other.km)
+            if isinstance(other, Distance)
+            else Distance(self.km + other)
+        )
```
Dimosphen1, line 19: It would be better to place code in multiple lines out of `()` so that if condition would be more readable. Change the code in all related places to encounter that

Dimosphen1 (changes requested):

app/main.py
CreepyMo: Done

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        return Distance(self.km + other.km) \
```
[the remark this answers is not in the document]
CreepyMo: Done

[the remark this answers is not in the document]
CreepyMo: My bad. Done

vasylhnatiuk (approved):

PR #1067, opened 2024-03-04
Solution

Boyok1337:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
LiudmylaKulinchenko, line 14: The other also could be Distance, add it to type annotations

LiudmylaKulinchenko: same everywhere

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
LiudmylaKulinchenko, line 55: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >
P.S. use operand `<` instead of `self.__lt__(other)` etc

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        return not Distance.__gt__(self, other)
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        return not Distance.__lt__(self, other)
```
LiudmylaKulinchenko, line 51: u are close)
but it can be written simpler

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
LGTM!

PR #1070, opened 2024-03-09
Solution

Bujhm99:


app/main.py
```diff
@@ -1,3 +1,41 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, second: object | int) -> object:
```
LiudmylaKulinchenko, line 11: do not use object as a type annotation. Every type is object in python
Make clear type annotations here: Disctance, integer, float

LiudmylaKulinchenko, line 11: `second` is not a descriptive name. Fix everywhere

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+        self.km *= mul
+        return self
+
+    def __truediv__(self, devv: int) -> object:
+        if devv == 0:
+            return float("inf")
+        self.km = round(self.km / devv, 2)
+        return self
+
+    def __lt__(self, sec: object | int) -> bool:
```
LiudmylaKulinchenko, line 28: Use whole words as a parameter names to make it more readable, fix everywhere

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+        return self.km > getattr(sec, "km", sec)
+
+    def __eq__(self, sec: object | int) -> bool:
+        return self.km == getattr(sec, "km", sec)
+
+    def __le__(self, sec: object | int) -> bool:
+        return self.km <= getattr(sec, "km", sec)
+
+    def __ge__(self, sec: object | int) -> bool:
+        return self.km >= getattr(sec, "km", sec)
```
LiudmylaKulinchenko, line 41: U can use previously created methods here
Hint: <= is the same as not >

LiudmylaKulinchenko (changes requested):
getattr is a good choice here 👍

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, addentum: object | int) -> Distance:
```
YuriiKindrat, line 14: object is still in type annotations, user more descriptive type annotations, please
object type means that method could accept any object in python (should accept only Distance/int objects)
fix this everywehre

YuriiKindrat (changes requested):

PR #1072, opened 2024-03-11
'Solution'

Bludger404:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return "Distance: {} kilometers.".format(self.km)
+
+    def __repr__(self) -> str:
+        return "Distance(km={})".format(self.km)
+
+    def __add__(self, other: "Distance") -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vasylhnatiuk: add `from __future__ import annotations` in the begging of file

vasylhnatiuk: fix annotation everywhere

vasylhnatiuk (changes requested):

PR #1073, opened 2024-03-11
Solution

Luxiv:


NazarSheremet (changes requested):
Incorrect annotation, you are checking if `other` is int and float, however, this is not specified in the annotation.

YuriiSmolii (approved):
LGTM

PR #1076, opened 2024-03-12
First_try

Sydorovich:


NazarSheremet (changes requested):
Good job, only need to update the annotation of the function, where you expect float data type

Sydorovich:
Updated and added float type in annotations.

Sydorovich:
Asking for new check and approve.

PR #1079, opened 2024-03-13
Solution

shafran-andriy:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self: "Distance", km: int) -> None:
+        self.km = km
+
+    def __str__(self: "Distance") -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self: "Distance") -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self: "Distance", other: "Distance") -> "Distance":
```
EdAlekseiev, line 11: other can be int and float. Fix everywhere
```suggestion
    def __add__(self: "Distance", other: "Distance" | int | float) -> "Distance":
```

EdAlekseiev (changes requested):

app/main.py
shafran-andriy: Fixed

app/main.py
```diff
@@ -1,3 +1,85 @@
...
+    def __init__(self: "Distance", km: int) -> None:
+        self.km = km
+
+    def __str__(self: "Distance") -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self: "Distance") -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self: "Distance",
```
```suggestion
    def __add__(self: Distance,
```

vasylhnatiuk (changes requested):

app/main.py
shafran-andriy: fixed

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #1082, opened 2024-03-14
'Solution'

PozniakVA:


app/main.py
```diff
@@ -1,3 +1,46 @@
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    @staticmethod
+    def isinstance_(other: int | float) -> int | float:
```
Dimosphen1, line 7: `other` could be `Distance` class as well. Indicate it in the type annotation and place this method after `__init__`. Use another method name to indicate that you return a `km`, not just performing an `isinstance` check

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Any:
```
Dimosphen1, line 19: Change your type annotation here and in other places to indicate the Distance class instance
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```
Use `from __future__ import annotations` import in case of errors

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __iadd__(self, other: int | float) -> Any:
+        self.km += Distance.isinstance_(other)
+        return self
+
+    def __mul__(self, other: int | float) -> Any:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Any:
+        if other != 0:
```
Dimosphen1, line 30: Rewrite this condition to take the boolean value into an account
```suggestion
        if other:
```

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #1083, opened 2024-03-15
Solution Class Distance March 15

DmytroPieskov:


app/main.py
```diff
@@ -1,3 +1,65 @@
+from __future__ import division
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
```
Dimosphen1, line 6: Indicate in the type annotation that `km` could be an integer as well.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, value: any) -> any:
```
Dimosphen1, line 15: Be specific with the type annotation here and in the other methods as well

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, value: Union[int, object]) -> object:
```
```suggestion
    def __add__(self, value: Union[int, object, Distance]) -> Distance:
```

vasylhnatiuk: change evrywhere

vasylhnatiuk (changes requested):

app/main.py
pavlopro: Please consider comment above and make annotation as int | float

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1084, opened 2024-03-15
Develop

saywin:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[int, float, Distance]) -> Distance:
```

vasylhnatiuk: use from __future__ import annotations

vasylhnatiuk (changes requested):

app/main.py
Saywin: If I remove the quotes, it gives the error "Unresolved reference 'Distance'"

vasylhnatiuk (changes requested):
you should write aslo these methods:` __lt__, __le__, __ge__`

vsmutok (approved):
Great!

PR #1085, opened 2024-03-15
Solution

DmytroHlazyrin:


app/main.py
```diff
@@ -1,3 +1,62 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
nicksetrakov, line 5: "km" can be int also, so in others annotayions change it too

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
nicksetrakov, line 17: You don't need to use "else", just return. Fix it in similar situations below

nicksetrakov (changes requested):

app/main.py
DmytroHlazyrin: You are not quite right! When we implement the magic method __truediv__ - it returns an instance of the class that takes as an argument the result of division of two numbers, which is more often of type float.

Oleksl888: > "km" can be int also, so in others annotayions change it too


Please fix the above request

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return self.km < other
+
+    def __gt__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
```
Oleksl888, line 44: When comparing two floats for equality you should also consider float imprecision

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __gt__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return round(self.km, 2) == round(other.km, 2)
+        return round(self.km, 2) == round(other, 2)
```
Oleksl888, line 45: round() may pass for a solution in our case but keep in mind it is not foolproof.
It is better to use more reliable options how to handle imprecision: like not using floats at all (converting to ints) or using tolerances when comparing float values may be a better option. 
Please read more below:

Oleksl888 (approved):

PR #1089, opened 2024-03-16
Solution

fishray33:


app/main.py
```diff
@@ -1,3 +1,71 @@
...
+
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, int | float):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, int | float):
+            return self.km >= other
+        return self.km >= other.km
```
NazarSheremet, line 71: If you expect that `other `can be the `Distance `class instance, add it to annotation

NazarSheremet (changes requested):
Minor changes of annotation

NazarSheremet (approved):
Good job!

PR #1090, opened 2024-03-16
Solution

panicua:


app/main.py
```diff
@@ -1,3 +1,122 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(
+        self, other: Union["Distance", int, float]
+    ) -> Union["Distance", TypeError]:
```
panicua, line 16: I don't know how it works with newer version of Python but on my 3.11 it didn't let me write ```other: Distance``` only ```other: "Distance"``` and after that if I wanted to use ```other: "Distance" | int | float``` tests were giving me an error, that's why I had to use ```Union```.

DmytroHlazyrin: You can use import annotation from __future__ in this case

app/main.py
```diff
@@ -1,3 +1,122 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            other = Distance(other)
+            return Distance(self.km + other.km)
+        raise TypeError(
+            f"unsupported type for __add__ method. 'Distance', "
+            f"<class 'float'> or <class 'int'> were expected. "
+            f"But received {type(other)}"
+        )
```
DmytroHlazyrin, line 26: if you had used type annotation from future you wouldn't have had to describe the type error in all this cases, as it would be pretty much the same by default.

DmytroHlazyrin (changes requested):
use import annotation from --future-- and refactor your code

panicua:
@DmytroHlazyrin I changed from ```typing > Unite``` to ```__future__ > annotations``` but I left my raise errors cause you can see that future for example for ```__add__``` magic method still doesn't indicate exactly at the problem. (screenshots below, 1st without my error raise, 2nd with my own error raise)
NVIDIA_Share_xSIw02vWNQ
NVIDIA_Share_5w0bZA9OHo

app/main.py
```diff
@@ -1,3 +1,119 @@
...
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        if isinstance(other, (int, float)):
+            other = Distance(other)
+            return self.km >= other.km
+        raise TypeError(
+            f"unsupported type for __ge__ method. <class 'float'> "
+            f"or <class 'int'> were expected. "
+            f"But received {type(other)}"
+        )
```
YuriiKindrat, line 119: u can use previous methods here, and implement them in one line )

YuriiKindrat (changes requested):

app/main.py
panicua: you mean that I should user the ```__lt__``` and ```__gt__``` under ```__le__``` and ```__ge__```? cause I didn't rly understand

app/main.py
```diff
@@ -94,26 +94,8 @@ def __eq__(self, other: int | float) -> bool | TypeError:
...
-        if isinstance(other, (int, float)):
-            other = Distance(other)
-            return self.km <= other.km
-        raise TypeError(
-            f"unsupported type for __le__ method. <class 'float'> "
-            f"or <class 'int'> were expected. "
-            f"But received {type(other)}"
-        )
+    def __le__(self, other: int | float | Distance) -> bool | TypeError:
+        return self.__eq__(other) or self.__lt__(other)
```
panicua, line 98: but in this case won't be able to raise a custom error for ```__le__``` and ```__ge__```, it will raise an error from ```__eq__```,```__lt__ or ```__gt__```

Oleksl888 (changes requested):
This comment by Yuri haven't been resolved yet. Can you please have a look?

> it would be better to write a custom method for error raising to avoid code duplication in every method also, u could write an decorator with try-except inside it, to optimize ur methods workflow and write less code and conditions

panicua:
> This comment by Yuri haven't been resolved yet. Can you please have a look?
> 
> > it would be better to write a custom method for error raising to avoid code duplication in every method also, u could write an decorator with try-except inside it, to optimize ur methods workflow and write less code and conditions

we didn't learn Try/Except methods yet, but I can create a static method for comparison magic methods.

app/main.py
```diff
@@ -1,3 +1,98 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
```
Oleksl888, line 6: What if user passes wrong data type to our class? Should we have some validation in place?

app/main.py
```diff
@@ -1,3 +1,98 @@
...
+        self, other: int | float | Distance
+    ) -> bool | callable(TypeError):
+        return not self > other
+
+    def __ge__(
+        self, other: int | float | Distance
+    ) -> bool | callable(TypeError):
+        return not self < other
+
+    @staticmethod
```
Oleksl888, line 78: 1. Since your function doesn't do anything except raising an error with a message you can just raise TypeError in your code with a appropriate message.
2. If you want an error handler with certain logic it is define it in a separate class - no need to mix it with other logic. 
3. Also you have 3 methods that do exactly the same - so you could actually have only 1 with different parameters.
4. I understand that exceptions are beyond current scope of what you have learned - so you can remove this block of code and make it more simple.
5. Let me know if you have any questions

app/main.py
```diff
@@ -1,3 +1,98 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance | TypeError:
+        if isinstance(other, Distance):
```
Oleksl888, line 15: Simplify your methods to use less if conditions

Oleksl888 (changes requested):

panicua:
-Removed ```@staticmethod```
-Added ```@property``` decorator to check type of an input value.
-Removed all ```raise``` with ```TypeError``` to the better time when we learn about ```Try/Except```

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    @property
+    def km(self) -> int | float:
+        return self._km
+
+    @km.setter
+    def km(self, value: int | float) -> None:
+        if not isinstance(value, (int | float)):
```
Oleksl888, line 14: well done with this! just need a small fix to the second argument to isinstance method. (how do we separate items in tuple?)

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        self._km = value
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (Distance, int, float)):
```
Oleksl888, line 29: you basically check for int and float in isinstance method twice. You can simplify this

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+            if isinstance(other, (int, float)):
+                other = Distance(other)
+            return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (Distance, int, float)):
```
Oleksl888, line 35: same as above - no need to check the same types twice + no need for nested if's. Also refactor this in other methods.

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+                other = Distance(other)
+            return self.km > other.km
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, (Distance, int, float)):
+            if isinstance(other, (int, float)):
+                other = Distance(other)
+            return self.km == other.km
+
+    def __le__(self, other: int | float | Distance) -> bool:
```
Oleksl888, line 69: Here you compare self (instance of Distance class) with possible int or float. Also why do you use 'not' operator here?

Oleksl888 (changes requested):

app/main.py
panicua: > Here you compare self (instance of Distance class) with possible int or float. Also why do you use 'not' operator here?
---Here you compare self (instance of Distance class) with possible int or float.
Actually I believe that's not true, I use ```return not self > other``` so it goes to ```__gt__``` method at this line and ```__gt__``` method is checking if other is ```int```, ```float``` or ```Distance```. If it's ```int``` or ```float``` it does ```other = Distance(other)``` under ```__gt__``` method as you can see. 
p.s. screenshot is attached with comparison of ```Distance``` with ```int/float/Distance```
NVIDIA_Share_wStengcfP4

---Also why do you use 'not' operator here?
Because when I use ```return not self > other``` in ```__le__``` ```less-equal``` magic method it returns ```True``` when ```self == other``` or ```self < other```, why should I use 2 comparisons when I can just write the key word ```not``` and just use ```self > other```? (Btw this was the reccomendation from @YuriiKindrat on our QnA I had: ```self == other``` or ```self < other``` before that.

app/main.py
panicua: True, totally missed it, already fixed, will push with the next commit.

app/main.py
panicua: I don't get how I can do it with 1 ```return``` and 1 ```if``` cause if I get other as ```Distance``` it's gonna work fine, if I get other as ```int``` or ```float```, I'll have to make ```other = Distance(other)``` so I could compare their ```.km``` after.

app/main.py
panicua: I don't get how I can do it with 1 return and 1 if cause if I get other as Distance it's gonna work fine, if I get other as int or float, I'll have to make other = Distance(other) so I could compare their .km after.

app/main.py
Oleksl888: `    if isinstance(other, (int, float)):
        other = Distance(other)
    return Distance(self.km + other.km)` try following this pattern in your methods to simplify code

app/main.py
Oleksl888: You can do it, I believe in you :)

app/main.py
Oleksl888: That's a good point and clever implementation. I totally overlooked that.
Also along with that you can use the following:
[https://docs.python.org/3/library/functools.html#functools.total_ordering](url)

Oleksl888 (changes requested):

app/main.py
panicua: > ` if isinstance(other, (int, float)): other = Distance(other) return Distance(self.km + other.km)` try following this pattern in your methods to simplify code

oh, right I didn't see it at first ;d

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+
+    @km.setter
+    def km(self, value: int | float) -> None:
+        if not isinstance(value, (int, float)):
+            raise ValueError(
+                f"Unsupported type. <class 'float'> "
+                f"or <class 'int'> were expected. "
+                f"Got {type(value)} instead."
+            )
+        self._km = value
```
YuriiKindrat, line 20: GJ!

YuriiKindrat (approved):
LGTM!

PR #1092, opened 2024-03-18
Solution

CharmNiovi:


Gilgumish (approved):
GJ!

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float | int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: float | int) -> bool:
+        return self.km < getattr(other, "km", other)
+
+    def __gt__(self, other: float | int) -> bool:
+        return self.km > getattr(other, "km", other)
```
Oleksl888, line 33: This is clever  👏 . But it also means that your code will accept any object that has "km" attribute while we want to compare only  between numeric data types and 'Distance' objects. Also, type annotations are missing.

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+
+    def __eq__(self, other: float | int) -> bool:
+        return self.km == getattr(other, "km", other)
```
Oleksl888, line 36: Make sure you handle possible imprecision issues when comparing float values for equality. 
Please read about it below:

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+from math import isclose
+
+
 class Distance:
-    # Write your code here
-    pass
+    __slots__ = "km"
+
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
```
Oleksl888, line 10: What if user passes wrong data type to our class? Should we have some validation in place?

app/main.py
Oleksl888: I see that you added the types 👍. 
However right now your code will perform comparison with any data type that will have a 'km' attribute and this may lead to unexpected beahviour. E.g
```
class NotADistance:
    def __init__(self, km: int) -> None:
        self.km = 1
``` 
will be considered a valid class for comparison operation while it should not be.

Oleksl888 (changes requested):
Please check again

app/main.py
Oleksl888: I see you haven't started topic with properties and descriptors yet. So if you can't do it right now you can ignore this comment. But if you can do it you will be amazing :)

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, (Distance, int, float)):
+            return isclose(self.km, getattr(other, "km", other))
+
+    def __le__(self, other: Union[Distance, int, float]) -> bool | None:
+        if isinstance(other, (Distance, int, float)):
+            return self.km <= getattr(other, "km", other)
+
+    def __ge__(self, other: Union[Distance, int, float]) -> bool | None:
+        if isinstance(other, (Distance, int, float)):
+            return self.km >= getattr(other, "km", other)
```
YuriiKindrat, line 65: u can use previouse created methods here, so there won't be need to make conditions

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            self.km += getattr(other, "km", other)
+            return self
+
+    def __mul__(self, other: Union[float, int]) -> Distance | None:
+        if isinstance(other, (Distance, int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[float, int]) -> Distance | None:
+        if isinstance(other, (Distance, int, float)):
+            return Distance(round(self.km / other, 2))
```
YuriiKindrat, line 45: but other could not be distance instance here

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return self._km
+
+    @km.setter
+    def km(self, value: int | float) -> None:
+        if not isinstance(value, (int, float)):
+            raise ValueError(
+                f"""Unsupported type. Expected float or int.
+                Got {type(value)} instead."""
+            )
+        self._km = value
```
YuriiKindrat, line 22: GJ!

YuriiKindrat (changes requested):

app/main.py
Oleksl888: Well done sir!

Oleksl888 (approved):
We got production-level solution here :) Good job!

PR #1093, opened 2024-03-18
Solution

vitalinamal:


Oleksl888 (changes requested):
request changes

app/main.py
```diff
@@ -1,3 +1,55 @@
+from typing import Union
+import math
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
```

Oleksl888 (changes requested):

app/main.py
Oleksl888: I see you haven't started topic with properties and descriptors yet. So if you can't do it right now you can ignore this comment. But if you can do it you will be amazing :)

Oleksl888 (approved):
Well done!
You got a bit inconsistent with numeric data types: you have int/float in one method and just int in another one. Here we assume that if you can multiply by int you can as well multiply by a float. But I will give it a pass.

PR #1095, opened 2024-03-18
Solution

Gilgumish:


app/main.py
```diff
@@ -1,3 +1,41 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> object:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> object:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
Oleksl888, line 28: The comparison methods below do not allow comparison between another Distance object

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+        return self.km < other
+
+    def __gt__(self, other: int | float) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
```
Oleksl888, line 34: Make sure you handle possible imprecision issues when comparing float values for equality.
You can read about it below:

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __add__(self, other: float | int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: float | int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
```
Oleksl888, line 24: This can be simplified so you use less comparison operators

app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+import math as m
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
```

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float | int | Distance) -> Distance:
+        return Distance(self.km + (other.km
+                                   if isinstance(other, Distance) else other))
```
YuriiKindrat, line 21: it is hard to read this, try to break the line with Distance brackets

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return self.km < (other.km if isinstance(other, Distance) else other)
+
+    def __gt__(self, other: float | int | Distance) -> bool:
+        return self.km > (other.km if isinstance(other, Distance) else other)
+
+    def __eq__(self, other: float | int | Distance) -> bool:
+        return math.isclose(self.km,
+                            (other.km if isinstance(other, Distance)
+                             else other),
+                            rel_tol=1e-9)
```
YuriiKindrat, line 43: same is here

YuriiKindrat (changes requested):

PR #1096, opened 2024-03-18
Solution#1

VitalyBashkiser:


mikhailyemets (approved):
gj

app/main.py
```diff
@@ -1,3 +1,71 @@
+import operator
+from typing import Union, Callable
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError("Unsupported type for addition.")
```
Oleksl888, line 21: simplify this and below methods to include less if-elif-else statements

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
Oleksl888, line 15: Is the Distance the only type for other?
Please check all below method and make sure type hints everywhere are set according to the task requirements

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    ) -> bool:
+        if isinstance(other, Distance):
+            return comparison_operator(self.km, other.km)
+        elif isinstance(other, (int, float)):
+            return comparison_operator(self.km, other)
+        else:
+            raise TypeError("Unsupported type for comparison.")
+
+    def __lt__(self, other: Union["Distance", float]) -> bool:
+        return self._compare_with_operand(other, operator.lt)
```
Oleksl888, line 59: This is clever 👏

Oleksl888 (changes requested):

app/main.py
Oleksl888: I see you haven't started topic with properties and descriptors yet. So if you can't do it right now you can ignore this comment. But if you can do it you will be amazing :)

Oleksl888: Nice! Now on object creation we will reject incorrect values and our program will be more stable! Later we will learn how to handle it in more pythonic way

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            raise TypeError("km must be a number")
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", float]) -> "Distance":
```
Oleksl888, line 17: You could be more consistent with typing - ints can be used for addition as well.

Oleksl888 (approved):
Well done!

PR #1097, opened 2024-03-18
Solution

SashaKisliy:


oleksandra-shershen (approved):
GJ!

tests/test_main.py
```diff
@@ -6,16 +6,7 @@
...
-    'kilometers',
-    [
-        50,
-        100,
-        300,
-        12.5,
-        0.6
-    ]
-)
+@pytest.mark.parametrize("kilometers", [50, 100, 300, 12.5, 0.6])
```
oleksandra-shershen, line 9: Why you changed a file with tests?

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        elif isinstance(other, Distance):
+            raise TypeError("Unsupported operation.")
+
+    def __truediv__(self, other: int | float) -> Distance | TypeError:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        elif isinstance(other, Distance):
+            raise TypeError("Unsupported operation.")
+
+    def __lt__(self, other: int | float) -> bool:
```
oleksandra-shershen, line 46: For all comparison magic methods add error handling. For example, 
```
    def __lt__(self, other: "Distance" | int | float)) -> bool:
            return self.km < other.km
            return self.km < other
        else:
            raise TypeError("Unsupported operand type for <")
```

oleksandra-shershen (changes requested):
Your code looks well-structured and follows good practices in terms of type annotations and handling unsupported operations

tests/test_main.py
SashaKisliy: I think it's black

app/main.py
SashaKisliy: Okay, I'll add it now

oleksandra-shershen (approved):
LGTM!

app/main.py
```diff
@@ -1,3 +1,84 @@
...
+        else:
+            raise TypeError("Unsupported operand type for >=")
+
+    def __le__(self, other: int | float) -> bool | TypeError:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        else:
+            raise TypeError("Unsupported operand type for <=")
```
YuriiKindrat, line 84: could u use previous created methods here ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+
+    def __ge__(self, other: int | float) -> bool | TypeError:
+        return self._compare_with_operand(other, operator.ge)
+
+    def __le__(self, other: int | float) -> bool | TypeError:
+        return self._compare_with_operand(other, operator.le)
+
+    def _compare_with_operand(
+            self,
+            other: Distance | float,
```
Oleksl888, line 65: Missing int type here

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        elif isinstance(other, Distance):
+            raise TypeError("Unsupported operation.")
+
+    def __truediv__(self, other: int | float) -> Distance | TypeError:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        elif isinstance(other, Distance):
+            raise TypeError("Unsupported operation.")
+
+    def __lt__(self, other: int | float) -> bool | TypeError:
```
Oleksl888, line 48: Did you mention all accepted types here and below?

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __mul__(self, other: int | float) -> Distance | TypeError:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        elif isinstance(other, Distance):
+            raise TypeError("Unsupported operation.")
+
+    def __truediv__(self, other: int | float) -> Distance | TypeError:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        elif isinstance(other, Distance):
```
Oleksl888, line 45: Since we support division only for ints and floats it means that all other types may result in TypeError thus you don't need this condition here.
Check your other methods and fix accordingly where necessary

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError(
+                f"Unsupported operand type(s) for +=: /n"
+                f" 'Distance' and '{type(other)}'"
+            )
+        return self
```
Oleksl888, line 34: you can simplify this - using less conditional statements

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1098, opened 2024-03-18
Solution for distance class

mikhailyemets:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
SashaKisliy, line 11: You need to implement error handling logic and also expand the type annotation. For example:
`    def __add__(self, other: int | float | Distance) -> Distance | TypeError:
        else:
            raise TypeError("Unsupported operand for addition")`

SashaKisliy (changes requested):
Good job, but please add error handling logic and expand the type annotations.

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError("Unsupported operand for:+ operation")
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
SashaKisliy, line 19: GJ) Just add in the annotations that an error may return. -> Distance | TypeError

SashaKisliy (approved):
Good job, keep it up)

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+            return Distance(self.km * other)
+        else:
+            raise TypeError("Unsupported operand for:* operation")
+
+    def __truediv__(self, other: (int, float)) -> "Distance":
+        if isinstance(other, (int, float)):
+            result = self.km / other
+            return Distance(round(result, 2))
+        else:
+            raise TypeError("Unsupported operand for:/ operation")
```
YuriiKindrat, line 39: do u need to use 'else' here ?

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+        else:
+            raise TypeError("Unsupported operand for bool operation")
+
+    def __ge__(self, other: ("Distance", float, int)) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (float, int)):
+            return self.km >= other
+        else:
+            raise TypeError("Unsupported operand for bool operation")
```
YuriiKindrat, line 79: u can use previous implemented methods here

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+    def __eq__(self, other: ("Distance", float, int)) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (float, int)):
+            return self.km == other
+        else:
+            raise TypeError("Unsupported operand for bool operation")
```
YuriiKindrat, line 63: smae about else here

YuriiKindrat (changes requested):

app/main.py
mikhailyemets: @YuriiKindrat 
I've done all my code without else TypeError.
but another student didn't want to accept this task without it

YuriiKindrat: @mikhailyemets U could discuss this with them, so they will approve if u are right )

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
YuriiKindrat, line 11: no need to use quotes for distance in type annotations

YuriiKindrat (approved):

PR #1099, opened 2024-03-18
Solution for task distance-class

oleksandra-shershen:


oleksandra-shershen:
I attempted to add a decorator to the methods in the Distance class. However, due to certain constraints and compatibility issues, I couldn't apply the decorator to all methods of the class.

I successfully added the validate_operand decorator to the following methods:
```
__add__
__lt__
__gt__
__eq__
__le__
__ge__
```
Unfortunately, I encountered difficulties in applying the decorator to the following methods:
```
__iadd__
__mul__
__truediv__
```
The challenges stemmed from the specific logic and type handling required by these methods, especially when dealing with the Distance class instances directly.

SerhiiMar (approved):
Well done!

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+def validate_operand(func: Callable) -> Callable:
+    def wrapper(self: "Distance", other: ("Distance", int, float)) -> Any:
+        if isinstance(other, Distance):
+            return func(self, other.km)
+        elif isinstance(other, (int, float)):
+            return func(self, other)
+        else:
+            raise TypeError(f"Unsupported operand type for {func.__name__}")
+
+    return wrapper
```
YuriiKindrat, line 13: GJ)

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +=")
+        return self
```
YuriiKindrat, line 37: why don't u use decorator here ?)

app/main.py
YuriiKindrat: but add a functool.wraps for inner function as a best practice

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+    def __eq__(self, other: ("Distance", int, float)) -> bool:
+        return self.km == other
+
+    @validate_operand
+    def __le__(self, other: ("Distance", int, float)) -> bool:
+        return self.km <= other
+
+    @validate_operand
+    def __ge__(self, other: ("Distance", int, float)) -> bool:
+        return self.km >= other
```
YuriiKindrat, line 69: no need to use validation here, u can use previouse created methods here
`hint: >= is not <`

YuriiKindrat (changes requested):

app/main.py
oleksandra-shershen: Done

app/main.py
oleksandra-shershen: Done

app/main.py
oleksandra-shershen: Done

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return self.km > other
+
+    def __eq__(self, other: ("Distance", int, float)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: ("Distance", int, float)) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __ge__(self, other: ("Distance", int, float)) -> bool:
+        return not self.__lt__(other)
```
LiudmylaKulinchenko, line 62: you are close)
but it can be written simpler

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return self.km < other
+
+    def __gt__(self, other: ("Distance", int, float)) -> bool:
+        return self.km > other
+
+    def __eq__(self, other: ("Distance", int, float)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: ("Distance", int, float)) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
LiudmylaKulinchenko, line 59: you can use not gt here

LiudmylaKulinchenko (changes requested):

app/main.py
oleksandra-shershen: Done

app/main.py
oleksandra-shershen: Done

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return self.km > other
+
+    def __eq__(self, other: ("Distance", int, float)) -> bool:
+        return self.km == other
+
+    def __le__(self, other: ("Distance", int, float)) -> bool:
+        return not self > other
+
+    def __ge__(self, other: ("Distance", int, float)) -> bool:
+        return not self < other
```
Oleksl888, line 62: this implementation will require your decorator to be added to make this code work

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
+import functools
+from typing import Callable, Any
+
+
+def validate_operand(func: Callable) -> Callable:
+    @functools.wraps(func)
+    def wrapper(self: "Distance", other: ("Distance", int, float)) -> Any:
```
Oleksl888, line 7: You could omit quotes on DIstance type if you use annotations from __future__

app/main.py
oleksandra-shershen: @Oleksl888 tests pass without the use of a decorator

app/main.py
oleksandra-shershen: Resolved

app/main.py
Oleksl888: This ' not self < other' construct should have worked fine if we defined handling ints and floats in __lt__ , __gt__ , __eq__ methods but they are not. let me check and get back to you on that.

Oleksl888 (approved):
WEll done!

app/main.py
Oleksl888: I see - since you added your decorator on _add__ methods it extends it validation on comparison methods as well. My mistake

PR #1101, opened 2024-03-18
Solution

SerhiiMar:


Nyxan (approved):
I don't see any mistaces. Good job

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | int") -> "Distance":
```
YuriiKindrat, line 11: why do u need to use quotes here?
the same is for other annoatations

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other.real)
```
YuriiKindrat, line 14: why do u access real attr of integer ?

YuriiKindrat: the same is for other part of code

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: "Distance | int") -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other.real
+
+    def __ge__(self, other: "Distance | int") -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other.real
```
YuriiKindrat, line 56: u can use previous created methods here, and make this methods in one line

YuriiKindrat (changes requested):

app/main.py
SerhiiMar: Done

app/main.py
SerhiiMar: Done

app/main.py
SerhiiMar: Done

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __gt__(self, other: Union["Distance", int]) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __le__(self, other: Union["Distance", int]) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Union["Distance", int]) -> bool:
+        return not self.__lt__(other)
```
YuriiKindrat, line 55: u can just write `<` or `>` here instead of __lt__, __gt__

YuriiKindrat (approved):

app/main.py
SerhiiMar: Done

PR #1102, opened 2024-03-19
Implement Distance class

bwaah1:


y-kondrashova (approved):
Good Job 😁👍

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
YuriiKindrat, line 61: could u use previouse created methods for this ones ? U could write them in one line

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
YuriiKindrat, line 14: no need to write Distnace with quotes in type annotation, it will work fine without them

YuriiKindrat (changes requested):

app/main.py
bwaah1: It gives an error "Unresolved reference 'Distance' ", because we do it directly in the body of the class

Oleksl888: You could import annotations from __future__ to solve the issue with type not being recognized inside class definition

Oleksl888 (approved):
Good job!

PR #1104, opened 2024-03-19
'Solution'

MarianaTykha:


app/main.py
```diff
@@ -1,3 +1,81 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,82 @@
...
```
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

vasylhnatiuk (changes requested):

PR #1105, opened 2024-03-19
'Solution'

dkotkod:


bwaah1 (approved):

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
Oleksl888, line 11: Check the typing in all your methods. For example addition is also allowed for built in numeric types

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError(
+                "Unsupported operand type for +: 'Distance' and {}"
+                .format(type(other))
+            )
```
Oleksl888, line 20: You can simplify this construct with less if statements. Please check this in all of your methods

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ("Distance", float, int)) -> "Distance":
```
Oleksl888, line 11: You can import annotations from __future__ module to omit using quotes on class name typing.
Also when you declare types in parenthesis with commas , those types are considered as individual arguments to type hints. You should use `|` symbol or use `Union` from typing to show that any of those types is possible input

app/main.py
```diff
@@ -1,3 +1,52 @@
...
+        return self.km > (
+            other if isinstance(other, (float, int)) else other.km
+        )
+
+    def __eq__(self, other: ("Distance", float, int)) -> bool:
+        return self.km == (
+            other if isinstance(other, (float, int)) else other.km
+        )
+
+    def __le__(self, other: ("Distance", float, int)) -> bool:
```
Oleksl888, line 44: try to rewrite __le__ and __ge__ in one line each based on magic methods you already defined above.
 (Hint: `<=` is the same as `not >`)

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            other if isinstance(other, (float, int)) else other.km
+        )
+
+    def __eq__(self, other: Union[Distance, float, int]) -> bool:
+        return self.km == (
+            other if isinstance(other, (float, int)) else other.km
+        )
+
+    def __le__(self, other: Union[Distance, float, int]) -> bool:
+        return self.km <= (
```
Oleksl888, line 49: Ok so think about it: you have `less then or equal` method here. You already defined handling of `other` object for comparison in your `__lt__` method. Now thinking logically, if a < b it also means that b >= a. Now try to use this logic  here and in `__ge__`, i promise you will like it

Oleksl888 (changes requested):

Oleksl888 (approved):
Very well!

PR #1107, opened 2024-03-19
Solution

y-kondrashova:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float | Distance) -> Distance:
```
panicua, line 26: ```other``` cannot be ```Distance```, right? so you don't need to write type ```Distance```

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self,
+                    other: int | float | Distance) -> Distance | None:
```
panicua, line 30: Same here I belive

NVIDIA_Share_y2g8C8uW2E

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
panicua, line 57: I don't know if mentors will request if from you but for me they suggested that I use previous methods so you don't need to use an ```if instance``` statement here + you'll have one return instead of 2 for each magic method.

panicua (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self,
+                    other: int | float) -> Distance | None:
+        if other:
```
YuriiKindrat, line 31: why do u need this ? the error will occure if we call  this method without other before this check )

app/main.py
YuriiKindrat: yeah, @panicua is right
use previous magic methods for qe,le

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ!

PR #1109, opened 2024-03-19
Solution 1 (class Distance)

semermaks:


yossosh (approved):
Good job! :)

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
YuriiKindrat, line 61: try to rewrite these two methods in one line each. Use previously created magic methods

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
Oleksl888, line 15: try to use the below pattern in future code to avoid extra if statements and returns:
```
    other = Distance(other) 

Oleksl888 (approved):
Good job!

PR #1110, opened 2024-03-19
Solution

sergeybrekhov:


Chub-andrew (approved):
Good job.

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return self.km == other.km if isinstance(other, Distance)\
+            else self.km == other
+
+    def __le__(self, other: "Distance") -> "Distance":
+        return self.km <= other.km if isinstance(other, Distance)\
+            else self.km <= other
+
+    def __ge__(self, other: "Distance") -> "Distance":
+        return self.km >= other.km if isinstance(other, Distance)\
+            else self.km >= other
```
YuriiKindrat, line 59: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        return self.km < other.km if isinstance(other, Distance)\
+            else self.km < other
+
+    def __gt__(self, other: "Distance") -> "Distance":
+        return self.km > other.km if isinstance(other, Distance)\
+            else self.km > other
+
+    def __eq__(self, other: "Distance") -> "Distance":
+        return self.km == other.km if isinstance(other, Distance)\
+            else self.km == other
```
YuriiKindrat, line 51: try to avoid using linebreakers, u can break the line with some of braces

YuriiKindrat (changes requested):

app/main.py
sergeybrekhov: > try to avoid using linebreakers, u can break the line with some of braces

should i use 1liner to avoid linebreakers?

YuriiKindrat: u could break the line by the brackets

app/main.py
YuriiKindrat: u didn't fix that

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: u didn't fix that

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __le__(self, other: "Distance") -> "Distance":
+        return not self.__gt__(other)
+
+    def __ge__(self, other: "Distance") -> "Distance":
+        return not self.__lt__(other)
```
YuriiKindrat, line 57: no need to call methods like this 
use `<, >` instead

YuriiKindrat (changes requested):

PR #1111, opened 2024-03-19
'Solution'

yossosh:


sergeybrekhov (approved):

YuriiKindrat (changes requested):

yossosh:
My code doesn't work, when I write Distance with quotes in annotations. The problem looks like: **Unresolved reference 'Distance'**. I can't fix it**:(**

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, int) or isinstance(other, float):
+            return self.km == other
+        else:
+            raise TypeError("Unsupported operand type(s) for ==: "
+                            "'Distance' and '{}'".format(type(other).__name__))
+
+    def __le__(self, other: int or float) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
Oleksl888, line 72: you could just do this so it returns the opposite of __gt__ method that you defined above

`
not self > other

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int or float) -> "Distance":
+        if isinstance(other, Distance):
```
Oleksl888, line 12: You could reduce the amount of conidtions by following this pattern:

```
            other = Distance(other)

Oleksl888 (approved):
it's a pass

PR #1112, opened 2024-03-20
Solution

VladyslavKsh:


app/main.py
```diff
@@ -1,3 +1,55 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Gilgumish, line 5: Add a condition that the user needs to provide only positive numbers

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __gt__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
```
Gilgumish, line 45: == does not always compare values correctly. This page was recommended to me by a mentor, and it describes this case in more detail
https://docs.python.org/3/tutorial/floatingpoint.html

Gilgumish (approved):
In general, the code works correctly, but it can be improved

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
YuriiKindrat, line 14: no need to use quotes for Distance annotation

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 55: could u use previous created methods here?
`hint <= is the same as not >`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __eq__(self, other: Union[Distance, int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Union[Distance, int, float]) -> bool:
+        if isinstance(other, Distance):
+            return not (self.km > other.km)
+        return not (self.km > other)
```
Oleksl888, line 53: since you defined __lt__ and __eq__ methods above you could simplify this to one line:
return not self > other

Oleksl888 (approved):
good job!

PR #1113, opened 2024-03-20
Solution

messedxup:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        return not self > other
```
VladyslavKsh, line 49: I believe that method __le__() is "less than or equal to" operator <=, so maybe we can change the method using this operator for different types of "other" and self.km (same with __ge__() )

VladyslavKsh (changes requested):
There is something to change

VladyslavKsh (approved):
Nice done!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: float | int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance | None:
+        if other:
```
YuriiKindrat, line 30: why do u need this check here ? it seems to me, that error will raised if arg won't be passed

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | float | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
YuriiKindrat, line 56: try to use previous created methods here
`hint: <= is the same as not >`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
```
Oleksl888, line 15: You could simplify this by using the following logic here and below:

```
            other = Distance(other)

Oleksl888 (approved):
good job!

PR #1117, opened 2024-03-20
Solution

siaserb:


alonasorochynska (approved):
Good job!

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 55: could u use previous created method here ?
`hint: <= is the same as not >`

YuriiKindrat (changes requested):

app/main.py
siaserb: This makes a lot more sense, thanks

Oleksl888 (approved):
Well done!

PR #1118, opened 2024-03-20
'Solution'

alonasorochynska:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        return self.km >= other.km
```
YuriiKindrat, line 56: try to rewrite these two methods in one line each. Use previously created magic methods
`Hint: <= is the same as not >`

YuriiKindrat (changes requested):

Oleksl888 (changes requested):
> try to rewrite these two methods in one line each. Use previously created magic methods
> `Hint: <= is the same as not >`

can you please change as per Yuri's request?

Oleksl888 (approved):
GJ!

PR #1119, opened 2024-03-21
Solution

AlexNechval:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
NazarSheremet, line 18: You can simplify the code by removing the `elif`, because if the first condition is met, the second check won't be executed. Also, in newer versions of Python, it's better to use` __future__` for annotations.

```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
        if isinstance(other, Distance):
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```

NazarSheremet (changes requested):
Need to implement suggestion for all methods

NazarSheremet (approved):
Good job!

PR #1121, opened 2024-03-22
'Solution'

March1205:


app/main.py
```diff
@@ -1,3 +1,78 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
vsmutok, line 2:
```suggestion
    def __init__(self, km: float | int) -> None:
```

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        self.km: float = km
+
+    def __str__(self) -> str:
+        return "Distance: {} kilometers.".format(self.km)
+
+    def __repr__(self) -> str:
+        return "Distance(km={})".format(self.km)
+
+    def __add__(self, other: "Distance") -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vsmutok: fix everywhere

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
vsmutok, line 16: else redundant here

vsmutok (changes requested):

March1205:
I couldn't use '|' operator so used 'Union' instead, as my code kept failing without it. I also tried to do Distance without quotes but I kept getting NameError as name couldn't be defined and even by defining it at the end I still had to use quotes in most parts. If you could suggest something to complete this code as you told me, please do, because I might be missing some things.

app/main.py
```diff
@@ -1,3 +1,80 @@
...
+            return self.km == other
+        else:
+            raise TypeError("Unsupported operand type for ==")
+
+    def __le__(self, other: Union["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        else:
```
vsmutok, line 71: `else` redundant in this case

vsmutok (changes requested):

vsmutok:
> I couldn't use '|' operator so used 'Union' instead, as my code kept failing without it. I also tried to do Distance without quotes but I kept getting NameError as name couldn't be defined and even by defining it at the end I still had to use quotes in most parts. If you could suggest something to complete this code as you told me, please do, because I might be missing some things.


try use:
```python
```

PR #1126, opened 2024-03-23
Solution

frontbastard:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        return Distance(self.km + self.validator(
+            other, int, float, distance=Distance
+        ))
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
vsmutok, line 35:
```suggestion
    def __iadd__(self, other: Distance | int | float) -> Distance:
```
fix everywhere

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    @staticmethod
+    def validator(
+            other: Union["Distance", int, float],
```
vsmutok, line 10:
```suggestion
            other: Union[Distance, int, float],
```
fix everywhere

vsmutok (changes requested):

app/main.py
frontbastard: When I do so I have an error:
image

frontbastard: Added `from __future__ import annotations` and now it's ok.

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __mul__(self, other: Union[Distance, int, float]) -> "Distance":
+        return Distance(self.km * self.validator(other, int, float))
+
+    def __truediv__(self, other: Union[Distance, int, float]) -> "Distance":
+        result_km = round(self.km / other, 2)
+        return Distance(result_km)
+
+    def __lt__(self, other: Union[Distance, int, float]) -> bool:
+        return self.km < other
```
vasylhnatiuk, line 48: if other is Distance you should compare other.km instead of other

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

app/main.py
frontbastard: The tests passed, so I decided that this is how it should be )

vsmutok (approved):
GJ!

PR #1127, opened 2024-03-24
Solution distance class

TenbReala:


app/main.py
```diff
@@ -1,3 +1,50 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Dimosphen1, line 5: Indicate the expected type explicitly in all places without omitting `float`

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int) -> bool:
+        tmp = other or other.km
+        return self.km < tmp
```
Dimosphen1, line 34: To avoid a possible attribute error, perform an explicit `isinstance` check here and in the other similar methods

Dimosphen1 (changes requested):

PR #1128, opened 2024-03-24
Solution

Lebid-Dmytro:


app/main.py
```diff
@@ -1,3 +1,55 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Dimosphen1, line 5: Indicate the expected type explicitly in all places without omitting `int` in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #1131, opened 2024-03-24
'Solution'

afaidiuk:


Ruxxell54 (approved):
Perfect)

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
Oleksl888, line 11: use `annotations` from `__future__` so you don't have to put class names in quotes.
Also check the types for this and below methods. (`other` can be other types as well)

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
```
Oleksl888, line 12: Think how you can simplify this. Ideally you can write it with one `if` statement and one `return`

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError("Unsupported operand type for +")
```
Oleksl888, line 17: I know you haven't learned exceptions yet but since you are using them the error message should be more descriptive - you can mention allowed types here or mention types that caused an error

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        else:
+            raise TypeError("Unsupported operand type for *")
+
+    def __truediv__(self, other: float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError("Unsupported operand type for /")
+
+    def __lt__(self, other: "Distance") -> bool:
```
Oleksl888, line 40: Same below: check the typing, try to simplify if-elif-else if possible.

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+
+    def __eq__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        else:
+            raise TypeError("Unsupported operand type for ==")
+
+    def __le__(self, other: "Distance") -> bool:
```
Oleksl888, line 64: `__le__` and `__ge__` methods can be simplified since you already defined methods above. (for example `less then or equal` is the opposite to `greater then`

tests/test_main.py
```diff
@@ -6,16 +6,7 @@
 from app.main import Distance
```
Oleksl888, line 6: Please revert changes you made in tests.

Oleksl888 (changes requested):

tests/test_main.py
vsmutok, line 1: Please revert changes you made in tests.

vsmutok (changes requested):

tests/test_main.py
afaidiuk: I reverted the changes

DaTrEvTeR (approved):
Well done!

PR #1133, opened 2024-03-25
'Solution'

kssshevchenko:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        if isinstance(other, (int, float)):
+            return self.km <= other
```
VladyslavKsh, line 55: I was told that we can use "not (self.km < other.km)", because we've already implemented the method ">" or __gt__, same with "<" or __lt__. But I guess it's not a mistake

VladyslavKsh (approved):
Looks fine

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km) -> int:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other):
```
Oleksl888, line 11: What types does the method take as params and which types does it return. Check this everywhere

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
```
Oleksl888, line 12: Think how you can simplify this to use only one `if` statement. Check in your other methods as well

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other):
+        if isinstance(other, (int, float)):
+            result = self.km / other
```
Oleksl888, line 30: You don't need to store result in variable since you are not going to reuse this value

app/main.py
Oleksl888: You can simplify __le__ and __ge__ methods using opposite operator logic because if `a <= b` it also means that `a not > b`

Oleksl888 (changes requested):
Your code fails flake8 check. Please make sure you run tests and flake8 check locally before you submit

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __eq__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, (int, float)):
+            return self.km == other
+
+    def __le__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
```
fsocie7y, line 52: do it in the same way as below (not self.km > self.other)

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, (int, float)):
+            return not self.km > other
+
+    def __ge__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        if isinstance(other, (int, float)):
+            return self.km >= other
```
fsocie7y, line 60: do it as in __le__

fsocie7y (changes requested):

PR #1147, opened 2024-04-05
The solution of the task

Ja0ni:


app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
Dimosphen1, line 14: Indicate that `other` should be `float` as well in the type annotation

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        return Distance(self.km + other.km) \
```
Dimosphen1, line 15: Place if condition in several lines to avoid using line breaking operator `\` here and in the other places as well

Dimosphen1 (changes requested):

PR #1148, opened 2024-04-08
Solution

IKruglyak:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        else:
+            raise TypeError("Unsupported operand type for /")
+
+    def __lt__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        else:
+            raise TypeError("Unsupported operand type for <")
```
NazarSheremet, line 46: In this method, you expect that `other` can be int or float, but in annotation, it is not added

NazarSheremet (changes requested):
Implement suggestions for all methods where it is needed

NazarSheremet (approved):
Good job!

PR #1152, opened 2024-04-11
Solution

BitRxp:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __mul__(self, other: Any) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Any) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Any) -> bool:
+        if isinstance(other, Distance):
+            return True if self.km < other.km else False
```
vasylhnatiuk, line 35:
```suggestion
            return self.km < other.km
```

vasylhnatiuk: fix it everywhere

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: Any) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> "Distance":
```
vasylhnatiuk, line 14: other can't be Any, use more accurate annotation

```suggestion
    def __add__(self, other: Any) -> Distance:
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" | int | float) -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

PR #1155, opened 2024-04-14
Solution

RuslanLevchuk:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @classmethod
+    def get_other(cls, other: Distance | int | float) \
+            -> Distance | int | float:
+        if isinstance(other, cls):
+            return other.km
+        return other
```
vsmutok, line 19: using a static method will be more natural because your method has no connection to other instances of the class.

vsmutok (changes requested):

PR #1161, opened 2024-04-18
Solution

Jejemira:


app/main.py
```diff
@@ -1,3 +1,70 @@
...
+                km=self.km + other
+            )
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
EdAlekseiev, line 24: `other` can be float too. You should show it in type hint
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
        if isinstance(other, Distance):
            return Distance(
                km=self.km + other.km
            )
        else:
            return Distance(
                km=self.km + other
            )

    def __iadd__(self, other: Distance | int | float) -> Distance:
        if isinstance(other, Distance):
            self.km += other.km
            return self
        else:
            self.km += other
            return self
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,70 @@
...
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(
+                km=self.km + other.km
+            )
+        else:
```
vasylhnatiuk, line 13: you can avoid else here. do it everywhere

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1162, opened 2024-04-18
Solution

IraMladanovych02:


app/main.py
```diff
@@ -1,3 +1,64 @@
+from __future__ import annotations
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
Dimosphen1, line 6: Indicate that `km` could be `int` as well

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        self.km: float = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
```
Dimosphen1, line 15: Indicate that `other` could be a `Distance` class instance as well. Add corresponding changes to the other methods  
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,64 @@
+from __future__ import annotations
+from typing import Union
```
Dimosphen1, line 2: Do not use `Union`, use `|` instead

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
pavlopro, line 5: int | float

pavlopro (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
pavlopro, line 27: So why did you include `Distance` here ?

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
```
pavlopro, line 31: And here

pavlopro (changes requested):

app/main.py
IraMladanovych02: Sorry, I double checked my code and understood the main issue of annotation. All changes were done

pavlopro (approved):
GJ!

PR #1163, opened 2024-04-18
Implemented Solution

illiamdo:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
YuriiSmolii, line 11: You did wrong Distance class annotation, "Distance" is a str type value.

YuriiSmolii (changes requested):
Fix annotation

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> str:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance") -> str:
```
Alkaponees:
```suggestion
    def __iadd__(self, other: "Distance") -> Distance:
```
Use Distance class to add iadd, mul and truediv functions

Alkaponees (changes requested):
Check my review

app/main.py
illiamdo: Should i use Distance class inside methods or in annotaions?

Alkaponees (approved):
LGTM!

PR #1169, opened 2024-04-23
triukhan solution

triukhan:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def operate(
+            self,
+            other: Union[int, "Distance"],
```
triukhan, line 17: знайшов, що подвійну анотацію для класу можна зробити тільки так
якщо напиписати other: int | "Distance", то видає помилку

LiudmylaKulinchenko: Можливо ти використовуєш пайтон 3.8 (чи 3.9), у якому ще немає можливості додавати анотації таким чином?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return self.compare(other, lambda x, y: x > y)
+
+    def __eq__(self, other: Union[int, "Distance"]) -> bool:
+        return self.compare(other, lambda x, y: x == y)
+
+    def __le__(self, other: Union[int, "Distance"]) -> bool:
+        return self.compare(other, lambda x, y: x <= y)
+
+    def __ge__(self, other: Union[int, "Distance"]) -> bool:
+        return self.compare(other, lambda x, y: x >= y)
```
LiudmylaKulinchenko, line 65: try to rewrite these two methods in one line each. Use previously created magic methods
Hint: <= is the same as not >

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            self.km += other
+            return self
+
+    def __mul__(self, other: int) -> "Distance":
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int) -> "Distance":
+        return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: Union[int, "Distance"]) -> bool:
```
LiudmylaKulinchenko, line 52: what if other is float?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def operate(
+            self,
+            other: Union[int, "Distance"],
+            operation: Callable
+    ) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(km=operation(self.km, other.km))
+        else:
```
LiudmylaKulinchenko, line 22: No need to use else after return inside `if` block

LiudmylaKulinchenko (changes requested):
I'm not sure it's necessary to use such complicated constructions with lambda functions in a simple cases. You anyway duplicate your code in this task, but now it's more diffirult to understand, as for me
You can use `other_km = other.km if isinstance(other, Distance) else other` in functions instead

app/main.py
triukhan: 3.11

triukhan:
> I'm not sure it's necessary to use such complicated constructions with lambda functions in a simple cases. You anyway duplicate your code in this task, but now it's more diffirult to understand, as for me You can use `other_km = other.km if isinstance(other, Distance) else other` in functions instead

ясно як завжди перестарався

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def compare(
+            self,
+            other: Union[int, "Distance"],
+            comparison: Callable
+    ) -> bool:
+        if isinstance(other, Distance):
+            return comparison(self.km, other.km)
+        else:
+            return comparison(self.km, other)
```
LiudmylaKulinchenko, line 32: You don't use this function, you can remove it

LiudmylaKulinchenko: Don't forget to remove Callable import

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __eq__(self, other: Union[int, float, "Distance"]) -> bool:
+        other_km = other.km if isinstance(other, Distance) else other
+        return self.km == other_km
+
+    def __le__(self, other: Union[int, "Distance"]) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Union[int, "Distance"]) -> bool:
+        return not self.__lt__(other)
```
LiudmylaKulinchenko, line 67: U can use operands instead of calling methods directly
> instead of .__gt__

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __add__(self, other: Union[int, "Distance"]) -> "Distance":
+        return self.operate(other, lambda x, y: x + y)
+
+    def __iadd__(self, other: Union[int, "Distance"]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
LiudmylaKulinchenko, line 43: avoid return duplicating. You can move it outside of if/else block

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+            other: Union[int, "Distance"],
+            operation: Callable
+    ) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(km=operation(self.km, other.km))
+        return Distance(km=operation(self.km, other))
+
+    def compare(
+            self,
+            other: Union[int, "Distance"],
```
LiudmylaKulinchenko, line 26: Use `from __future__ import annotations` and remove quotes in annotations

app/main.py
LiudmylaKulinchenko: change to | from Union, don't forget to remove import

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self,
+            other: int | Distance,
+            operation: Callable
+    ) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(km=operation(self.km, other.km))
+        return Distance(km=operation(self.km, other))
+
+    def __add__(self, other: int | Distance) -> "Distance":
+        return self.operate(other, lambda x, y: x + y)
```
LiudmylaKulinchenko, line 25: No need to create an `operate` method if you'll use it only once

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🔥🔥

PR #1176, opened 2024-04-28
Solution1

D0mln0:


vasylhnatiuk (changes requested):
fix tests

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> object:
```
EdAlekseiev, line 14: `other` can be `Distance` according to the task. Fix everywhere, please
```suggestion
    def __add__(self, other: int | float | Distance) -> object:
```

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        if isinstance(other, int | float):
+            distance2 = self.km + other
+            return Distance(distance2)
+        else:
+            distance3 = self.km + other.km
+            return Distance(distance3)
```
EdAlekseiev, line 20: You can rewrite it with without `else` statement. It makes your code more readable
```suggestion
        if isinstance(other, int | float):
            distance2 = self.km + other
            return Distance(distance2)
        distance3 = self.km + other.km
        return Distance(distance3)
```

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+
+    def __iadd__(self, other: int | float) -> object:
+        if isinstance(other, int | float):
+            self.km += other
+            return self
+        else:
+            self.km += other.km
+            return self
```
EdAlekseiev, line 28: You can rewrite it with one `return` statement. It makes your code more readable
```suggestion
        if isinstance(other, int | float):
            self.km += other
        else:
            self.km += other.km
        return self
```

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+        return self.km > other
+
+    def __eq__(self, other: object) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Any) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Any) -> bool:
+        return self.km >= other
```
EdAlekseiev, line 51: `other` can be `Distance`, and you should handle this case inside your code

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __mul__(self, distance: int | float) -> object:
+        distance2 = self.km * distance
+        return Distance(distance2)
+
+    def __truediv__(self, distance: int | float) -> object:
+        distance2 = self.km / distance
+        return Distance(round(distance2, 2))
+
+    def __lt__(self, distance: Any) -> bool:
+        return self.km < distance
```
vasylhnatiuk, line 37: distance can be Distance as well, you should use distance.km in this case

vasylhnatiuk (changes requested):

app/main.py
D0mln0: if i use distance.km code don`t work

vasylhnatiuk: you should add two conditions, and use `distance.km` or `distance` depends on object

vasylhnatiuk (changes requested):
you should add two conditions, and use distance.km or distance depends on object

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, distance: "Distance") -> object:
```
EdAlekseiev, line 11: This function returns `Distance`. Fix type hint for return type everywhere
```suggestion
    def __add__(self, distance: "Distance") ->Distance:
```

EdAlekseiev, line 11: `distance` can be float and int. Fix it everywhere
```suggestion
    def __add__(self, distance: "Distance" | int | float) -> object:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1179, opened 2024-04-30
Solution

TsvetanKichuk:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | float | int") -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            return NotImplemented
```
pavlopro, line 17: Return Type Error with some message instead of NotImplemented
Fix it everywhere

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | float | int") -> "Distance":
```
pavlopro, line 11: Import annotations from __future__ module and correct annottions

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
pavlopro, line 16: Else is redundunt here
Fix it everywhere

pavlopro (changes requested):

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
pavlopro, line 19: If you don't want to return a TypeError, it means you expect the inputs to be either Distance, float, or int. Therefore, the elif is redundant in this case. You can simply return Distance(self.km + other). Please make this adjustment throughout your code.

pavlopro (changes requested):

app/main.py
TsvetanKichuk: remove all lines  the elife ?

pavlopro: if isinstance(other, Distance):

TsvetanKichuk: I already understood, but I did as you originally asked

TsvetanKichuk:
I saw it late

pavlopro (approved):
Good Job!

PR #1182, opened 2024-05-03
'Solution'

VadimkaLYZ:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, int | float):
+            return Distance(self.km + other)
```
EdAlekseiev, line 19: `elif` statement is redundant, because `if` contains return
```suggestion
        if isinstance(other, Distance):
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, int | float):
+            self.km += other
+            return self
```
EdAlekseiev, line 27: The same comment as above
```suggestion
        if isinstance(other, Distance):
            self.km += other.km
            return self
        self.km += other
        return self
```

EdAlekseiev (changes requested):

PR #1185, opened 2024-05-07
Solution

Katy4S:


app/main.py
```diff
@@ -1,3 +1,107 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
EdAlekseiev, line 11: other can be `float` and `int`, you should show it in type hint. Fix everywhere, please
```suggestion
    def __add__(self, other: "Distance" | float | int) -> "Distance":
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,110 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", float, int]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[Distance, float, int]) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,110 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError(
+                "Unsupported operand type(s) for +: 'Distance' and '{}'"
+                .format(type(other).__name__)
+            )
```
vsmutok, line 23:
```suggestion
        elif isinstance(other, (int, float)):
            return Distance(self.km + other)
        raise TypeError(
            "Unsupported operand type(s) for +: 'Distance' and '{}'"
            .format(type(other).__name__)
        )
```
in this case, it is not necessary to write `else'

vsmutok (changes requested):

app/main.py
Katy4S: why? thank you

Katy4S: It all has fallen, 0 tests passes

app/main.py
Katy4S: this also does not work. Can I leave it as it is, please? 

If it works == True : 
     do not touch

app/main.py
EdAlekseiev: If you add `from __future__ import annotations` it will work

EdAlekseiev (changes requested):

PR #1193, opened 2024-05-15
py-distance-class v1

BornToLivee:
Please tell me, should i add #comments in a future?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,59 @@
+from typing import Any
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


Oleg-s-s:
> Please tell me, should i add #comments in a future?

Up to you, if you think you need to explain some places more for better understanding you can do it

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Any) -> Any:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
+            self.km += other
+            return self
```
Oleg-s-s, line 26: You can use return only once. Optimize this part

Oleg-s-s (changes requested):
Fix your annotations. other cant be `Any`, and your methods are not returning `Any`

DevSpaciX (approved):
Great!

PR #1196, opened 2024-05-18
"Solution"

arsenmarkotskyi:


app/main.py
```diff
@@ -1,3 +1,42 @@
...
+            self.km += float(other)
+        return self
+
+    def __mul__(self, other: int) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int) -> bool:
```
vasylhnatiuk, line 29: other can be Distance instance, float as well

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [int, float, "Distance"]) -> "Distance":
```
```suggestion
    def __add__(self, other: [int, float, Distance]) -> Distance:
```

vasylhnatiuk (changes requested):

app/main.py
arsenmarkotskyi: can u plz explain me, why my code doesn’t working , when i use Distance without  " "?
 NameError: name 'Distance' is not defined

app/main.py
arsenmarkotskyi: Should i also do it for others func, yes?

app/main.py
vsmutok: try to use
```python
```
 to solve this problem

vsmutok (changes requested):

app/main.py
arsenmarkotskyi: great, thanks!

vsmutok (approved):
Great!

PR #1198, opened 2024-05-19
'Solution'

BohdanKuzik:


Oleg-s-s (changes requested):
Fix your type annotations, object is bad annotation for parameter, use | and specify which params can be inputted.
Fix your methods where you have multiple returns, optimize it to use return only once. (take in mind there is no need to check for isinstance everywhere if your type annotations will be correct)

DevSpaciX (approved):
Good job !

PR #1200, opened 2024-05-20
solution

AndrewwSSS:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, dist: int) -> Distance:
+        return Distance(self.km * dist)
+
+    def __truediv__(self, dist: int) -> Distance:
+        return Distance(round(self.km / dist, 2))
```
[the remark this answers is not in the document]
AndrewwSSS: Can be error, but there are not cases in task how to be in that case. I don't know what to do in this case. Return self or division by 1  instead of 0? Raise exception ? We haven't done this theme

app/main.py
```diff
@@ -1,3 +1,51 @@
...
+            other = Distance(other)
+        return self.km > other.km
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if not isinstance(other, Distance):
+            other = Distance(other)
+        return self.km == other.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self < other or self == other
```
DevSpaciX, line 48: It is better to check by using this operator -> `<=`
Also why you do not check possible variants if `other` will be `Distance` type ? ( And in __ge__ method also )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __mul__(self, dist: int | float) -> Distance:
+        return Distance(self.km * dist)
+
+    def __truediv__(self, dist: int | float) -> Distance:
+        return Distance(round(self.km / dist, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if not isinstance(other, Distance):
+            other = Distance(other)
```
DevSpaciX, line 34: I think we do not need really to create instance of class just for check `if self.km < other` and return it 
And in case `if  isinstance(other, Distance):` --> check as you did `return self.km < other.km`

DevSpaciX (changes requested):
One last error for check , sorry I didn't see it right away

app/main.py
AndrewwSSS: https://github.com/mate-academy/py-distance-class/pull/1200#discussion_r1607269337
But mentor here say the opposite)

Oleg-s-s (approved):
:+1:

PR #1202, opened 2024-05-21
Solution

Yuruch:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+                km=self.km + other.km
+            )
+        return Distance(
+            km=self.km + other
+        )
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
DevSpaciX, line 26: Try to use here inly one `return`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1204, opened 2024-05-21
Solution

tpavliohlo:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return self
+
+    def __mul__(self, other: float | int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float | int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return self.km < other
```
vasylhnatiuk, line 34: if other is Distance instance you should use other.km

vasylhnatiuk (changes requested):

app/main.py
tpavliohlo: I don't understand your comment, why I should use other.km, because code is working

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
vsmutok, line 18:
```suggestion
            return Distance(self.km + other.km)
        return Distance(self.km + other)
```
`else` redundant in this case

app/main.py
```diff
@@ -1,3 +1,46 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```
vsmutok, line 5:
```suggestion
    def __init__(self, km: float | int) -> None:
```
`int` also possible

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1205, opened 2024-05-21
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __mul__(self, other: Union[float, int]) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[float, int]) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union["Distance", float, int]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
DevSpaciX, line 40: You don’t need `else` if you have a `return` statement in `if`. Just call the second terurnt on the same lvl as `if` statement

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __iadd__(self, other: Union["Distance", float, int]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[float, int]) -> "Distance":
+        return Distance(self.km * other)
```
DevSpaciX, line 32: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __truediv__(self, other: Union[float, int]) -> "Distance":
+        return Distance(round(self.km / other, 2))
```
DevSpaciX, line 35: `other` can not be of type `Distance` in this method

DevSpaciX (changes requested):

app/main.py
dsokolovskiy: what exactly do i need to change in the code?

app/main.py
DevSpaciX: Yes, I wanted to see additional validation here, but it’s not necessary here, you’re right, everything is correct : )

app/main.py
DevSpaciX: All correct for now ) Never mind , you did well 👍

app/main.py
DevSpaciX: @dsokolovskiy 
I see you corrected the code in this place, what about other places where the same error occurs?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __lt__(self, other: Union["Distance", float, int]) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        if isinstance(other, (float, int)):
+            return self.km < other
+
+    def __gt__(self, other: Union["Distance", float, int]) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
```
DevSpaciX, line 46: Fix it here as you did in method `__lt__` ( And in other places aslo )

DevSpaciX (changes requested):
Try to correct the error in all places at once.

dsokolovskiy:
Okay, thank you. I will fix these errors tomorrow as I don't have access to
my laptop

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", float, int]) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
DevSpaciX, line 18: Just for future 
image

DevSpaciX (approved):
Good , but one more place for improove exists )

PR #1206, opened 2024-05-21
add solution

jpdreamthug:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        return Distance(self.km + other.km
```
DevSpaciX, line 16: It is better to write it in like in checklist Imho 
image

DevSpaciX (approved):
Good! But look at the comment )

PR #1207, opened 2024-05-21
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+    def __le__(self, other: Distance | int | float) -> bool:
+        return (
+            self.km <= other.km
+            if isinstance(other, Distance)
+            else self.km <= other
+        )
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return (
+            self.km >= other.km
```
DevSpaciX, line 64: Imho more readable will be like here 
image

DevSpaciX (approved):
Nice !

PR #1208, opened 2024-05-21
distance-class solution

DimaYankovskiy:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
```
DevSpaciX, line 61: Yiu can remove `else` and write second `return` on the same level as `if` statement it will be more Pythonic : )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
DevSpaciX, line 25: Try too call `return` only once

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice

PR #1209, opened 2024-05-21
Solution

SrMusienko:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        return Distance(km=self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
DevSpaciX, line 24: `return self` is duplicated, take it outside if-else block : )

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(km=self.km * other)
```
DevSpaciX, line 27: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        return self.km > other
```
DevSpaciX, line 36: What if `other` is of type `Distance` here and in all the following methods? ( Same on `__lt__` method )

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(km=round(self.km / other, 2))
```
DevSpaciX, line 30: `other` can not be of type `Distance` in this method , pls check type here too

DevSpaciX (changes requested):

DevSpaciX (changes requested):
Pls pay attention to comments )

DevSpaciX (approved):

PR #1215, opened 2024-05-22
'solution'

MaglorFeanorson:


app/main.py
```diff
@@ -1,3 +1,100 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
DevSpaciX, line 20: You do not need actually rise the error here and in other places

DevSpaciX (changes requested):

DevSpaciX (approved):
Perfect !

PR #1218, opened 2024-05-22
Solution

juliazelena:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
vasylhnatiuk, line 14: other can float as well, write everywhere

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        if isinstance(other, Distance):
+            return Distance(
+                km=self.km + other.km,
+            )
+        return Distance(
+            km=self.km + other,
+        )
+
+    def __iadd__(self, other: Distance) -> Distance:
```
vasylhnatiuk, line 23:
```suggestion
    def __iadd__(self, other: Distance | float | int) -> Distance:
```

vasylhnatiuk (changes requested):

PR #1219, opened 2024-05-22
solution

AndreiVed:


app/main.py
```diff
@@ -1,3 +1,66 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object | int) -> object:
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vasylhnatiuk: change everywhere

vasylhnatiuk (changes requested):

app/main.py
AndreiVed: Screenshot 2024-05-23 at 12 29 33
i try to do it, but catch the error:  F821 undefined name 'Distance'

app/main.py
```diff
@@ -1,3 +1,69 @@
...
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(
+            km=round(self.km / other, 2)
+        )
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
```
vasylhnatiuk, line 44: you can avoid else if you use 2 returns, fix it everywhere

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1222, opened 2024-05-23
sol1

Unlie9:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+            self.km += other.km
+        if isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            total = self.km * other
+            return Distance(total)
+        else:
```
DevSpaciX, line 32: That is not necessary to raise the error here and in other places: )

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            total = self.km * other
+            return Distance(total)
```
DevSpaciX, line 31: You can return it without creating additional variables `return Distance(self.km * other)`

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+        return Distance(total)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
DevSpaciX, line 28: Also `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1223, opened 2024-05-23
'Solution'

Oleh323:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __add__(self, add_: int | Distance) -> Distance:
+        if isinstance(add_, (int, float)):
+            return Distance(self.km + add_)
+        return Distance(self.km + add_.km)
+
+    def __iadd__(self, iadd_: int | Distance) -> Distance:
+        if isinstance(iadd_, (int, float)):
+            self.km += iadd_
+            return self
```
DevSpaciX, line 22: Try to call `return` here only once ; )

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        else:
+            self.km += iadd_.km
+            return self
+
+    def __mul__(self, mul_: int) -> Distance:
```
DevSpaciX, line 27: type of mul_ can be `float` also . Check in other places where argument can be `int` and `float`

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job!

PR #1224, opened 2024-05-23
'Solution'

Irina17191:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise TypeError
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
```
DevSpaciX, line 25: Try to call `return` here only once

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __repr__(self: Distance) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
DevSpaciX, line 19: That is not necessary to raise the error here : )

DevSpaciX (changes requested):

DevSpaciX (approved):
Perfect !

PR #1227, opened 2024-05-26
Solution

YevaYe:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
DevSpaciX, line 26: Pls try to call `return` only once

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
```
DevSpaciX, line 37: Pls remove redundat `else` statement . Just call the second `return` on the same lvl as `if isinstance(other, Distance):`

DevSpaciX (changes requested):

app/main.py
YevaYe: I don't understand what you mean. A have failed tests after it

PR #1235, opened 2024-05-29
test

VasylRysan:


app/main.py
```diff
@@ -1,3 +1,76 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            if self.km < other.km:
+                return True
+            return False
+        if self.km < other:
+            return True
+        return False
```
EdAlekseiev, line 40: You can simplify your code
```suggestion
        if isinstance(other, Distance):
            return self.km < other.km
        return self.km < other
```

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            if self.km > other.km:
+                return True
+            return False
+        if self.km > other:
+            return True
+        return False
```
EdAlekseiev, line 49: The same comment as above

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            if self.km == other.km:
+                return True
+            return False
+        if self.km == other:
+            return True
+        return False
```
EdAlekseiev, line 58: The same comment as above

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            if self.km <= other.km:
+                return True
+            return False
+        if self.km <= other:
+            return True
+        return False
```
EdAlekseiev, line 67: The same comment as above

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            if self.km >= other.km:
+                return True
+            return False
+        if self.km >= other:
+            return True
+        return False
```
EdAlekseiev, line 76: The same comment as above

EdAlekseiev (changes requested):

PR #1239, opened 2024-06-02
Solution

Vobilous:


app/main.py
```diff
@@ -1,3 +1,72 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float = 0) -> None:
```
Dimosphen1, line 2: Indicate that `km` could be `int` as well. Do not provide default values

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return "Distance: {} kilometers.".format(self.km)
+
+    def __repr__(self) -> str:
+        return "Distance(km={})".format(self.km)
+
+    def __add__(self, other: "Distance") -> "Distance":
```
Dimosphen1, line 11: Indicate that other could be int, float or a `Distance` class instance. Do not use quotes for type annotation
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```
Apply the same changes to the other places as well. Use `from __future__ import annotations` in case you face errors

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+        else:
+            raise TypeError("Unsupported operand type(s) for +")
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type(s) for +")
```
Dimosphen1, line 25: There is no need to raise `TypeError` you could omit it

Dimosphen1 (changes requested):

app/main.py
pavlopro: This was not resolved

app/main.py
pavlopro: This was not resolved

pavlopro (changes requested):
Not resolved.
Please check comments

PR #1241, opened 2024-06-04
Solution

andreykolomiec:


app/main.py
```diff
@@ -1,3 +1,69 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
```
EdAlekseiev, line 12: other can be int, float and Distance. You should show it in type hint. Fix everywhere.
```suggestion
    def __add__(self, other: int | float | Distance) -> "Distance":
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #1256, opened 2024-06-16
Solution

x0wer:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __init__(self, km: float | int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
Dimosphen1, line 14: Indicate that `other` could be `int` and `float` in the type annotation here and in the other places as well

Dimosphen1 (changes requested):

app/main.py
x0wer: yeah i already did this. check this request https://github.com/mate-academy/py-distance-class/pull/1256/commits/2825128ee55f3ed378beb4bc4cb0a09a4000065f

Dimosphen1: You should do it in the `__add__`, `__iadd__` and other methods. Your changes should be represented in the current PR (pull request), not in a separate commit.

Dimosphen1 (changes requested):

app/main.py
x0wer: Understandable, will do it tomorrow

PR #1258, opened 2024-06-17
Solution

Wallian169:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Distance | int) -> Distance:
```
Dimosphen1, line 16: Indicate that `other` could be `float`

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            raise TypeError("Operand must be of type Distance or int, float")
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
Dimosphen1, line 37: Indicate that other `could` be `Distance` class instance here and in the other methods as well where it is lacking

Dimosphen1 (approved):

Dimosphen1 (changes requested):

app/main.py
Wallian169: Should I add round() to other methods like in __truediv__ in case that input could be float?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, int | float):
```
pavlopro, line 25: Simplify this to just:
```
            self.km += other.km
else:
            self.km += other.km
return self
```

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            self.km += other
+        else:
+            raise TypeError("Operand must be of type Distance or int, float")
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round((self.km / other), 2))
```
pavlopro, line 35: Simplify to 
`return Distance(round(self.km / other, 2))`

app/main.py
```diff
@@ -1,3 +1,65 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    distances = []
```
pavlopro, line 5: Do you need this variable?

pavlopro (changes requested):

app/main.py
pavlopro: How about this one?

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1261, opened 2024-06-17
Solution

oukiz:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float, "Distance")) -> "Distance":
```
```suggestion
    def __add__(self, other: (int, float, Distance)) -> Distance:
```

vasylhnatiuk: use `from __future__ import annotations`

vasylhnatiuk (changes requested):

app/main.py
oukiz: Ye, sorry, I first wrote this code and only then saw another lesson.

PR #1264, opened 2024-06-19
Solution of magoc methods

Alex-Plugin:


app/main.py
```diff
@@ -1,3 +1,91 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ("Distance", int, float)) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
pavlopro, line 17: Else is redundant here
Correct it in all other places below where it can be corrected.

pavlopro (changes requested):

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ("Distance", int, float)) -> "Distance":
```
Dimosphen1, line 12: Change your type annotation here and in the other places to use `|` and class instances without `""`
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```
Use `from __future__ import annotations` import in case you face errors

Dimosphen1 (changes requested):

app/main.py
Alex-Plugin: Thank you for the suggestion, because I did it in the beginning but flake8 gave me the errors. but with __future__ everything became correct.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    # def __add__(self, other: ("Distance", int, float)) -> "Distance":
```
pavlopro, line 15: Remove commented code

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
```
pavlopro, line 21: When I said that the else is redundant, I meant this:
```
...
raise TypeError(
        f"Unsupported operand type(s) for +: "
        f"'Distance' and {type(other)}"
  )
```

In case you don't want to raise a TypeError:
```
```

pavlopro (changes requested):

app/main.py
Alex-Plugin: Thank you for your comments. Here are my corrections if I've got it right.

Dimosphen1 (approved):
LGTM, well done!

PR #1266, opened 2024-06-23

vldyu:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, float | int):
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | float | int) -> Distance:
```
Dimosphen1, line 27: Since you indicate in the type annotation that `other` could be `Distance` class instance as well, either add it's processing in the method or remove from the type annotation. This relates to the `__truediv__` method as well

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1270, opened 2024-06-24
Solution

Bielichenko:


app/main.py
```diff
@@ -1,3 +1,72 @@
...
+
+        operations = {
+            "lt": curr_inst_value < comp_value,
+            "gt": curr_inst_value > comp_value,
+            "eq": curr_inst_value == comp_value,
+            "le": curr_inst_value <= comp_value,
+            "ge": curr_inst_value >= comp_value,
+        }
+
+        return operations.get(operation_type)
```
LiudmylaKulinchenko, line 57: It will take 15 lines less if you just write this logic in each method instead 👀
In le, ge methods you can also use previously created methods. Hint: <= is the same as not >

Don't forget you can use operand `<` instead of `self.__lt__(other)` etc.

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+        return Distance.compare_objs(self, obj, "gt")
+
+    def __eq__(self, obj: Distance | int | float) -> bool:
+        return Distance.compare_objs(self, obj, "eq")
+
+    def __le__(self, obj: Distance | int | float) -> bool:
+        return Distance.compare_objs(self, obj, "le")
+
+    def __ge__(self, obj: Distance | int | float) -> bool:
+        return Distance.compare_objs(self, obj, "ge")
```
LiudmylaKulinchenko, line 72: Use full names for variables, you can use `other` instead, for example. Not really descriptive also, but will be better than object :)

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __eq__(self, other_el: Distance | int | float) -> bool:
+        if isinstance(other_el, Distance):
+            return self.km == other_el.km
+
+        return self.km == other_el
+
+    def __le__(self, other_el: Distance | int | float) -> bool:
+        if isinstance(other_el, Distance):
+            return self.km <= other_el.km
```
LiudmylaKulinchenko, line 54: You can use previously created `__gt__` method to make it in one line. Don't forget you can use `>` instead of directly calling a method

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+        return Distance(self.km + addends.km)
+
+    def __iadd__(self, addends: Distance | int | float) -> Distance:
+        if isinstance(addends, int | float):
+            self.km = self.km + addends
+            return self
+
+        self.km = self.km + addends.km
+        return self
```
LiudmylaKulinchenko, line 26: try to avoid duplicating return here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other_el, Distance):
+            return self.km == other_el.km
+
+        return self.km == other_el
+
+    def __le__(self, other_el: Distance | int | float) -> bool:
+        if isinstance(other_el, Distance):
+            return self.km <= other_el.km
+
+        return self.km <= other_el
```
LiudmylaKulinchenko, line 53: You can use `not >` instead to make it one line

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __ge__(self, other_el: Distance | int | float) -> bool:
+        return self > other_el or self == other_el
```
LiudmylaKulinchenko, line 56: same here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🌷

PR #1273, opened 2024-06-27
Implement Magic Methods

Morphin20th:


app/main.py
```diff
@@ -1,3 +1,57 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int or float) -> None:
```
Dimosphen1, line 2: Use `|` insred of `or` in the type annotation here and in the other places as well

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km!r})"
+
+    def __add__(self, other: int or "Distance") -> "Distance":
```
Dimosphen1, line 11: Indicate that other could be `float` as well. Do not use `""` in the type annotation 
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```
In case you face errors, use `from __future__ import annotations` import

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(
+                self.km + other
+            )
```
Dimosphen1, line 15: There is no need to place `self.km + other` on a separate line, avoid line breaking in such cases

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return Distance(
+            self.km + other.km
+        )
+
+    def __iadd__(self, other: int or "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
Dimosphen1, line 25: Perform type checking before conducting this and other similar operations as well

Dimosphen1 (changes requested):

app/main.py
Morphin20th: according to tests method __ mul __ and __ truediv __ cannot receive Distance. So is there any need to check?

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __iadd__(self, other: int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        self.km *= other
+        return self
```
pavlopro, line 28: Should return new instance

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km!r})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: int | Distance) -> Distance:
```
pavlopro, line 19: Include float to annotations as well and in methods below

pavlopro (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #1279, opened 2024-07-02
Solution

NazarSerdiuk1:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" or float or int) -> "Distance":
```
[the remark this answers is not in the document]
Dimosphen1: Type annotation is not fixed. Use `from __future__ import annotations` import in case you face errors

app/main.py
```diff
@@ -1,3 +1,70 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float | int) -> None:
+        self.km = float(km)
+
+    def __str__(self) -> str:
+        km_str = int(self.km) if self.km.is_integer() else self.km
```
Dimosphen1, line 9: This line is redundant, you could use `self.km` directly. Change the implementation of `__str__` and `__repr__` accordingly

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,76 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float | int) -> None:
+        self.km = float(km)
```
Dimosphen1, line 6: Do not convert to float value here
```suggestion
        self.km = km
```

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __str__(self) -> str:
+        return (
+            f"Distance: {int(self.km) if self.km.is_integer() else self.km} "
+            "kilometers."
+        )
```
Dimosphen1, line 12: Do not use if condition here and in the `__repr__` as well
```suggestion
        return f"Distance: {self.km} kilometers."
```

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1280, opened 2024-07-02
Solved

Krasul1a:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float or int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, distance2: Distance or float or int) -> Distance:
```
EdAlekseiev, line 14: You can use `or` for type hint, you should use `|`. 
https://docs.python.org/3/library/typing.html#typing.Union
Fix everywhere
```suggestion
    def __add__(self, distance2: Distance | float | int) -> Distance:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1281, opened 2024-07-03
'Solution'

dryzhenko:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> "Distance":
```
EdAlekseiev, line 11: `other` can be `int` and `Distance`, you should show it in type hint. Fix everywhere
```suggestion
    def __add__(self, other: float | int | Distance) -> "Distance":
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1282, opened 2024-07-04
solution

vpurshega:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: float) -> float:
```
vasylhnatiuk, line 11: other can be int and Distance instance as well

vasylhnatiuk: fix everywhere

vasylhnatiuk (changes requested):

PR #1283, opened 2024-07-04
Solution

PavelTym:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
```
vsmutok, line 61:
```suggestion
        return self.km <= other
```
`else` redundant in this case 
please do it everywhere

vsmutok (changes requested):

PR #1287, opened 2024-07-09
solution

dvitalijb:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __eq__(self, other: int) -> bool:
+        return self.km == other
+
+    def __gt__(self, other: Distance | float) -> bool:
+        return (self.km > other.km) \
+            if isinstance(other, Distance) \
+            else self.km > other
```
NazarSheremet, line 40: This implementation of the function will work, but using a one-liner if, then splitting it doesn't look very good. It's better to do it like this:
```suggestion
def __gt__(self, other: Distance | float) -> bool:
    if isinstance(other, Distance):
        return self.km > other.km
    return self.km > other

```

NazarSheremet (changes requested):
Fix suggestions

YuriiSmolii (approved):
LGTM

PR #1291, opened 2024-07-12
Solution

elfredburger:


app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
```
NazarSheremet, line 19: In this case, it's better to use `else` or `elif` because the app won't check other cases if the first condition is true.
```suggestion
        if isinstance(other, (int, float)):
            return Distance(self.km + other)
        else:
            return Distance(self.km + other.km)
```

NazarSheremet (changes requested):
Need to implement suggestion to other functions

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+        else:
+            raise TypeError(f"Can't multiply {type(other)} with Distance")
```
NazarSheremet, line 33:
```suggestion
    def __mul__(self, other: Distance | int | float) -> Distance:
        if isinstance(other, (int, float)):
            return Distance(self.km * other)
        else:
            raise TypeError(f"Can't multiply {type(other)} with Distance")
```

NazarSheremet (approved):
Good job!

PR #1295, opened 2024-07-15
implement py-distance-class

dkn3r:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> object:
```
EdAlekseiev, line 14: `other` can be float and Distance too. Show it in type hint and fix everywhere
```suggestion
    def __add__(self, other: int | float | Distance) -> object:
```

EdAlekseiev (changes requested):

app/main.py
dkn3r: pls check my last commit

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[int, float, Distance]) -> Distance:
```

vasylhnatiuk: use `from __future__ import annotations`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1305, opened 2024-07-19
Develop

georgiy-nezbritskiy:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        return Distance(km=self.km + other.km) \
+            if isinstance(other, Distance) else Distance(km=self.km + other)
```
LiudmylaKulinchenko, line 18: It's better to avoid using `\` as a linebreaker. You can use regular conditions instead of ternary operator

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Any:
+        self.km += other.km if isinstance(other, Distance) else other
+        return self
```
LiudmylaKulinchenko, line 22: Are you sure `self` could  be any? Fix return annotation

app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km == other.km \
+            if isinstance(other, Distance) else self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= other.km \
+            if isinstance(other, Distance) else self.km <= other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= other.km \
+            if isinstance(other, Distance) else self.km >= other
```
LiudmylaKulinchenko, line 48: try to rewrite these two methods in one line each. Use previously created magic methods

LiudmylaKulinchenko (changes requested):
Fix comments throughout the code

app/main.py
georgiy-nezbritskiy: quite strange to see this suggestion and at the same time use comprehension

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self.__lt__(other)
```
LiudmylaKulinchenko, line 49: You can use requal operators instead of calling methods directly

LiudmylaKulinchenko (approved):

app/main.py
LiudmylaKulinchenko, line 49: You can use requal operators instead of calling methods directly

LiudmylaKulinchenko (changes requested):
Lo

LiudmylaKulinchenko (approved):
Looks great! 🌷

PR #1309, opened 2024-07-21
Solution

Atikiho:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return self.km > other
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.km == other
+
+    def __ge__(self, other: int | float) -> bool:
+        return self.km >= other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
```
AnyoneClown, line 47: What if `other` is of type `Distance` here and in all the following methods?

AnyoneClown (changes requested):

PR #1310, opened 2024-07-21
Resolve

ihorutkin:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return Distance(
+            km=self.km + other.km
+        )
+
+    def __iadd__(self, other: "Distance") -> "Distance":
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
```
MNDonut, line 18: add operation also supports integet but you didn't annotate it

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __le__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            return self.km >= other
+        return self.km >= other.km
```
MNDonut, line 55: comparison supports integer as well but you didn't annotate it

MNDonut (changes requested):

app/main.py
```diff
@@ -3,7 +3,7 @@ class Distance:
     def __init__(self, km: int) -> None:
         self.km = km
 
-    def __add__(self, other: "Distance") -> "Distance":
+    def __add__(self, other: "Distance | int | int") -> "Distance":
```
MNDonut, line 6: why double int?
also you can move int out of quotes

app/main.py
```diff
@@ -29,27 +29,27 @@ def __truediv__(self, other: "Distance") -> "Distance":
...
+    def __le__(self, other: "Distance | int") -> bool:
         if not isinstance(other, Distance):
             return self.km <= other
         return self.km <= other.km
 
-    def __ge__(self, other: "Distance") -> bool:
+    def __ge__(self, other: "Distance | int") -> bool:
         if not isinstance(other, Distance):
             return self.km >= other
         return self.km >= other.km
```
MNDonut, line 55: you can move int out of quotes

MNDonut (changes requested):

app/main.py
ihorutkin: Yeah, done

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if not isinstance(other, Distance):
+            return self.km >= other
+        return self.km >= other.km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
```
oliasymak24, line 63: Usually it is better to put __str__ and __repr__ methods to the begging

oliasymak24 (approved):
In general your code looks correct, small comment
P.s. usually to avoid diplicates better to create separate method, in your case I mean all these checks

PR #1314, opened 2024-07-22
Solution

mnezbrytska:


app/main.py
```diff
@@ -1,3 +1,62 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __add__(self, other: int | float) -> "Distance":
```
LiudmylaKulinchenko, line 5: If your function gets only int and float types, why do you check `if isinstance(other, Distance)` in the next line and cover this case? Fix annotations everywhere

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
LiudmylaKulinchenko, line 8: Will it work the same if you remove this line?

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __iadd__(self, other: int | float) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            return NotImplemented
```
LiudmylaKulinchenko, line 23: You cover this case only in this function, you should have similar logic in your code
You also should `raise` errors instead of `return`. It will be covered in the next lessons, so you can delete it or make a little reasearch, up to you in this case!

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
```
LiudmylaKulinchenko, line 62: In ge, le you can use previously created methods
Hint: >= is same as not <

LiudmylaKulinchenko (changes requested):

app/main.py
Oleksl888: Try to simplify it further - no need for `elif` here. If `other` is of a wrong type python will throw an exception

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        if isinstance(other, (int, float)):
```
Oleksl888, line 23: would make more sense to use `elif` or `else` here - because as of now both conditions are being checked every time even if first condition already was executed

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
Oleksl888, line 27: multiplication and division do not expect `Distance` as `other`. Besides your code will now break if `Distance` is being passed: `self.km * other` since multiplication distance*distance has not been defined. 
Long story short - simplify `mul & truediv` methods to single return statement, no need to check for `isinstance` and use correct types in function parameters

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, (Distance | int | float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (Distance | int | float)):
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
```
Oleksl888, line 36: try to rewrite is following same recommendations as for `__add__` method. Let it fit in 3 lines with one if statement

Oleksl888 (changes requested):

app/main.py
Oleksl888: You need to fix type annotation for `other` in this method and in `__true_div__`

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1317, opened 2024-07-23

AndriiHamasa:


app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            self.km * num
+        )
+
+    def __truediv__(self, num: Union[Distance, float, int]) -> Distance:
+        return Distance(
+            round(self.km / num, 2)
+        )
+
+    def __lt__(self, other: Union[Distance, float, int]) -> bool:
+        km_of_other = other if not isinstance(other, Distance) else other.km
```
MNDonut, line 37: Instead of Union you can use |

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        return Distance(
+            self.km + km_of_other
+        )
+
+    def __iadd__(self, other: Union[Distance, float, int]) -> Distance:
+        km_of_other = other if not isinstance(other, Distance) else other.km
+        self.km += km_of_other
+        return self
+
```
MNDonut, line 25: it's better to use `other.km if isinstance`(remove not condition and simplify if statement). fix everywhere

MNDonut (changes requested):

MNDonut (approved):

PR #1318, opened 2024-07-23
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1,3 +1,46 @@
+from __future__ import annotations
+from functools import total_ordering
+
+
+@total_ordering
```
MNDonut, line 5: Check the note below(it may slow down your program) This task is to learn how to override methods and which magic methods exist. So let's implement them by yourself ;)

MNDonut: image

MNDonut (changes requested):

MNDonut (approved):

PR #1319, opened 2024-07-23
Distance class

yashkunn:


app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= float(other)
+        return NotImplemented
+
+    def __ge__(self, other: (int, float)) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 69: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == float(other)
+        return NotImplemented
+
+    def __le__(self, other: (int, float)) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 62: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > float(other)
+        return NotImplemented
+
+    def __eq__(self, other: (int, float)) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 55: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < float(other)
+        return NotImplemented
+
+    def __gt__(self, other: (int, float)) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 48: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+
+    def __truediv__(self, other: (int, float)) -> "Distance":
+        if isinstance(other, (int, float)) and other != 0:
+            return Distance(round(self.km / float(other), 2))
+        return NotImplemented
+
+    def __lt__(self, other: (int, float)) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 41: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + float(other))
+        return NotImplemented
+
+    def __iadd__(self, other: (int, float)) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 22: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float)) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
oliasymak24, line 15: You have a lot of diplicate code, try to make separate method for checking

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+            return NotImplemented
+        return self
+
+    def __mul__(self, other: (int, float)) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * float(other))
+        return NotImplemented
+
+    def __truediv__(self, other: (int, float)) -> "Distance":
+        if isinstance(other, (int, float)) and other != 0:
```
oliasymak24, line 34: Try to divide this check to two separate, we have some special case for raising errors when we have dividing to zero

oliasymak24 (changes requested):
Some small comments

app/main.py
```diff
@@ -1,3 +1,49 @@
+def _convert_to_km(other: (int, float)) -> float:
```
Oleksl888, line 1: If you use such a helper it would make sense to have it as a method so it is bound to the class

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __init__(self, km: float = 0.0) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float)) -> "Distance":
```
Oleksl888, line 20: 1. Use annotations from `__future__` module so you can use class name as a type hint inside of the class
2. Typing of (int, float) is misleading because it implies that other should be a tuple of int and float, correct way is to use Union[int, float] or use pipe int | float.
3. other can also be of Distance type.
4. Please check and fix in all methods where applicable

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+        self.km += _convert_to_km(other)
+        return self
+
+    def __mul__(self, other: (int, float)) -> "Distance":
+        return Distance(self.km * _convert_to_km(float(other)))
+
+    def __truediv__(self, other: (int, float)) -> "Distance":
+        other_km = _convert_to_km(other)
+        if other_km == 0:
+            raise ValueError("Cannot divide by zero")
```
Oleksl888, line 33: There is a more specific error type for a case when you divide by zero. It will be triggered anyway so raising a ValueError instead with a custom message will not bring anything to the table

app/main.py
```diff
@@ -1,3 +1,49 @@
...
+    def __lt__(self, other: (int, float)) -> bool:
+        return self.km < _convert_to_km(other)
+
+    def __gt__(self, other: (int, float)) -> bool:
+        return self.km > _convert_to_km(other)
+
+    def __eq__(self, other: (int, float)) -> bool:
+        return self.km == _convert_to_km(other)
+
+    def __le__(self, other: (int, float)) -> bool:
```
Oleksl888, line 45: Can you simplify your `__le__ and __ge__` methods? Hint: you already have other methods specified so `>=` will be the opposite to `<`

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    @staticmethod
+    def _convert_to_km(other: Union[int, float, Distance]) -> float:
+        if isinstance(other, Distance):
+            return other.km
+        elif isinstance(other, Union[int, float, Distance]):
+            return float(other)
+        raise TypeError("Unsupported type for arithmetic operation")
+
+    def __add__(self, other: Union[int, float, Distance]) -> "Distance":
```
LiudmylaKulinchenko, line 24: Does `|` works intead of Union with your python version? Your code will look more clean if you'll use it

LiudmylaKulinchenko, line 24: Why do u use both Distance and "Distance" version in your annotations? Remove quotes

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return Distance(self.km + self._convert_to_km(other))
+
+    def __iadd__(self, other: Union[int, float, Distance]) -> "Distance":
+        self.km += self._convert_to_km(other)
+        return self
+
+    def __mul__(self, other: Union[int, float, Distance]) -> "Distance":
+        return Distance(self.km * self._convert_to_km(float(other)))
+
+    def __truediv__(self, other: (int, float)) -> "Distance":
```
LiudmylaKulinchenko, line 34: `Typing of (int, float) is misleading because it implies that other should be a tuple of int and float, correct way is to use Union[int, float] or use pipe int | float.` -- you didn't fix it here

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            raise TypeError("Unsupported type for arithmetic operation")
+        other_km = self._convert_to_km(other)
+        if other_km == 0:
+            raise ZeroDivisionError("Cannot divide by zero")
+        return Distance(round(self.km / other_km, 2))
```
LiudmylaKulinchenko, line 40: I think, you'll get this error anyway in the 40th line if other_km will be 0, so there's no need to raise it

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @staticmethod
+    def _convert_to_km(other: Union[int, float, Distance]) -> float:
+        if isinstance(other, Distance):
+            return other.km
+        elif isinstance(other, Union[int, float, Distance]):
```
LiudmylaKulinchenko, line 20: You already cover `Distance` type in this function. Not any Distance type will go to this line, since you cover in in 18th line. You can remove it here

LiudmylaKulinchenko, line 20: Actually, only here it's ok to use quotes. Union will be good for annotations, not inside a function

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):
Good job!

PR #1321, opened 2024-07-24
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1,3 +1,105 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __add__(self, other: "Distance") -> "Distance":
```
oliasymak24, line 5: Usually __add__ method better to put after __str__ and __repr__

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, divisor: int | float) -> "Distance":
+        return Distance(round(self.km / divisor, 2))
+
+    def __eq__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            if isinstance(other, int | float):
+                return self.km == other
+            raise TypeError(
```
oliasymak24, line 45: You have a lot of duplicate code for checking, try to put is in separate method.

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+                f"{type(other)}"
+            )
+
+        return self.km == other.km
+
+    def __ne__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            if isinstance(other, int | float):
+                return self.km != other
+            raise TypeError(
```
oliasymak24, line 56: You have a lot of duplicate code for checking, try to put is in separate method.

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+                f"{type(other)}"
+            )
+
+        return self.km != other.km
+
+    def __gt__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            if isinstance(other, int | float):
+                return self.km > other
+            raise TypeError(
```
oliasymak24, line 67: You have a lot of duplicate code for checking, try to put is in separate method.

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+                f"{type(other)}"
+            )
+
+        return self.km >= other.km
+
+    def __lt__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            if isinstance(other, int | float):
+                return self.km < other
+            raise TypeError(
```
oliasymak24, line 89: You have a lot of duplicate code for checking, try to put is in separate method.

app/main.py
```diff
@@ -1,3 +1,105 @@
...
+                f"{type(other)}"
+            )
+
+        return self.km < other.km
+
+    def __le__(self, other: "Distance") -> bool:
+        if not isinstance(other, Distance):
+            if isinstance(other, int | float):
+                return self.km <= other
+            raise TypeError(
```
oliasymak24, line 100: You have a lot of duplicate code for checking, try to put is in separate method.

app/main.py
```diff
@@ -1,3 +1,105 @@
...
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if not isinstance(other, Distance):
+            if isinstance(other, int | float):
+                return Distance(self.km + other)
+            raise TypeError(
```
oliasymak24, line 9: You have a lot of duplicate code for checking, try to put is in separate method.

oliasymak24 (changes requested):
Pay attantion to come commnets

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
LiudmylaKulinchenko, line 14: Why do you need Union here? You use `|` in other functions, you can use same everywhere

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(round(self.km / divisor, 2))
+
+    def __eq__(self, other: Union["Distance", int, float]) -> bool:
+        if not isinstance(other, Distance):
+            return self.km == other
+
+        return self.km == other.km
+
+    def __ne__(self, other: "Distance") -> bool:
+        return not self.__eq__(other)
```
LiudmylaKulinchenko, line 41: You can use operators instead of calling methods directly: `=` instead of `__eq__`

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return self.km > other.km
+
+    def __ge__(self, other: "Distance") -> bool:
+        return self.__gt__(other) or self.__eq__(other)
+
+    def __lt__(self, other: "Distance") -> bool:
+        return not self.__ge__(other)
+
+    def __le__(self, other: "Distance") -> bool:
+        return self.__lt__(other) or self.__eq__(other)
```
LiudmylaKulinchenko, line 56: These methods shouldn't only get distance objects. They should work for int and float as well, same as gt, eq, ect. So fix annotations

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __mul__(self, multiplier: int | float) -> "Distance":
+        return Distance(self.km * multiplier)
+
+    def __truediv__(self, divisor: int | float) -> "Distance":
+        return Distance(round(self.km / divisor, 2))
+
+    def __eq__(self, other: Union["Distance", int, float]) -> bool:
+        if not isinstance(other, Distance):
+            return self.km == other
```
LiudmylaKulinchenko, line 36: Why don't u just remove `not` from this check and change places lines 36 and 38? It will make code more readable and easier to understand

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return not self.__eq__(other)
+
+    def __gt__(self, other: Union["Distance", int, float]) -> bool:
+        if not isinstance(other, Distance):
+            return self.km > other
+
+        return self.km > other.km
+
+    def __ge__(self, other: "Distance") -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
LiudmylaKulinchenko, line 50: `>=` is same as `not <`

LiudmylaKulinchenko (changes requested):

app/main.py
Bogdan-Rylov: In the `__ge__` method, I used `self > other or self == other` to avoid the `__ge__` and `__lt__` methods calling each other endlessly

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: "Distance" | int | float) -> "Distance":
```
Oleksl888, line 14: no need to use quotes on class name here and everywhere since you have import from futures

Oleksl888 (changes requested):

oliasymak24 (approved):
Godd job!

PR #1324, opened 2024-07-24
'Solution'

Ivan-Shakhman:
Solution of task

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,56 @@
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (int, float)) -> object:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            raise TypeError(f"Cannot add {type(other)} to Distance")
+
+    def __iadd__(self, other: (int, float)) -> object:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError(f"Cannot add {type(other)} to Distance")
+        return self
```
oliasymak24, line 27: use one style of code for return statement, in the __add__ fucntion you have different one

oliasymak24 (changes requested):
In general your code looks good, some small comment

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __add__(self, other: (int, float)) -> object:
+        if not (
+            isinstance(other, Distance)
+            or isinstance(other, (int, float))
+        ):
+            raise TypeError(f"Cannot add {type(other)} to Distance")
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
```
oliasymak24, line 20: Use two same returns is not good.
Better to be to create an object for some variable and then return it) To keep one style
P.s. in previous comment I missed this statement  `self.km += other.km`, so you can use that likke you did before.

oliasymak24 (changes requested):
Small comment

oliasymak24 (approved):
Good job!

PR #1328, opened 2024-07-24
Solution

dpiuro:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
MNDonut, line 14: you can use | instead of Union

MNDonut (changes requested):

app/main.py
```diff
@@ -11,14 +8,14 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"Distance(km={self.km})"
 
-    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
+    def __add__(self, other: "Distance") -> "Distance":
```
MNDonut, line 11: why did you remove int and float?

app/main.py
```diff
@@ -11,14 +8,14 @@ def __str__(self) -> str:
...
         if isinstance(other, Distance):
             return Distance(self.km + other.km)
         elif isinstance(other, (int, float)):
             return Distance(self.km + other)
         return NotImplemented
 
-    def __iadd__(self, other: Union["Distance", int, float]) -> "Distance":
+    def __iadd__(self, other: "Distance") -> "Distance":
```
MNDonut, line 18: why did you remove int and float?

app/main.py
```diff
@@ -27,45 +24,45 @@ def __iadd__(self, other: Union["Distance", int, float]) -> "Distance":
...
         return NotImplemented
 
-    def __truediv__(self, other: Union[int, float]) -> "Distance":
+    def __truediv__(self, other: int | float) -> "Distance":
         if isinstance(other, (int, float)):
             return Distance(round(self.km / other, 2))
         return NotImplemented
 
-    def __lt__(self, other: Union["Distance", int, float]) -> bool:
+    def __lt__(self, other: "Distance") -> bool:
```
MNDonut, line 37: why did you remove int and float?

app/main.py
```diff
@@ -27,45 +24,45 @@ def __iadd__(self, other: Union["Distance", int, float]) -> "Distance":
...
         if isinstance(other, Distance):
             return self.km < other.km
         elif isinstance(other, (int, float)):
             return self.km < other
         return NotImplemented
 
-    def __gt__(self, other: Union["Distance", int, float]) -> bool:
+    def __gt__(self, other: "Distance") -> bool:
```
MNDonut, line 44: why did you remove int and float?
fix everywhere

MNDonut (changes requested):

app/main.py
```diff
@@ -1,3 +1,74 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" | int | float) -> "Distance":
```
Oleksl888, line 15: You don't need to put class name in quotes since you imported future annotations

app/main.py
```diff
@@ -1,3 +1,74 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return NotImplemented
```
Oleksl888, line 20: Not  really necessary to return NotImplemented here and in other methods. If you don't use it then it can simplify your code and make it more readable and contain less instructions. For example:
```
    def __add__(self, other: Distance | int) -> Distance:
Here you have only three lines of code and only one if condition. And then if data type is incorrect and adding could not happen an exception will be raised which is perfectly fine in this scenario

File: app/main.py [hunk 8]
```diff
@@ -1,3 +1,74 @@
...
+
+    def __eq__(self, other: "Distance" | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        return NotImplemented
+
+    def __le__(self, other: "Distance" | int | float) -> bool:
```
Oleksl888, line 62: Can you simplify your `__le__` and `__ge__` methods? Hint: you already have other methods specified so `>=` will be the opposite to `<`

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
```
oliasymak24, line 29: You have two same returns

oliasymak24 (changes requested):

app/main.py
```diff
@@ -27,12 +27,12 @@ def __iadd__(self, other: Distance | int | float) -> Distance:
...
         if isinstance(other, (int, float)):
             return Distance(self.km * other)
-        return Distance(self.km * other)
+        raise "Error"
 
     def __truediv__(self, other: int | float) -> Distance:
         if isinstance(other, (int, float)):
             return Distance(round(self.km / other, 2))
-        return Distance(round(self.km / other, 2))
+        raise "Error"
```
MNDonut, line 35: You'll get error: "Exception should inherit from BaseException class" if it reaches raise statement(after raise there should be a class that inherits from BaseException class). Let's raise NotImplementedError

MNDonut (changes requested):

MNDonut (approved):

PR #1329, opened 2024-07-24
Solution

pashawarganov:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
MNDonut, line 31: what if other = 0?

MNDonut (changes requested):

app/main.py
pashawarganov: > what if other = 0?

OK, I understand. What should I return? Throw some kind of error, print something, or maybe return zero?

MNDonut (approved):

PR #1330, opened 2024-07-24
task done.

Sergi0bbb:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return None
+
+    def __iadd__(self, other: int) -> int:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            return None
+        return self
```
MNDonut, line 26: you annotated other as int, but process Distance as well

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
+            return None
+
+    def comparison(self, other: int) -> bool:
+        if isinstance(other, Distance):
+            return self.km - other.km
+        elif isinstance(other, (int, float)):
+            return self.km - other
+        else:
+            return None
```
MNDonut, line 47: you annotated other as int, but process Distance as well

MNDonut (changes requested):

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
Oleksl888, line 17: It's not a good design to simply return `None` when user will try to add unacceptable data types - it should raise an exception to signal something went wrong. You may create custom exceptions to process this error but in this task is better just to simplify this method to have three lines of code - one if statement and two return statements.
Check your other methods where you return `None` and change accordingly

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+            return Distance(self.km + other)
+        else:
+            return None
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
```
Oleksl888, line 25: You can simplify this too so you don't have this `elif` statement

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def comparison(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km - other.km
+        elif isinstance(other, (int, float)):
+            return self.km - other
+        else:
+            return None
+
+    def __lt__(self, other: int) -> bool:
+        return self.comparison(other) < 0
```
Oleksl888, line 53: This is not very obvious implementation of comparison methods. Why do this math instead of comparing values directly?

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __gt__(self, other: int) -> bool:
```
Oleksl888, line 55: `other` can be of other types as well

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            return None
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
```
Oleksl888, line 31: `mul and truediv` methods assume you multiply or divide only by numeric type not Distance type

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    def __lt__(self, other: int) -> bool:
+        return self.comparison(other) < 0
+
+    def __gt__(self, other: int) -> bool:
+        return self.comparison(other) > 0
+
+    def __eq__(self, other: int) -> bool:
+        return self.comparison(other) == 0
+
+    def __le__(self, other: int) -> bool:
```
Oleksl888, line 61: Can you simplify your `__le__ and __ge__` methods? Hint: you already have other methods specified so `>=` will be the opposite to `<`

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        raise TypeError
```
Oleksl888, line 29: It is a bad practice to raise exception without an error message.
Also here you can simplify this to one line so you don't have to check  type of `other.  Perform multiplication with whatever comes in `other`,  exception will be raised by python if it is not a numeric type and it will not be able to perform multiplication. You can do the same in `truediv` method

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            result = self.km / other
+            return Distance(round(result, 2))
```
Oleksl888, line 34: No need to store `result` in variable - you can do all operations  in one line

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        return self.comparison(other) < 0
+
+    def __gt__(self, other: int | float) -> bool:
+        return self.comparison(other) > 0
+
+    def __eq__(self, other: int | float) -> bool:
+        return self.comparison(other) == 0
+
+    def __le__(self, other: int | float) -> bool:
+        return not self > other
```
Oleksl888, line 52: Well done!

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            result = self.km / other
+            return Distance(round(result, 2))
+        raise TypeError
+
+    def comparison(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return self.km - other.km
+        return self.km - other
+
+    def __lt__(self, other: int | float) -> bool:
```
Oleksl888, line 42: `other` can be of a Distance type as well

Oleksl888 (changes requested):

LiudmylaKulinchenko (changes requested):
Review your annotations everywhere. Check all comment in the pr again and make sure to fix everywhere throughout the code. You should always do so

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
Oleksl888, line 47: You could have left previous implementation of your methods:
`return not self > other`
`return not self < other`

Oleksl888 (approved):
You made it! :+1:

PR #1333, opened 2024-07-25
Task 7

Poznanskyi:


app/main.py
```diff
@@ -1,3 +1,79 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> None:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> None:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
MNDonut, line 12: but other can be int or float as well
fix everywhere

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
MNDonut, line 17: you don't need else because if and elif have return.
fix everywhere

MNDonut (changes requested):

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
-    def __str__(self) -> None:
+    def __str__(self) -> str:
         return f"Distance: {self.km} kilometers."
 
-    def __repr__(self) -> None:
+    def __repr__(self) -> str:
         return f"Distance(km={self.km})"
 
-    def __add__(self, other: "Distance") -> "Distance":
+    def __add__(self, other: float) -> "Distance":
```
MNDonut, line 12: but other can be Distance and int as well because you do these checks below

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
         if isinstance(other, Distance):
             return Distance(self.km + other.km)
         elif isinstance(other, (int, float)):
             return Distance(self.km + other)
-        else:
-            return NotImplemented
+        return NotImplemented
 
-    def __iadd__(self, other: "Distance") -> "Distance":
+    def __iadd__(self, other: float) -> "Distance":
```
MNDonut, line 19: but other can be Distance and int as well because you do these checks below

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
         if isinstance(other, Distance):
             self.km += other.km
         elif isinstance(other, (int, float)):
             self.km += other
-        else:
-            return NotImplemented
         return self
 
     def __mul__(self, factor: float) -> "Distance":
```
MNDonut, line 26: but other can be int as well because you do these checks below

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
         if isinstance(factor, (int, float)):
             return Distance(self.km * factor)
-        else:
-            return NotImplemented
+        return NotImplemented
 
     def __truediv__(self, divisor: float) -> "Distance":
```
MNDonut, line 31: but other can be int as well because you do these checks below

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
         if isinstance(divisor, (int, float)):
             return Distance(round(self.km / divisor, 2))
-        else:
-            return NotImplemented
+        return NotImplemented
 
     def __lt__(self, other: "Distance") -> bool:
```
MNDonut, line 36: but other can be int and float as well because you do these checks below

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
     def __eq__(self, other: "Distance") -> bool:
         if isinstance(other, Distance):
             return self.km == other.km
         elif isinstance(other, (int, float)):
             return self.km == other
-        else:
-            return NotImplemented
+        return NotImplemented
 
     def __le__(self, other: "Distance") -> bool:
```
MNDonut, line 57: but other can be int and float as well because you do these checks below

app/main.py
```diff
@@ -3,77 +3,67 @@ class Distance:
...
         if isinstance(other, Distance):
             return self.km <= other.km
         elif isinstance(other, (int, float)):
             return self.km <= other
-        else:
-            return NotImplemented
+        return NotImplemented
 
     def __ge__(self, other: "Distance") -> bool:
```
MNDonut, line 64: but other can be int and float as well because you do these checks below

MNDonut (changes requested):

MNDonut (changes requested):
You didn't push a new commit after the comments I left

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
MNDonut, line 15: 1. You can use | instead of Union
2. No need to use quotes, import annotations from future and everything will work

MNDonut (changes requested):

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> "Distance":
```
oliasymak24, line 15: No need to use quotes, check in all places

app/main.py
```diff
@@ -1,3 +1,72 @@
...
+        if isinstance(other, Distance):
```
oliasymak24, line 16: For better optimization you can try to create a separate method for your code duplicates in all function for checking

oliasymak24 (changes requested):

oliasymak24 (approved):
Good job!

PR #1334, opened 2024-07-25
Solution

Anfibik:


app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
```
vsmutok, line 26:
```suggestion
    def __iadd__(self, other: Distance | int | float) -> Distance:
        if isinstance(other, Distance):
            self.km += other.km
        if isinstance(other, (int, float)):
            self.km += other
        return self
```

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if type(other) is Distance:
```
vsmutok, line 15: better to use `isinstance `

vsmutok (changes requested):

app/main.py
Anfibik: Explain please, why I need to use this function.

vasylhnatiuk (approved):

vasylhnatiuk (changes requested):
sorry, fix your tests

vsmutok (approved):
Great!

PR #1336, opened 2024-07-27
add magic methods

egor250330:


app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if not isinstance(other, Distance):
+            return Distance(km=self.km + other)
+        return Distance(
+            km=self.km + other.km
+        )
+
+    def __iadd__(self, other: Distance) -> Distance:
+        if not isinstance(other, Distance):
```
MNDonut, line 22: you annotated other as Distance but process case when other is not Distance D:

MNDonut (changes requested):

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        )
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: Distance | int) -> Distance:
```
LiudmylaKulinchenko, line 28: How about float?

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if not isinstance(other, Distance):
+            return Distance(
+                km=self.km * other
+            )
```
LiudmylaKulinchenko, line 32: No need to split in lines if it fits in one

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        return self.km == other.km
+
+    def __le__(self, other: Distance | int) -> bool:
+        if not isinstance(other, Distance):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if not isinstance(other, Distance):
+            return self.km >= other
```
LiudmylaKulinchenko, line 62: U can use previously created methods here
Hint: `>=` is the same as `not <`

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1346, opened 2024-08-01
Solution

alsagir:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        else:
+            return self.km <= other.km
+
+    def __ge__(self, other: [int, float, "Distance"]) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        else:
+            return self.km >= other.km
```
vsmutok, line 59:
```suggestion
    def __ge__(self, other: int | float | Distance) -> bool:
        if isinstance(other, (int, float)):
            return self.km >= other
        return self.km >= other.km
```

1. fix the annotation
2. `else` redundant in this case

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __iadd__(self, other: [int, float, Distance]) -> Distance:
```
vsmutok, line 21:
```suggestion
    def __iadd__(self, other: int | float | Distance) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, (int, float)):
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: [int, float]) -> Distance:
```
vsmutok, line 31:
```suggestion
    def __truediv__(self, other: int| float) -> Distance:
```

vsmutok (changes requested):

PR #1350, opened 2024-08-05
Solution

Serhii-Chubur:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> object:
```
EdAlekseiev, line 11: `other` can be int, float and Distance. Fix type hint everywhere
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

EdAlekseiev (changes requested):

app/main.py
Serhii-Chubur: def __add__(self, other: int | float | Distance) -> Distance:

E   NameError: name 'Distance' is not defined

Why does the error occur?    (Python 3.12.2)

vasylhnatiuk (approved):

PR #1352, opened 2024-08-07
Solution

bdnnk:


app/main.py
```diff
@@ -1,3 +1,54 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
Dimosphen1, line 11: Indicate that `other` could be `int` or `float` in the type annotation as well using `|`. Do not use `"` in the type annotation, indicate class directly. Apply similar changes in the other methods as well. In case you face errors, use `from __future__ import annotations` import

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(km=self.km * other)
+
+    def __truediv__(self, other: "Distance") -> "Distance":
```
Dimosphen1, line 27: It does not seem that this method expects `Distance` class instance since it is not processing it

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: "Distance") -> "Distance":
```
Dimosphen1, line 31: Use correct return type annotation

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1362, opened 2024-08-14
'Solution'

KrenkySpenky:


app/main.py
```diff
@@ -1,3 +1,68 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
```

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
```
Dimosphen1, line 11: Be more specific in the type annotation indicating that `other` could be `int`, `float`, or `Distance` class instance using `|`. Besides that, do not place the class name inside `""`, use `from __future__ import annotations` import in case you face errors. Revise type annotation in all methods

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        return NotImplemented
+
+    def __lt__(self, other: object) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        return NotImplemented
```
Dimosphen1, line 40: `NotImplemented` should be not returned but rather raised

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, other: float) -> Distance:
```
vsmutok, line 30:
```suggestion
    def __mul__(self, other: float | int) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,73 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```
vsmutok, line 6:
```suggestion
    def __init__(self, km: float | int) -> None:
        self.km = km
```

vsmutok (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1373, opened 2024-08-22
Solution

Deadmarvald:


app/main.py
```diff
@@ -1,3 +1,41 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+        self.real = km
```
YuriiSmolii, line 7: What is self.real?

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
YuriiSmolii, line 15: other can be int, float and Distance. Fix type hint everywhere

YuriiSmolii (changes requested):
You should use isinstance() method to check type of 'other' variable to choose right way to manipulate with data.

For example:
```
            return self.km > other
        return self.km > other.km
```

NazarSheremet (approved):
Good job!

PR #1381, opened 2024-08-28
Solution

asynyshyn:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: ["Distance", int, float]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
vsmutok, line 61:
```suggestion
        return self.km >= other
```
`else` redundant in this case
please fix everywhere

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ["Distance", int]) -> "Distance":
```
```suggestion
    def __add__(self, other: [Distance, int, float]) -> Distance:
```
please fix everywhere

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1382, opened 2024-08-28
Solution

MahdaArtem-py:
There is a question how to optimize the code. In tests where the second argument is a num, an error occurs: "AttributeError: 'int' object has no attribute 'km'". I got around this problem by adding instance variable .real. What is the best way to solve this problem?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,41 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+        self.real = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
+        return Distance(km=self.km + other.real)
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,41 @@
...
+        self.km = km
+        self.real = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
vsmutok, line 15:
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,41 @@
...
+        self.km = km
+        self.real = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
```
vasylhnatiuk, line 15: do `other: Distance | int | float` in each method

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1385, opened 2024-09-03
Solution

denys-badass:


.flake8
```diff
@@ -1,7 +1,7 @@
 [flake8]
 inline-quotes = "
 ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
-max-line-length = 79
+max-line-length = 119
```
NazarSheremet, line 4: It is generally recommended length of the line by PEP8, so no need to change it 
https://realpython.com/python-pep8/#maximum-line-length-and-line-breaking

NazarSheremet (changes requested):
Need to refactor code accoding to PEP8

.flake8
denys-badass: fixed

NazarSheremet (approved):
Good job!

PR #1392, opened 2024-09-09
Solution

Viacheslav417865:


app/main.py
```diff
@@ -1,3 +1,79 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
```
vasylhnatiuk, line 12: use more specific type instead of object

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union[Distance, int, float]) -> "Distance":
```
```suggestion
    def __add__(self, other: Union[Distance, int, float]) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+    def __iadd__(self, other: Union[Distance, int, float]) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError("Unsupported operand type for +=")
+        return self
+
+    def __mul__(self, other: object) -> Distance:
```
vasylhnatiuk, line 32: use more specific type instead of `object` everywhere!

vasylhnatiuk (changes requested):

PR #1393, opened 2024-09-11
'Solution'

KKvladys:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if not isinstance(other, Distance):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if not isinstance(other, Distance):
+            self.km += other
+            return self
+        self.km += other.km
+        return self
```
AnyoneClown, line 24: `return self` is duplicated, take it outside if-else

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, int) or isinstance(other, float):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if (isinstance(other, int) or isinstance(other, float)) and other != 0:
+            return Distance(round(self.km / other, 2))
```
AnyoneClown, line 32: `other` can not be of type `Distance` in this method

Also, for future reference, you can specify multiple types to check in one method.

For example: `if isinstance(other, (int, float)):`

app/main.py
```diff
@@ -19,16 +19,16 @@ def __add__(self, other: Distance | int | float) -> Distance:
...
 
     def __mul__(self, other: int | float) -> Distance:
-        if isinstance(other, int) or isinstance(other, float):
+        if isinstance(other, (int, float)):
             return Distance(self.km * other)
 
     def __truediv__(self, other: int | float) -> Distance:
-        if (isinstance(other, int) or isinstance(other, float)) and other != 0:
+        if isinstance(other, (int, float)) and other != 0:
             return Distance(round(self.km / other, 2))
```
AnyoneClown, line 32: You didn't fix it. `other` can not be of type `Distance` in this method, so you don't need to check for int/float types

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1395, opened 2024-09-13
Solution

andriiovod:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | int | float") -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

vasylhnatiuk: use `from __future__ import annotations`

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: change annotation!

vasylhnatiuk (changes requested):

PR #1397, opened 2024-09-14
distance-class

dmytrik:


app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
AnyoneClown, line 25: `return self` is duplicated, take it outside if-else

AnyoneClown (changes requested):

AnyoneClown (approved):
Good Job!

PR #1410, opened 2024-09-23
Solution-py-distance-class

skyfoxwork:


Mxbely (approved):
we can use a more easy syntax: `if isinstance(other, (int, float))`
and good job

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #1411, opened 2024-09-23
Change __iadd__ and completed task

Mxbely:


skyfoxwork (approved):
I think you can delete comments and you can use "return" without "else" and you can use "-> bool" when you method "return" bool, don't use "-> Distance"
like this:

    def __lt__(self, other: Distance) -> bool:
            if self.km < other.km:
                return True

            return False

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+            # return Distance(distance=self.km + other.real)
+            self.km += other.real
+        # return Distance(distance=self.km)
+        return self
+
+    def __mul__(self, other: Distance or int) -> Distance or None:
+        if isinstance(other, Distance):
+            return None
+        else:
+            return Distance(distance=self.km * other.real)
```
AnyoneClown, line 34: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km. You should always return a Distnace instance in this method

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+    def __truediv__(self, other: Distance or int) -> Distance or None:
+        if isinstance(other, Distance):
+            return None
+        else:
+            return Distance(distance=round(self.km / other.real, 2))
```
AnyoneClown, line 40: `other` can not be of type `Distance` in this method. You should always return a `Distnace` instance in this method

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+    def __eq__(self, other: Distance or int) -> Distance:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other.real
```
AnyoneClown, line 46: Inside of your annotations use logical operator `|`, it works like `or`. Also, all of this cases might have `float` type of km, you should add annotation to this too.

Refactor all functions yousing only 1 `return`. Check your `return` annotation, are you sure that you are returning `Distance` instance in all methods?

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+            return self.km >= other.real
+
+
+# distance1 = Distance(30)
+# distance2 = Distance(3)
+# distance3 = distance1 / distance2
+# print(distance1)
+# print(repr(distance3))
+# distance1 += distance2
+# print(distance1)
```
AnyoneClown, line 79: Delete all unused code before Pull Request

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other.real
+
+    def __ge__(self, other: Distance or int) -> Distance:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other.real
```
AnyoneClown, line 70: Try to refuctor using `not` operator where possible(Check all of this methods)

app/main.py
```diff
@@ -1,3 +1,79 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, distance: Distance) -> None:
+        self.km = distance
```
AnyoneClown, line 6: Inside of your `__init__` method you take the arguments that are needed to initialise your object's class, i.e. there must be `km` of type int | float

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance or int) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+        else:
+            self.km = self.km + other.real
+        return Distance(distance=self.km)
```
AnyoneClown, line 19: If we follow your logic, then every time we use `add` method, we will create a new instance. Return `self`, not a new class object

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+        else:
+            self.km = self.km + other.real
+        return Distance(distance=self.km)
```
Nattalli, line 19: you can return Distance in if statement and avoid `else` statement

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __lt__(self, other: int | float) -> bool:
+        return self.km < other
+
+    def __gt__(self, other: int | float) -> bool:
+        return self.km > other
+
+    def __le__(self, other: int | float) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: int | float) -> bool:
```
Nattalli, line 46: Знімок екрана 2024-09-26 о 11 54 54

these methods can have `other` value as a `Distance` class

Nattalli (changes requested):

app/main.py
```diff
@@ -31,17 +29,27 @@ def __mul__(self, other: int | float) -> Distance:
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other.real
 
-    def __ge__(self, other: int | float) -> bool:
-        return self.km >= other
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other.real
```
AnyoneClown, line 55: You didn't fix previous comment:

Refactor all functions using only 1 `return`. Operator `not` might help you in this case, use it.

AnyoneClown: Example:

```
return not self.km < other
```

AnyoneClown (changes requested):

AnyoneClown (approved):
Good, but you could not be lazy and implement the logic through reuse of methods and `not` operator :)

PR #1412, opened 2024-09-24
'Solution'

spa7id:


Mxbely (approved):
I think, you could not use `rise` in every method, and it would be easier to read. Please check my comments. Now i approve it, but i will be reject it at the next time.

app/main.py
```diff
@@ -1,3 +1,78 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
AnyoneClown, line 2: Add type hint for `float` too

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | int | float") -> "Distance":
+        if isinstance(other, Distance):
+            new_distance = self.km + other.km
+        elif isinstance(other, (int, float)):
+            new_distance = self.km + other
+        else:
+            raise TypeError
+        return Distance(new_distance)
```
AnyoneClown, line 18: 1. We don't need to raise Error in this case
2. Don't do 2 isinstance method in this case, just use `else`
3. Don't use `""` for your type hints, import `from __future__ import annotations`, if you need to use more advanced annotations and your python version is lower then you need

Good example:
`def __add__(self, other: Distance | int | float) -> Distance:`
Fix all of this in all methods

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        return self
+
+    def __mul__(self, other: int) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        raise TypeError
+
+    def __truediv__(self, other: int) -> "Distance":
+        if isinstance(other, Distance):
+            raise TypeError
```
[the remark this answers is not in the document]
AnyoneClown: Right, `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        else:
+            raise TypeError
+
+    def __ge__(self, other: "Distance | int | float") -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        else:
+            raise TypeError
```
AnyoneClown, line 78: Refactor all functions using only 1 `return`. Operator `not` might help you in this case, use it. Use `isinstance` only for checking `Distance`, and use `else` in other cases

AnyoneClown (changes requested):

PR #1414, opened 2024-09-25
Solution

honelyq:


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | float | int") -> "Distance":
```
pavlopro, line 12: `def __add__(self, other: Distance | int | float) -> Distance`

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
```
pavlopro, line 15: `else `operator is redundant

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __iadd__(self, other: "Distance | float | int") -> "Distance":
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        self.km += other.km
+        return self
+
+    def __mul__(self, other: "Distance | float | int") -> "Distance":
+        if isinstance(other, (Distance, int, float)):
```
pavlopro, line 26: no need this `if` here

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: "Distance | float | int") -> "Distance":
+        if isinstance(other, (Distance, int, float)) and other:
```
pavlopro, line 30: here as well

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+            return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: "Distance | float | int") -> bool:
+        return self.km < other
```
pavlopro, line 34: You should correct this operator and below as well to check if `other `will be a `Distance `object you should compare with a `km` attribute

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1415, opened 2024-09-25
Solution

VladyslavBon:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if type(other) == int or type(other) == float:
```
AnyoneClown, line 15: Use `isinstance()` method to check type of arguments. Check everywhere

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: Distance | int | float) -> Distance | None:
+        if type(other) == int or type(other) == float:
+            return Distance(self.km * other)
+        return None
```
AnyoneClown, line 29: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

You don't need to return `None` here

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+
+    def __truediv__(self, other: Distance | int | float) -> Distance | None:
+        if type(other) == int or type(other) == float:
+            return Distance(round(self.km / other, 2))
+        return None
```
AnyoneClown, line 34: `other` can not be of type `Distance` in this method

AnyoneClown (changes requested):

app/main.py
```diff
@@ -12,48 +12,46 @@ def __repr__(self) -> str:
...
-        if type(other) == int or type(other) == float:
+    def __mul__(self, other: int | float) -> Distance | None:
+        if isinstance(other, (int, float)):
             return Distance(self.km * other)
-        return None
 
-    def __truediv__(self, other: Distance | int | float) -> Distance | None:
-        if type(other) == int or type(other) == float:
+    def __truediv__(self, other: int | float) -> Distance | None:
+        if isinstance(other, (int, float)):
```
AnyoneClown, line 31: `other` can not be of type `Distance` in this methods, so we don't need to do a check for type of instance. Also fix return type annotatiohs

AnyoneClown (changes requested):

PR #1417, opened 2024-09-27
implement class Distance

94nj111:


dmytrik (approved):
good job!

app/main.py
```diff
@@ -1,3 +1,56 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
+        self.km = km
```
AnyoneClown, line 7: km can also be an `int`

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if type(other) is float or type(other) is int:
```
AnyoneClown, line 16: Read about method `isinstance()` and use it here. You can pass multiple types into 1 `isinstance` method to check type of your `object`.

AnyoneClown: Refactor everywhere

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if type(other) is float or type(other) is int:
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if type(other) is float or type(other) is int:
+            return self.km >= other
+        return self.km >= other.km
```
AnyoneClown, line 56: We can refactor this methods using `not` operator to avoid redundant code. Use `not` +  oposite method that you implemented erlier.

Example: `x < y` is the logical negation of `x <= y`, so we can refuctor using `not` + some method here

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job

PR #1418, opened 2024-09-28
Solution

vladyslav-tmf:


dmytrik (approved):

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        return self.km == other if isinstance(other, (int, float))\
+            else self.km == other.km
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return self.km <= other if isinstance(other, (int, float))\
+            else self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return self.km >= other if isinstance(other, (int, float))\
+            else self.km >= other.km
```
AnyoneClown, line 50: We can fix code style, so we don't need to use `\` here.

 ```suggestion
        return (
            self.km >= other 
            if isinstance(other, (int, float)) 
            else self.km >= other.km
        )
```

AnyoneClown: Fix in all cases

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job

PR #1422, opened 2024-09-28
implemented py-distance-class

yuliiapyrih:


app/main.py
```diff
@@ -1,3 +1,50 @@
...
-    # Write your code here
-    pass
+    def __init__(self, km: (float, int)) -> None:
+        self.km = km
+
+    def __add__(self, other: (int, float, "Distance")) -> "Distance":
+        km = other if isinstance(other, (int, float)) else other.km
+        return Distance(
+            km=self.km + km
+        )
```
LiudmylaKulinchenko, line 9: No need to split into lines if it fits in one

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        km = other if isinstance(other, (int, float)) else other.km
+        return self.km == km
+
+    def __le__(self, other: (int, float, "Distance")) -> bool:
+        km = other if isinstance(other, (int, float)) else other.km
+        return self.km <= km
+
+    def __ge__(self, other: (int, float, "Distance")) -> bool:
+        km = other if isinstance(other, (int, float)) else other.km
+        return self.km >= km
```
LiudmylaKulinchenko, line 50: try to rewrite these two methods in one line each. Use previously created magic methods

app/main.py
```diff
@@ -1,3 +1,50 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: (float, int)) -> None:
+        self.km = km
+
+    def __add__(self, other: (int, float, "Distance")) -> "Distance":
```
LiudmylaKulinchenko, line 5: You can use `|` operator for type annotations instead, it's prefered over using tuple-like parentheses

LiudmylaKulinchenko (changes requested):

app/main.py
yuliiapyrih: Done

app/main.py
yuliiapyrih: Done

app/main.py
yuliiapyrih: Done

Nattalli (approved):

PR #1423, opened 2024-09-29
Develop

sashasyrota:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
Dimosphen1, line 14: Please, indicate all possible data types for `other` here

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+        return Distance(self.km + other.km)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, int | float):
+            self.km += other
+            return self
+        self.km += other.km
+        return self
```
Dimosphen1, line 24: Try to not write two same `return` statements, rewrite this method to avoid it

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __mul__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, int | float):
+            return Distance(km=self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
```
Dimosphen1, line 30: In case you indicate that `other` could be `Distance` class instance, process it as well here and in the other methods

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, int | float):
```
pavlopro, line 15: `if isinstance(other, (int, float)):`

Correct it all other places below

pavlopro (changes requested):

app/main.py
sashasyrota: please explain me, why  **if isinstance(other, int | float)** not correct in this case

pavlopro (approved):
GJ!

PR #1428, opened 2024-09-30
distance-class_solution

OlhaStadnik:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        result = self + other
+        self.km = result.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance | int:
```
VladyslavBon, line 24: Return anotation here return only class, not int

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: int | float | Distance) -> Distance | int:
```
VladyslavBon, line 14: Return anotation here return only class, not int

VladyslavBon (approved):

app/main.py
AnyoneClown: +, fix it

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        result = self + other
+        self.km = result.km
+        return self
```
AnyoneClown, line 22: It's really cool that you were able to reuse the overridden `__add__` method here, well done!

app/main.py
AnyoneClown: +, fix it

app/main.py
```diff
@@ -1,3 +1,53 @@
...
+
+    def __mul__(self, other: int | float) -> Distance | int:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
```
AnyoneClown, line 27: `other` can also be `Distance` type. Add it to your type hint. Fix in all cases

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1430, opened 2024-09-30
'Solution'

Andrey-Dementyev:


spa7id (approved):

app/main.py
```diff
@@ -1,3 +1,54 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+
+    def __init__(self, km: float) -> None:
+        self.km = km
```
AnyoneClown, line 7: km can also be `int` here and in all other cases.

app/main.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __le__(self, other: float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
AnyoneClown, line 54: `other` can be `int` or `float` or `Distance` here

AnyoneClown (approved):
Good job, but fix some type hints

PR #1436, opened 2024-10-01
solution

VolodymyrSemchysyn:


YaroslavBordovoy (approved):
I would import annotations from __future__ for type annotations, no need to use quotes then. And reconsider the "other" parameter, int, float and Distance can come almost everywhere (__ge__, __le__, __eq__, __gt__, __lt__, __add__).

app/main.py
```diff
@@ -1,3 +1,87 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```

app/main.py
```diff
@@ -1,3 +1,87 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
AnyoneClown, line 11: Add more descriptive annotation to other. It can also be of type `int` or `float`.

Why are you using quotes for `"Distance"` everywhere? You can do it without them example: `Dinstance`

Fix in all cases

app/main.py
```diff
@@ -1,3 +1,87 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
AnyoneClown, line 15: Нou can only have 3 types. `float` `int` or `Distance`. There is no need to check for `Distance` if we have checked the type for `int` and `float`. If it's not Distnace - it's `int` or `float` type. 

Refactor in all cases using only 1 `isinstance()` method

app/main.py
```diff
@@ -1,3 +1,87 @@
...
+        else:
+            raise TypeError(f"unsupported operand type(s) for *: "
+                            f"'Distance' and {type(other)}")
+
+    def __truediv__(self, other: float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError(f"unsupported operand type(s) for /: "
+                            f"'Distance' and {type(other)}")
```
AnyoneClown, line 42: `other` can also be `int` here. Add it in type hint.

`other` can not be of type `Distance` in this method, because if you multiply km * km you get square km. Don't check for instance type here

app/main.py
```diff
@@ -1,3 +1,87 @@
...
+
+    def __ge__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        else:
+            raise TypeError(f"unsupported operand type(s) for >=: "
+                            f"'Distance' and {type(other)}")
```
AnyoneClown, line 87: Refactor all of this by comments above

AnyoneClown (changes requested):

PR #1437, opened 2024-10-01
Solution

MrEgorLite:


SerhiiK05 (approved):
GJ!

app/main.py
```diff
@@ -1,3 +1,61 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
```
AnyoneClown, line 3: km can also be `float` type here

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
AnyoneClown, line 11: Add more descriptive annotation to `other`. It can also be of type `int` or `float`.

Why are you using quotes for "Distance" everywhere? You can do it without them example: `Dinstance`
(Import annotations from future for this)

Fix in all cases

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
```
AnyoneClown, line 16: You don't need to change original object here. Create a new after calculating self.km

Example:
```
```

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __mul__(self, other: float) -> "Distance":
+        self.km *= other
+        return self
+
+    def __truediv__(self, other: float) -> "Distance":
+        self.km = round(self.km / other, 2)
+        return self
```
AnyoneClown, line 31: You don't need to change original object here. Create new `Dinstnace` instance

`other` can also be `int` here

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __iadd__(self, other: "Distance") -> "Distance":
```
AnyoneClown, line 18: Fix type hint for `other`

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
AnyoneClown, line 61: Fix type hints for `other`. You don't need to use `else` here, refactor without using it

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #1439, opened 2024-10-01
Solution

molodsh1y1:


mshmygel:
GJ!

mshmygel (approved):
GJ!

mshmygel (approved):

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    @staticmethod
+    def validate(other: Union[int, float, "Distance"]) -> float:
```
AnyoneClown, line 9: From python 3.10+ we can use `|` for annotations. Fix in all cases

```suggestion
    def validate(other: int | float | Distance) -> float:
```

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        else:
+            raise TypeError(f"Unsupported type for operation: {type(other)}")
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}(km={self.km})"
+
+    def __add__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
AnyoneClown, line 23: We can import annotations from `future` to use class type in annotation without quotes

AnyoneClown (approved):
Good, but fix annotations

PR #1441, opened 2024-10-02
solution

SerhiiK05:
pytest

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,46 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self):
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,46 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> callable:
```
zaietsmo, line 11:
```suggestion
    def __add__(self, other: Distance | float | int) -> Distance:
```

zaietsmo (changes requested):
other can be: Distance float and int(not only int).
use `from __future__ import annotations`

zaietsmo (approved):
GJ!

AnyoneClown (changes requested):

PR #1442, opened 2024-10-02
Solution

zaietsmo:


pohgen (approved):
good

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: (Distance, int, float)) -> Distance:
```
AnyoneClown, line 14: In this annotation you are waiting other to be a tuple with 3 arguments?

You did correct in `__init__` method. We can use `|` to provide multiple types here. Fix in all cases pls

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
AnyoneClown, line 18: We don't need to check for 2 types at the same time. If it's not `Disntance` type - it will be `int` or `float`. So we can do only 1 check in all cases. Fix this in other methods too

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __iadd__(self, other: (Distance, int, float)) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: (Distance, int, float)) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
```
AnyoneClown, line 29: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

Fix annotation too

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __truediv__(self, other: (Distance, int, float)) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
AnyoneClown, line 33: `other` can not be of type `Distance` in this method

AnyoneClown (changes requested):

Nattalli (approved):

PR #1443, opened 2024-10-02
solution

pohgen:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        else:
+            return self.km < other
```
YaYaYaroslav, line 39: It looks better this way

`if isinstance(other, Distance):
            return self.km < other.km
return self.km < other`

YaYaYaroslav (approved):

app/main.py
```diff
@@ -1,3 +1,45 @@
+from __future__ import annotations
+from functools import total_ordering
+
+
+@total_ordering
```
AnyoneClown, line 5: Cool that u know about this decorator :)

app/main.py
```diff
@@ -1,3 +1,45 @@
...
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
```
AnyoneClown, line 8: `km` can also be `float` type

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
AnyoneClown, line 19: We don't need to use `else` in this case. You can do a return from  a new line.

`other` can also be `float` type. Fix in all cases

app/main.py
AnyoneClown: +, fix it 👍 (in all cases)

AnyoneClown (changes requested):
Where is other methods like `gt` `le` `ge`? You should implement all of them

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
AnyoneClown, line 16: `other` can be also `float`. Fix in all cases

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __eq__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
```
AnyoneClown, line 42: You still didn't implement 3 more methods

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job

PR #1444, opened 2024-10-02
solution

YaYaYaroslav:


sumseer (approved):
Good job

app/main.py
```diff
@@ -1,3 +1,71 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: Union[int, float]) -> None:
```
AnyoneClown, line 5: From python 3.10+ we can use `|` in annotations.

```suggestion
    def __init__(self, km: int | float) -> None:
```
Fix in all cases

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        self.km: Union[int, float] = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", int, float]) -> "Distance":
```
AnyoneClown, line 14: Import annotation from `future` to give annotation for classes without quotes

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1445, opened 2024-10-02
Solution

sumseer:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
```
KKvladys, line 30: in metods "mul" and "truediv" you dont need to check "other" for int/float types, because he not be of type Distance

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __lt__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
```
KKvladys, line 36: you can skip "elif" her and below and just use "return"

KKvladys (approved):

app/main.py
```diff
@@ -1,3 +1,60 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
```
AnyoneClown, line 11: `other` can be `Distance` type here

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
AnyoneClown, line 15: We don't need to check for type twice. If it's not `Distance` -> it will be `float` or `int` type

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __iadd__(self, other: int | float) -> "Distance":
```
AnyoneClown, line 17: Import annotations from `future` to use classes as type hints without quotes

app/main.py
AnyoneClown: 👍 fix this

app/main.py
AnyoneClown: 👍 fix this in all cases

Also add annotation for `Disntace` in `other` argument

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #1446, opened 2024-10-02
Solution

MatvijSafonov:


app/main.py
```diff
@@ -1,3 +1,37 @@
...
+        return self
+
+    def __mul__(self, other: float) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: "Distance | float") -> bool:
+        return self.km < (other.km if isinstance(other, Distance) else other)
```
Den-k0, line 25: Maybe, you can create separate method for "(other.km if isinstance(other, Distance) else other)" to reduce code duplication. But in general looks good to me.

app/main.py
```diff
@@ -1,3 +1,37 @@
+from dataclasses import dataclass
+
+
+@dataclass
```
AnyoneClown, line 4: Cool that you know about `@dataclass` decorator :)

app/main.py
```diff
@@ -1,3 +1,37 @@
...
 class Distance:
-    # Write your code here
-    pass
+    km: float
```
AnyoneClown, line 6: km can also be `int`

app/main.py
```diff
@@ -1,3 +1,37 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __add__(self, other: "Distance | float") -> "Distance":
```
AnyoneClown, line 11: Import annotation from `future` to use this types without quotes. other can also be `int` her. Check all other cases

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,24 +1,32 @@
...
-    def __iadd__(self, other: "Distance | float") -> "Distance":
+    def __add__(self, other: Distance | float | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (float, int)):
+            return Distance(self.km + other)
+        raise TypeError(
+            f"Unsupported operand type(s) for +: 'Distance' and '"
+            f"{type(other).__name__}'"
+        )
```
Nattalli, line 20: last raising is unnecessary, but can be

app/main.py
```diff
@@ -1,24 +1,32 @@
...
+
+    def __iadd__(self, other: "Distance | float") -> Distance:
```
Nattalli, line 22: why do you use `""`

app/main.py
```diff
@@ -1,24 +1,32 @@
...
         self.km += (other.km if isinstance(other, Distance) else other)
         return self
 
-    def __mul__(self, other: float) -> "Distance":
+    def __mul__(self, other: float) -> Distance:
```
Nattalli, line 26: other can be int also

app/main.py
```diff
@@ -1,24 +1,32 @@
...
         return Distance(self.km * other)
 
-    def __truediv__(self, other: float) -> "Distance":
+    def __truediv__(self, other: float) -> Distance:
```
Nattalli, line 29: and here `other` can be int

Nattalli (changes requested):

Nattalli (approved):

PR #1447, opened 2024-10-02
Develop

Den-k0:


dxrrkwm (approved):
good job, clean and well readable code!

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+from functools import total_ordering
+from typing import Union
+
+
+@total_ordering
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: Union[int, float]) -> None:
+        self.km = km
```
AnyoneClown, line 9: From python 3.10+ you can use `|` to add multiple type hints

```suggestion
    def __init__(self, km: int | float) -> None:
        self.km = km
```

Change in all cases

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+    def __truediv__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError("Distance instance is not supported")
+        return Distance(round(self.km / self.validate(other), 2))
+
+    def __lt__(self, other: Union[int, float, Distance]) -> bool:
+        return self.km < self.validate(other)
+
+    def __eq__(self, other: Union[int, float, Distance]) -> bool:
+        return self.km == self.validate(other)
```
AnyoneClown, line 47: You also need to implement other methods (gt, le, ge)

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+        return Distance(self.km + self.validate(other))
+
+    def __iadd__(self, other: Union[int, float, Distance]) -> Distance:
+        self.km += self.validate(other)
+        return self
+
+    def __mul__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError("Distance instance is not supported")
+        return Distance(self.km * self.validate(other))
```
AnyoneClown, line 36: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def __truediv__(self, other: Union[int, float, Distance]) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError("Distance instance is not supported")
+        return Distance(round(self.km / self.validate(other), 2))
```
AnyoneClown, line 41: `other` can not be of type `Distance` in this method

AnyoneClown (changes requested):

app/main.py
Den-k0: It already implemented by total_ordering module from functools, that allow us to make code more shorter. Or I must do it without other modules?

AnyoneClown (approved):
Good job!

PR #1448, opened 2024-10-02
long solution

dxrrkwm:


sofiiasavkova (commented):
seems to be okay)

sofiiasavkova (approved):

app/main.py
```diff
@@ -1,3 +1,58 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance") -> "Distance":
```
AnyoneClown, line 11: Import annotation from `future` to use imports without quotes.

`other` can also be `int` or `float` type(check all other cases)

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
+
+    def __mul__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        return NotImplemented
```
AnyoneClown, line 28: `other` can not be of type `Distance` in this method, because if you multiply km * km you get square km

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __truediv__(self, other: "Distance") -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        return NotImplemented
```
AnyoneClown, line 33: `other` can not be of type `Distance` in this method

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return self.km > other
+
+    def __eq__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: "Distance") -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km or self.km == other.km
```
AnyoneClown, line 52: You can use `<=` at once, why both methods?

Fix in other cases too

tests/test_main.py
AnyoneClown, line 1: Don't make any changes to test files

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job, next time don't change tests)

PR #1451, opened 2024-10-02
Solution

Rabb1t-Artem:


YaroslavK87 (approved):

app/main.py
```diff
@@ -1,3 +1,42 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance) -> Distance:
```
AnyoneClown, line 16: `other` can be also `float` or `int` type here

app/main.py
```diff
@@ -1,3 +1,42 @@
+from __future__ import annotations
+from functools import total_ordering
+
+
+@total_ordering
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
AnyoneClown, line 7: km can also be `float` here. Fix in all places

AnyoneClown (approved):
Good, but fix type hints

PR #1453, opened 2024-10-02
Solution

Bondzik-S:


OlhaStadnik (approved):

app/main.py
```diff
@@ -1,3 +1,46 @@
+from functools import total_ordering
+
+
+@total_ordering
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
+        self.km = km
```
AnyoneClown, line 7: km can also be `float`. Fix in all cases

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
```
AnyoneClown, line 15: Import annotation from `future` to use type hints without quotes

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
```
AnyoneClown, line 19: We don't need to check it twice. If it's not `Distance` - it will be `int` or `float`

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+
+    def __eq__(self, other: int) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
```
AnyoneClown, line 46: Where is 3 more methods?

AnyoneClown (changes requested):
Read task properly

AnyoneClown (approved):
Good job!

PR #1460, opened 2024-10-04
Add distance class solution

charmheroku:


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(
+            self, new_distance: Union["Distance", int, float]
+    ) -> "Distance":
+        if isinstance(new_distance, (int, float)):
+            return Distance(self.km + new_distance)
+        else:
```
pavlopro, line 19: `else` operator is redundant here. Just `return Distance(self.km + new_distance.km)`
Fix in places below as well

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        self.km = kilometers
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(
+            self, new_distance: Union["Distance", int, float]
```
pavlopro, line 15: PLease correct annotation to Distance | int | float
correct in places below as well
use `from __future__ import annotation` if needed

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+    ) -> "Distance":
```
pavlopro, line 16: Also correct return annotaion to just `Distance`

pavlopro (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(distance, (int, float)):
+            return self.km == distance
+        return self.km == distance.km
+
+    def __le__(self, distance: int | float) -> bool:
+        if isinstance(distance, (int, float)):
+            return self.km <= distance
+        return self.km <= distance.km
+
+    def __ge__(self, distance: int | float) -> bool:
```
vsmutok, line 52:
```suggestion
    def __ge__(self, distance: int | float | Distance) -> bool:
```
please fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1461, opened 2024-10-07
Implemented Distance

This-is-Lapwing:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" or float) -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | float | int) -> Distance:
```
please fix everywhere

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            return NotImplemented
```
vsmutok, line 17:
```suggestion
            return Distance(self.km + other)
        return NotImplemented
```
`else` redundant in this case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1463, opened 2024-10-08
Solution

RadeonT800:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance | float") -> "Distance":
```
LLkaia, line 14: Remove quotes for type annotation here and all cases bellow.
Import this to allow everything works: `from __future__ import annotations`

Also, are you sure that float is the only one builtin type that can be used for your class operations?

LLkaia (changes requested):

PR #1464, opened 2024-10-09
Distance class completed

Waldemar-wal:


app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __iadd__(self, other: float | int | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: float | int | Distance) -> Distance:
+        return Distance(self.km * other)
```
Dimosphen1, line 27: It seems that `other` could be only an instance of the `float` or `int` type to process multiplication correctly. Make sure that you indicate verified types that are applicable for the function in the type annotation in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #1498, opened 2024-10-29
solve_task

Olha-y:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
```
vasylhnatiuk, line 11: km can be also Distance instance

vasylhnatiuk: fix everywhere

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,58 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: Union[int, float]) -> None:
```
pavlopro, line 5: Instead of `Union` use `int | float`
Correct in places below

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[int, float, "Distance"]) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
pavlopro, line 29: remove  "Distance" from other annotation

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Union[int, float, "Distance"]) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[int, float, "Distance"]) -> "Distance":
```
pavlopro, line 26: remove "Distance" from other annotation

pavlopro (changes requested):

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: int | float) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
```
vsmutok, line 24:
```suggestion
    def __mul__(self, other: int | float | Distance) -> Distance:
```
please fix everywhere 
please read https://github.com/mate-academy/py-distance-class/blob/master/checklist.md

vsmutok (changes requested):

app/main.py
Olha-y: how to do that? I tried to fix this but the last I have a problem : Unresolved reference 'Distance'

Olha-y: Thank you for help me with this part of code, I forgot to do this import(

app/main.py
```diff
@@ -1,3 +1,59 @@
...
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> Distance:
```
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

vasylhnatiuk: add everywhere where other can be Distance instance

vasylhnatiuk (approved):

PR #1511, opened 2024-11-08
Solution

Chukhno-Maksym:


app/main.py
```diff
@@ -1,3 +1,71 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> int:
```
petrykivd, line 12: Please fix type annotations for func arguments everwhere, try smtn like this Distance | int | float etc.

app/main.py
```diff
@@ -1,3 +1,71 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(
+                km=self.km + other
+            )
+        else:
+            return Distance(
+                km=self.km + other.km
+            )
```
petrykivd, line 20: Here you can do without else, check another, fix it in other places too

petrykivd (changes requested):

app/main.py
```diff
@@ -1,3 +1,66 @@
...
+        return Distance(
+            km=self.km + other
+        )
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        self.km += other
+        return self
```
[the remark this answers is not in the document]
L1nk3rrr: Figure out how to use only 1 return

L1nk3rrr (changes requested):
Look at my comments, also you have declaration like
`Distance(
                km=self.km + other.km
            )`
  I think you can use simple one line, because you don't have a lot of attributes or long names
  `Distance(km=self.km + other.km)`

L1nk3rrr (approved):

PR #1531, opened 2024-11-14
Solution

julia4406:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @staticmethod
+    def is_distance(some_data: Distance | int | float) -> Any:
+        if isinstance(some_data, Distance):
+            return some_data.km
+        else:
+            return some_data
```
L1nk3rrr, line 20: No need to use `else`

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @staticmethod
+    def is_distance(some_data: Distance | int | float) -> Any:
```
L1nk3rrr, line 16: The name of method looks like checking if `is`, maybe we can do a better naming, also it returns not `Any`, only int or float looks like

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        self.km += self.is_distance(other)
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if not isinstance(other, (int, float)):
+            raise TypeError(
+                f"Unsupported operation class Distance * {type(other)}"
+            )
```
L1nk3rrr, line 33: If you already added a check, why we are checking only at `__mul__` and `__truediv__` methods?

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if not isinstance(other, (int, float)):
+            raise TypeError(
+                f"Unsupported operation class Distance / {type(other)}"
+            )
```
L1nk3rrr, line 40: If you already added a check, why we are checking only at `__mul__` and `__truediv__` methods?

app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        result = round(self.km / other, 2)
```
L1nk3rrr, line 41: I think you don't need that variable here

L1nk3rrr (changes requested):
Look for a comments that I left, also if you want -  you can check for a `total_ordering` decorator from a `functools` lib, that can simplify some part of your code.
Here is some links for you [python docs](https://docs.python.org/3/library/functools.html#functools.total_ordering), [example of usage](https://testdriven.io/tips/d6fbb8c8-6303-418e-98de-629fd841c156/)

app/main.py
julia4406: Staticmethod above is not bool (returns data). May be it's possible to use it in __mul__ and __truediv__ methods, but code became knotty (cause only in this methods we have to raise error message). I recompose methods to avoid using 'not', but prefer to keep this construction without using staticmethod here. Are you agree?

L1nk3rrr: In this situation no need to remove it because it's logical. Thanks for explaining your solution here, and yep, I agree with you

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    @staticmethod
+    def conversion_data(
+            some_data: Distance | int | float) -> Distance | int | float:
```
L1nk3rrr, line 18: Looks like we don't need a new line here

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        self.km += self.conversion_data(other)
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        else:
+            raise TypeError(
+                f"Unsupported operation class Distance * {type(other)}"
+            )
```
L1nk3rrr, line 36: In this situation, I don't left comments about `not`, because you are raising errors, so with `not` your code will look more logical and style-guided.

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError(
+                f"Unsupported operation class Distance / {type(other)}"
+            )
```
L1nk3rrr, line 44: In this situation, I don't left comments about not, because you are raising errors, so with not your code will look more logical and style-guided.

app/main.py
L1nk3rrr: Also you don't return `Distance `, only int or float

app/main.py
L1nk3rrr: At the previous task I make accent on it because we have `else` already, at this situation without `not` our code will look more complicated

petrykivd (approved):
Nice! 😎

PR #1555, opened 2024-11-22
solution

VovaGld:


app/main.py
```diff
@@ -1,3 +1,79 @@
...
+            km=self.km + self.check_value(other)
+        )
+
+    def __iadd__(self, other: [Distance, int]) -> Distance:
+        self.km += self.check_value(other)
+        return self
+
+    def __mul__(self, other: [Distance, int]) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError("Cannot divide one Distance instance by another")
```
julia4406, line 33: Not divide, multiply.

app/main.py
```diff
@@ -1,3 +1,79 @@
...
+                f"Unsupported type for division: {type(other).__name__}"
+            )
+
+    def __truediv__(self, other: [Distance, int]) -> Distance:
+        if isinstance(other, Distance):
+            raise TypeError("Cannot divide one Distance instance by another")
+        elif isinstance(other, (int, float)):
+            return Distance(
+                round((self.km / other), 2)
+            )
```
julia4406, line 50: In this return(48) Distance gives a float value (but accordingly to  def __init__ is int)

PR #1574, opened 2024-11-30
Solution

taraskhom:


app/main.py
```diff
@@ -1,3 +1,78 @@
...
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, int):
+            return Distance(self.km + other)
+        if isinstance(other, float):
+            return Distance(self.km + other)
```
julia4406, line 21: You return same Distance(self.km + other) two times, possible to combine them: if isinstance(other, int | float):... or somehow else. 
The same about other methods below.

julia4406 (changes requested):

julia4406 (approved):

app/main.py
taraskhom: Thanks for the review! I didn't notice for the first time that I could simplify these expressions.

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, int | float):
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int | Distance | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        if isinstance(other, int | float):
+            self.km += other
+            return self
```
Nattalli, line 27: you can avoid using `return self` twice

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, int | float):
+            return self.km == other
+
+    def __le__(self, other: int | Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        if isinstance(other, int | float):
+            return self.km <= other
```
Nattalli, line 57: you can make it in 1 line with using of `__lt__` and `__eq__` methods

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+
+    def __ge__(self, other: int | Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        if isinstance(other, int | float):
+            return self.km >= other
```
Nattalli, line 63: the same here

app/main.py
```diff
@@ -1,3 +1,63 @@
...
+    def __gt__(self, other: int | Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        if isinstance(other, int | float):
+            return self.km > other
+
+    def __eq__(self, other: int | Distance | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        if isinstance(other, int | float):
```
Nattalli, line 50: you can avoid usage of second `if` statement. Read the checklist

Nattalli (changes requested):

Nattalli (approved):

PR #1575, opened 2024-12-01
Created a class with all needed arithmetic magic methods

DanilTrukhanov:


VovaGld:
You could create a separate class method to check the type of the value to avoid code duplication, but overall it's fine )

VovaGld (approved):

PR #1578, opened 2024-12-02
Implemented magic methods for Distance class

4ikkamony:


app/main.py
```diff
@@ -1,3 +1,61 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
```
bogdAAAn1, line 2: Alternatively, you could put these three “Distance, float, int” types into a separate variable “Variables” and then use it to annotate “other”.
Example:

```
Variables = Distance, int, float

def __add__(self, other: Variables) -> Distance: 
...
```

Otherwise, cool code!

bogdAAAn1 (approved):

PR #1582, opened 2024-12-02
Distance

maxakim0v:


app/main.py
```diff
@@ -1,3 +1,70 @@
...
+        return self
+
+    def __mul__(self, factor: int) -> None:
+        if isinstance(factor, (int, float)):
+            return Distance(self.km * factor)
+        return NotImplemented
+
+    def __truediv__(self, divisor: int) -> None:
+        if isinstance(divisor, (int, float)):
+            return Distance(round(self.km / divisor, 2))
```
ruslanyemets, line 34: It will be better if you exclude division by zero

ruslanyemets (approved):
Good job! But if you wrote in 'isinstance' (int, float) types, it would be better to write that in the annotations as well. Also you can check for division by zero in __truediv__ method.

PR #1619, opened 2024-12-07
solved

Danylo-Ok:


4ikkamony (changes requested):
Nice job! Pytest's tests pass, you just gotta add type hints for your methods and their arguments, as flake says

PR #1650, opened 2024-12-24
implement Distance class

yarzhuravlov:


app/main.py
```diff
@@ -1,3 +1,99 @@
...
+        return (
+            "unsupported operand type(s) for "
+            + f"{operation}: 'Distance' and {other_type}"
+        )
+
+    @staticmethod
+    def get_comparison_error(comparison_sign: str, other_type: type) -> str:
+        return (
+            f'Operator "{comparison_sign}" '
+            + 'not supported for types "Distance" and {other_type}'
```
oliasymak24, line 27: Use double quotes

oliasymak24 (approved):
Good job!

app/main.py
yarzhuravlov: So, if I understand correctly, double quotes should be preferred even if they are used in a string? But pep8 states that backslashes should be avoided. https://peps.python.org/pep-0008/#string-quotes
Also, I see that I made a mistake in the second part of the string - the second string is not formatted.
Thanks for the feedback.

PR #1675, opened 2025-01-11
Solution

Nikita-pyth:


kram3ko (approved):
"I like reusing method:  km = self.__add__(other).km

PR #1676, opened 2025-01-11
The Distance class has been implemented with underscore methods!

kram3ko:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+                            f" 'Distance' and '{type(other).__name__}")
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(km=round(self.km / other, 2))
+        else:
+            raise TypeError(f"Unsupported operand type(s) for /: "
+                            f" 'Distance' and '{type(other).__name__}")
+
+    def __lt__(self, other: int | float) -> bool:
```
oliasymak24, line 42: What if `other` is of type `Distance` here and in all the following methods?

app/main.py
```diff
@@ -1,3 +1,68 @@
+from typing import Type
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: any) -> None:
+        self.km = km
+
+    def __add__(self, other: Type["Distance"] | int | float) -> "Distance":
```
oliasymak24, line 8: Please check your checklist after the task)

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            raise TypeError(f"Unsupported operand type(s) for +: "
+                            f" 'Distance' and '{type(other).__name__}")
+
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
```
oliasymak24, line 29: You don't need this check here, your function anyway takes only int and float types, according to your type annotations

oliasymak24 (changes requested):

kram3ko (commented):
Done

app/main.py
kram3ko: Fixed all following methods.

app/main.py
kram3ko: Thank you :)

app/main.py
kram3ko: Fixed it

kram3ko (commented):
fixed mistakes

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __le__(self, other: Distance | int | float) -> bool:
```
oliasymak24, line 37: try to rewrite this method in one line. Use previously created magic methods

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __gt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        return self.km > other
+
+    def __ge__(self, other: Distance | int | float) -> bool:
```
oliasymak24, line 47: try to rewrite this method in one line. Use previously created magic methods

oliasymak24 (changes requested):

PR #1714, opened 2025-02-02
solution

vladoromus:


app/main.py
```diff
@@ -1,3 +1,61 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
Dimosphen1, line 2: Indicate that `km` can be either `float` or `int` here and in the other places

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> object:
```
Dimosphen1, line 11: Indicate that `other` can be either `float`, `int` or `Distance` class instance here and in the other places. Correct your return type annotation as well to indicate that method will return `Distance` class instance in all corresponding methods

Dimosphen1 (changes requested):

SerhiyTo (approved):

PR #1715, opened 2025-02-02
Solution

chiper9:


app/main.py
```diff
@@ -1,3 +1,93 @@
...
+    def __add__(self, other: int | float) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise (
+                TypeError(f"Unsupported operand type(s):"
+                          f" 'Distance' and {type(other)}")
+            )
```
MNDonut, line 20: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float) -> "Distance":
+        if isinstance(other, Distance):
```
MNDonut, line 12: fix annotation. "other" can also be Distance object

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        else:
+            raise (
+                TypeError(f"Unsupported operand type(s):"
+                          f" 'Distance' and {type(other)}")
+            )
+
+    def __iadd__(self, other: int | float) -> "Distance":
```
MNDonut, line 22: fix annotation. "other" can also be Distance object

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
```
MNDonut, line 37: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        if isinstance(other, (int, float)):
+            if other == 0:
+                raise ZeroDivisionError("Division by zero is not allowed")
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
```
MNDonut, line 46: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+
+    def __lt__(self, other: int | float) -> bool:
```
MNDonut, line 48: fix annotation. "other" can also be Distance object

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
```
MNDonut, line 55: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+
+    def __gt__(self, other: int | float) -> bool:
```
MNDonut, line 57: fix annotation. "other" can also be Distance object

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
+
```
MNDonut, line 65: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
```
MNDonut, line 82: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+    def __ge__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        else:
+            raise (
+                TypeError(f"Unsupported operand type(s): "
+                          f"'Distance' and {type(other)}")
+            )
```
MNDonut, line 93: No need to add "else" because the other two conditions have "return", thus it'll never reach "else" statement if some "if" statement is run

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+    def __le__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
+
+    def __ge__(self, other: int | float) -> bool:
```
MNDonut, line 84: fix annotation. "other" can also be Distance object

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+    def __eq__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        elif isinstance(other, (int, float)):
+            return self.km == other
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
+
+    def __le__(self, other: int | float) -> bool:
```
MNDonut, line 75: fix annotation. "other" can also be Distance object

app/main.py
```diff
@@ -1,3 +1,93 @@
...
+    def __gt__(self, other: int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        elif isinstance(other, (int, float)):
+            return self.km > other
+        else:
+            raise TypeError(f"Unsupported operand type(s): "
+                            f"'Distance' and {type(other)}")
+
+    def __eq__(self, other: int | float) -> bool:
```
MNDonut, line 66: fix annotation. "other" can also be Distance object

MNDonut (changes requested):

app/main.py
```diff
@@ -8,16 +8,19 @@ def __str__(self) -> str:
...
 
+    # якщо вказувати Distance в анатації other
+    # зявляється помилка
+    # TypeError: unsupported
+    # operand type(s)
+    # for |: 'types.UnionType' and 'str'
+
     def __add__(self, other: int | float) -> "Distance":
         if isinstance(other, Distance):
             return Distance(self.km + other.km)
```
Nattalli: but `other` can be a `Distance` class. So I believe that the error is caused because of wrong type annotation syntax

Nattalli (changes requested):
so you still have to fix annotations

app/main.py
```diff
@@ -1,3 +1,90 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    # якщо вказувати Distance в анатації other
+    # зявляється помилка
+    # TypeError: unsupported
+    # operand type(s)
+    # for |: 'types.UnionType' and 'str'
```
MNDonut, line 18: to avoid it add this import:
It'll allow you to add class name as annotation within the same class. And then you can do:
`def __add__(self, other: int | float | Distance): ...`

MNDonut (changes requested):

PR #1726, opened 2025-02-08
Solution

Yevhenii0536:


app/main.py
```diff
@@ -1,3 +1,50 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    @staticmethod
+    def check_is_instance(other: Distance | int) -> bool:
+        return isinstance(other, Distance)
```
MNDonut, line 7: actually no need to create a separate method that contains one simple line

app/main.py
```diff
@@ -1,3 +1,50 @@
...
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Distance | int) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int) -> Distance:
+        self.km = round(self.km / other, 2)
+        return self
```
MNDonut, line 35: what if we divide by 0?

MNDonut (changes requested):

PR #1732, opened 2025-02-11
Implement Distance class

Sotnikov-100:


app/main.py
```diff
@@ -1,3 +1,75 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km: float = km  # вказуємо тип атрибута
```
oliasymak24, line 6: Delete comments

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        return False
+
+    def __ge__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        return False
```
oliasymak24, line 75: try to rewrite these two methods in one line each. Use previously created magic methods

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        else:
+            return NotImplemented
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        if isinstance(other, (int, float)):
```
oliasymak24, line 31: You don't need this check here, your function anyway takes only int and float types, according to your type annotations

oliasymak24 (changes requested):

app/main.py
Sotnikov-100: thank you for the tip. you way is much more concise.

app/main.py
Sotnikov-100: I agree with you!

PR #1759, opened 2025-02-18
solution

cemen2009:


AndyStarGitHub (approved):
Hello!
Several general comments for the future:
1) Result annotations (e.g. -> None in __init__, object in __mul__) should be changed to Distance.
2) km can be not float only, but int too.
3) Checklist in README file requires to avoid else and elif.

PR #1778, opened 2025-02-20
Solution Distance_Class

Roksolana-K:


pewlay (approved):
Good job!

Y-Havryliv (changes requested):
why didn't you fix AI comments? They are mostly valid, only not valid for `mul` and `truediv` methods.

Roksolana-K:
> why didn't you fix AI comments? They are mostly valid, only not valid for `mul` and `truediv` methods.

Sorry, my fault!

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __init__(self, km: "Distance" | int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: "Distance" | int | float) -> "Distance":
```
oliasymak24, line 15: As you are importing `from __future__ import annotations` then you need to use this
In this case you don't need `"Distance"` just `Distance`

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        self.km *= other
+        return self
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        self.km = round(self.km / other, 2)
+        return self
```
oliasymak24, line 31: this method should return a new instance of Distance class

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+        return Distance(self.km + other)
+
+    def __radd__(self, other: "Distance" | int | float) -> "Distance":
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: "Distance" | int | float) -> "Distance":
+        self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
```
oliasymak24, line 25: this method  should return a new instance of Distance class

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: "Distance" | int | float) -> bool:
+        if self.km <= other:
+            return True
+        return False
+
+    def __ge__(self, other: "Distance" | int | float) -> bool:
+        if self.km >= other:
+            return True
+        return False
```
oliasymak24, line 56: try to rewrite these two methods in one line each. Use previously created magic methods

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+    def __mul__(self, other: int | float) -> "Distance":
+        self.km *= other
+        return self
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        self.km = round(self.km / other, 2)
+        return self
+
+    def __lt__(self, other: "Distance" | int | float) -> bool:
+        if self.km < other:
```
oliasymak24, line 34: What if `other` is of type `Distance` here and in all the following methods?

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1787, opened 2025-02-23
task solution

yarms1:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
Maaax256, line 14: you don't need to put elif rows in these methods because after "return" nothing executes

Maaax256 (changes requested):
the rest is fine

PR #1794, opened 2025-02-25
'Solution'

Sezaro104:


app/main.py
```diff
@@ -1,3 +1,58 @@
...
+        self.km = km
+        Distance.distance.append(self)
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
```
oliasymak24, line 17: What if `other` is float here? check everywhere

app/main.py
```diff
@@ -1,3 +1,58 @@
...
+
+    def __le__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        return self.km <= other
+
+    def __ge__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
```
oliasymak24, line 58: try to rewrite these two methods in one line each. Use previously created magic methods

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1873, opened 2025-03-28
implement py-distance-class

romko2003:


romko2003:
I did changes, re-check, please

PR #1898, opened 2025-04-13
init

Bohdan-Salamakha:


Bohdan-Salamakha:
@mateacademy-ai-mentor 
this is not required by the description of the task:
These comparison methods should also accept a Distance object as an argument for comparison, not just an integer. Modify these methods to correctly compare two Distance objects.

Bohdan-Salamakha:
@mateacademy-ai-mentor 
you are not correct.
__iadd__ should return the same object because of the logic

PR #1918, opened 2025-04-25
Solution

Nib20:


app/main.py
```diff
@@ -1,3 +1,60 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Any) -> "Distance":
```
MNDonut, line 14: It's better to avoid "Any" because you won't get hints what methods the "other" object has. So annotate it correctly. For example other can't be "str" because self.km + other will cause error.
The same for other methods

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: Any) -> "Distance":
```
MNDonut, line 19: Instead of "Distance" add the import `from __future__ import annotations` and you won't get an error that you can use itself class in annotation

app/main.py
```diff
@@ -1,3 +1,60 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, other: Any) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Any) -> "Distance":
+        return Distance(round(self.km / other, 2))
```
MNDonut, line 30: Hmm, what if you do 1 / 0? 👀

MNDonut (changes requested):

app/main.py
Nib20: fixed

app/main.py
Nib20: fixed

MNDonut (changes requested):

app/main.py
```diff
@@ -11,49 +12,51 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"Distance(km={self.km})"
 
-    def __add__(self, other: Any) -> "Distance":
+    def __add__(self, other: Union[Distance, int]) -> Distance:
```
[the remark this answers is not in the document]
Nib20: fixed

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: [Distance | int]) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
+
+    def __iadd__(self, other: [Distance | int]) -> Distance:
```
MNDonut, line 19: No need to add [], just use | operator
Here is what happens:
image

MNDonut (changes requested):

app/main.py
Nib20: fixed

PR #1941, opened 2025-05-11
Writed class Distance

vasyl-main-dev:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+    def __init__(self, km: (int, float)) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: ("Distance", int)) -> "Distance":
```
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```
Please fix everywhere

vsmutok (changes requested):

app/main.py
RangerLeoo: As I know you showed me the new type annotation syntax. Is this a best practice in real-world cases than old i used?

vasylhnatiuk (approved):

PR #2015, opened 2025-06-21
'Solution'

DanilHamster:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        elif isinstance(other, (int, float)):
+            total_km = self.km + other
+        return Distance(total_km)
+
+    def __iadd__(self, other: object) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        elif isinstance(other, (int, float)):
+            self.km += other
+        return self
```
Plovuezz, line 24: instead of returning self with no changes, you can raise TypeError

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __mul__(self, other: object) -> "Distance":
+        return Distance(self.km * other)
```
Plovuezz, line 27: if other is Distance class object it will cause error. You can also handle this with raising TypeError

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __truediv__(self, other: object) -> "Distance":
+        result = round(self.km / other, 2)
+        return Distance(result)
```
Plovuezz, line 31: same as above and you also can doo oneliner

PR #2018, opened 2025-06-21
Solution

Plovuezz:


Arch0998 (approved):
Recommendation: Don't use Copilot while you're still learning

PR #2019, opened 2025-06-21
Solution

Arch0998:


app/main.py
```diff
@@ -1,3 +1,44 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
```
Vskesha, line 15: You can also handle here another case when `other` can be neither Distance nor (int, float). In that situations `raise ValueError("message")`
like this:
```python
            return Distance(km=self.km + other.km)
        raise TypeError(
            f"unsupported operand type(s) for +: "
            f"'{self.__class__.__name__}' and '{other.__class__.__name__}'"
        )
```
Use this also in another methods with `other`

Vskesha (approved):
Code is clean and understandable. It also meets all the requirements and works properly.

PR #2020, opened 2025-06-21
Solution

Vskesha:


app/main.py
```diff
@@ -1,3 +1,88 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        if isinstance(other, (int, float)):
+            self.km += other
+            return self
+        raise TypeError(
+            f"unsupported operand type(s) for +: "
+            f"'{self.__class__.__name__}' and '{other.__class__.__name__}'"
+        )
```
yagginov, line 28: Here the error information does not specify the correct operator because this is a method for the '+=' operator.

yagginov (changes requested):
Very good. Just one small mistake

app/main.py
Vskesha: Fixed. Thanks for review.

yagginov (approved):

app/main.py
```diff
@@ -1,3 +1,88 @@
...
+        raise TypeError(
+            f"unsupported operand type(s) for comparing: "
+            f"'{self.__class__.__name__}' and '{other.__class__.__name__}'"
+        )
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        return not self.__gt__(other)
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        return not self.__lt__(other)
```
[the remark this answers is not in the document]
Vskesha: fixed

PR #2032, opened 2025-06-23
Implemented solution.

Otst0y:


Artcrafterrra (approved):
lgtm

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+
+    def __le__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km <= other
+        return self.km <= other.km
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km >= other
+        return self.km >= other.km
```
AnyoneClown, line 56: Refactor the implementation of the le, gt, and ge methods to make them more concise by reusing the already implemented comparison operators lt and eq.

Instead of duplicating logic, use the relationships between comparison operations:

le (≤) is "less than or equal to"

gt (>) is "not less than or equal to"

ge (≥) is "not less than"

app/main.py
```diff
@@ -1,3 +1,56 @@
...
+            self.km += other
+        else:
+            self.km += other.km
+        return self
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Distance | int | float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
AnyoneClown, line 31: `other` type can only be int/float. Change type hint

AnyoneClown (changes requested):

PR #2034, opened 2025-06-23
Implement Distance class with arithmetic and comparison operations

viannik:


B1g-Bear (approved):

L1nk3rrr (approved):
Well done! Nice solution with a decorator!

PR #2043, opened 2025-06-24
solution py distance class

PavelSemenikhin:


app/main.py
```diff
@@ -1,3 +1,55 @@
+from typing import Union
+
```
Salmon-eat, line 2: All good, but u may **from functools import total_ordering**, this will shorten the code

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
```
Salmon-eat, line 3: @total_ordering

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: Union[float, int]) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: Union[float, int]) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: Union["Distance", float, int]) -> bool:
```
Salmon-eat, line 32: in basi need save __eq__ and __lt__ or other

Salmon-eat (approved):
it's great, man:)

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", float, int]) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        return Distance(self.km + other)
```
AnyoneClown, line 17: Use class type hint without quotes

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __ge__(self, other: Union["Distance", float, int]) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        return self.km >= other
+
+    def __eq__(self, other: Union["Distance", float, int]) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
```
AnyoneClown, line 55: Refactor the implementation of the le, gt, and ge methods to make them more concise by reusing the already implemented comparison operators lt and eq.

Instead of duplicating logic, use the relationships between comparison operations:

le (≤) is "less than or equal to"

gt (>) is "not less than or equal to"

ge (≥) is "not less than"

app/main.py
```diff
@@ -1,3 +1,55 @@
+from typing import Union
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km = km
```
AnyoneClown, line 6: `km` can be `int` too

AnyoneClown (changes requested):

PR #2048, opened 2025-06-24
Solution

DenisPrudius:
done

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -34,7 +34,7 @@ class Distance:
             return Distance(round(self.km / other, 2))
         return NotImplemented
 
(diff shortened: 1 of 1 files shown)


Vskesha (commented):
The Distance class is cleanly implemented with proper use of magic methods, allowing intuitive arithmetic and comparison operations. Type annotations are thoughtfully used, and the logic handles both numeric and Distance instances gracefully.

Vskesha (approved):
The Distance class is cleanly implemented with proper use of magic methods, allowing intuitive arithmetic and comparison operations. Type annotations are thoughtfully used, and the logic handles both numeric and Distance instances gracefully.

PR #2049, opened 2025-06-25
Solution

Bebaaaaa:


Sw1then (changes requested):
The code works correctly, but the type annotations do not meet the requirements. Please fix the annotations according to the guide.

PR #2054, opened 2025-06-25
created a class Distance

trellle:


app/main.py
```diff
@@ -1,3 +1,45 @@
+from __future__ import annotations
+from functools import total_ordering
+
+
+@total_ordering
```
MNDonut, line 5: what is the purpose of this decorator?

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
```
MNDonut: if statement contains return, thus no need to add "else"

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, number: int) -> Distance:
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: int) -> Distance:
+        return Distance(round(self.km / number, 2))
```
MNDonut, line 34: what if we dividy it by 0?

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+
+    def __lt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
```
MNDonut, line 38: if statement contains return, thus no need to add "else"

app/main.py
```diff
@@ -1,3 +1,45 @@
...
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
+        return self
+
+    def __mul__(self, number: int) -> Distance:
+        return Distance(self.km * number)
+
+    def __truediv__(self, number: int) -> Distance:
```
MNDonut, line 31: Did you implement all methods? What about __gt__, __le__, __ge__?
image

MNDonut (changes requested):

app/main.py
trellle: for comparison functions

app/main.py
trellle: ok, what should I return when parametr is 0? raise an error?

app/main.py
trellle: it works because of @total_ordering, Serhiy showed us it

app/main.py
MNDonut: ZeroDivisionError

MNDonut (changes requested):

AnyoneClown (approved):
LGTM

PR #2056, opened 2025-06-25
Develop

reinvvay:


app/main.py
```diff
@@ -1,3 +1,73 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: float) -> None:
+        self.km: float = km
```
AnyoneClown, line 6: It can be `int` too

AnyoneClown (approved):
Good, only small type hint fix

PR #2075, opened 2025-07-03
feat: solution

ArtemZakhar:
Hello)
Very interesting task)
Have a nice day)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,67 @@
+from typing import Union
+
+AliasOtherType = Union["Distance", int, float]
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: AliasOtherType) -> "Distance":
+        if self.is_number(other):
+            return Distance(km=self.km + other)
+
+        return Distance(
+            km=self.km + other.km
+        )
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,67 @@
+from typing import Union
+
+AliasOtherType = Union["Distance", int, float]
```
MNDonut: We don't use Union in the newer versions of Python. You can use | operator like: `User | Company | list`

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+        else:
+            self.km += other.km
+
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        return Distance(round(self.km / other, 2))
```
MNDonut, line 43: What if you divide by zero?

app/main.py
```diff
@@ -1,3 +1,67 @@
...
+
+    def __iadd__(self, other: AliasOtherType) -> "Distance":
+        if self.is_number(other):
+            self.km += other
+        else:
+            self.km += other.km
+
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
```
MNDonut: Don't use quotes for annotation. You can add the import
`from __future__ import annotations` that will not raise an error in case you remove quotes

MNDonut (changes requested):
Have a nice day ;)

app/main.py
ArtemZakhar: And now it looks better?

MNDonut: yep

MNDonut (approved):

PR #2164, opened 2025-09-01
create-distance-class: added specific methods

kenu21:


kenu21:
@mateacademy-ai-mentor 

> Invalid isinstance usage
> 
> Location: app/main.py — in __add__ and __iadd__ (lines where you check other’s type)
> Issue: Using isinstance(other, int | float) is not valid Python. This will raise a TypeError at runtime.
> Requirement: Checklist “Support correct types”.
> How to fix: Use a tuple of types: isinstance(other, (int, float)).

It's valid in new version of Python. You're just too old...

kenu21:
@mateacademy-ai-mentor 
> Decision: REQUEST_CHANGES
> 
> Reason (critical blocker): Your implementation broadly meets the requirements for the Distance class, but it violates the checklist item “Support correct types” due to invalid isinstance usage. Using isinstance(other, int | float) (and similar unions inside isinstance) is not valid and will raise a TypeError at runtime, breaking arithmetic/comparison paths. This is a critical blocker because it causes runtime errors and prevents correct functionality.
> 
> Issues to fix (with specifics and how to resolve):
> 
> Invalid isinstance syntax in app/main.py
> Examples (line numbers may vary):
> In methods like __mul__, __truediv__, and possibly comparisons: isinstance(other, int | float)
> Why it’s wrong: | unions are valid in type annotations, but not in isinstance() checks. Python expects a type or a tuple of types.
> How to fix:
> Replace isinstance(other, int | float) with isinstance(other, (int, float)).
> If you check against multiple types including Distance, use isinstance(other, (int, float, Distance)).
> What you did well:
> 
> All required dunder methods are present and correctly implemented for both Distance and numeric types.
> String representations match the exact formats; division rounding to two decimals is correct.
> Clean structure, correct type annotations, and use of from __future__ import annotations.
> Next steps: Apply the tuple-based isinstance fixes throughout app/main.py. After this change, your solution should fully meet the requirements. Nice work overall—this is a quick fix to unblock approval!

You're too old to know new style of python. You're wrong!

PR #2172, opened 2025-09-08
Soulution

Vorongor:


app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        return NotImplemented
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        return NotImplemented
+
+    def __gt__(self, other: Distance | int | float) -> bool:
```
DonDmytron: You can use opposite methods for cleaner code, for example,if 5 > 2 than 2 <= 5 (le and gt)

app/main.py
```diff
@@ -1,3 +1,75 @@
...
+        elif isinstance(other, (int, float)):
+            return self.km <= other
+        return NotImplemented
+
+    def __ge__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        elif isinstance(other, (int, float)):
+            return self.km >= other
+        return NotImplemented
```
DonDmytron, line 67: Please look at the examples from the checklist and clean up the code.

DonDmytron (changes requested):

DonDmytron (approved):
ok, but i mean that it isn't necessary to return NotImplemented, because by default without return functions return None

PR #2214, opened 2025-09-25
Solution

tradewcs:


.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 inline-quotes = "
-ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
+ignore = E203, E266, E501, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
```
DonDmytron: don't ignore 501 mistake, example how fix it :
long_string = (
    "This is a very long string that needs to be broken "
    "into multiple lines to avoid the E501 error."
)

app/main.py
```diff
@@ -1,3 +1,89 @@
...
+    def __lt__(self, other: Distance) -> bool:
+        if isinstance(other, (int, float)):
+            return self.km < other
+
+        if not isinstance(other, Distance):
+            raise TypeError(f"unsupported operand type(s) for <: 'Distance' and '{type(other).__name__}'")
+
+        return self.km < other.km
+
+    def __gt__(self, other: Distance) -> bool:
```
DonDmytron: You can use opposite methods for cleaner code, for example,if 5 > 2 than 2 <= 5 (__le__ and __gt__)

DonDmytron (changes requested):

DonDmytron (approved):
ok

PR #2266, opened 2025-10-22
add solution

VadymVynokur:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
```
petrykivd: other can be not only int :)

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        else:
```
petrykivd: Also you can remove else here, first return will finish func execution, if not func continue witout else construction

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: int) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        else:
+            self.km += other
+            return self
```
petrykivd: Same here about type hinting

petrykivd: try to rewrite without having return duplicated

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        if isinstance(other, Distance):
+            return self.km <= other.km
+        else:
+            return self.km <= other
+
+    def __ge__(self, other: int) -> bool:
+        if isinstance(other, Distance):
+            return self.km >= other.km
+        else:
+            return self.km >= other
```
petrykivd: Let's simplify without if else construction, remember, you already implemented lt and gt methods

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+            return self
+
+    def __mul__(self, other: int) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> "Distance":
+        if isinstance(self, Distance):
+            return Distance(round(self.km / other, 2))
+        else:
+            raise TypeError("Cannot divide distance123")
```
petrykivd: Fix with f string

app/main.py
petrykivd: Check type hints in all program

petrykivd (changes requested):

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        if isinstance(other, Distance):
+            return self.km == other.km
+        else:
+            return self.km == other
+
+    def __le__(self, other: Distance | int) -> bool:
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __ge__(self, other: Distance | int) -> bool:
+        return self.__gt__(other) or self.__eq__(other)
```
petrykivd: Replace method calls with < , >  😉

app/main.py
```diff
@@ -1,3 +1,55 @@
...
+        else:
+            return self.km < other
+
+    def __gt__(self, other: Distance | int) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
+        else:
+            return self.km > other
+
+    def __eq__(self, other: Distance | int) -> bool:
```
petrykivd, line 45: What about floats in type hints?:)

petrykivd (changes requested):

app/main.py
VadymVynokur: float were not mentioned in the task condition, or rather their mention was only for __truediv__

petrykivd: So we can first divide the distance and then add it to the other one, in which case other can be float? :)

Also please before pushing your solution, read [https://github.com/mate-academy/py-distance-class/blob/master/checklist.md](checklist) 😉

petrykivd (approved):
I approve this, but please correct the changes I requested 😃

PR #2272, opened 2025-10-25
Solution

leemonch1k-space:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+            self.km += other.km
+            return self
+        self.km += other
+        return self
+
+    def __mul__(self, other: float) -> Distance:
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: float) -> Distance:
+        return Distance(round(self.km / other, 2))
```
vadym-zahrebelnyi, line 31: There aren't any instructions for the ways how to avoid zerodivision issues in README, that's why i don't give u any advice what to do with that, just an optional hint

vadym-zahrebelnyi (approved):

PR #2275, opened 2025-10-26
Completed Task

ostapshpeha:
All tests passed, flake8 without errors


app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: object) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
+        else:
+            return Distance(km=self.km + other)
```
Vorongor: Currently, you don’t check the type of other strictly — e.g., adding a string "10" would cause a runtime error.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+
+    def __iadd__(self, other: object) -> "Distance":
+        if isinstance(other, Distance):
+            self.km += other.km
+        else:
+            self.km += other
```
Vorongor, line 20: Same thing as with __add__.

app/main.py
```diff
@@ -1,3 +1,65 @@
...
+        return self
+
+    def __mul__(self, other: float) -> "Distance":
+        if isinstance(other, Distance):
+            return None
+        else:
+            return Distance(km=self.km * other)
```
Vorongor: Same thing as with add.

Vorongor (approved):
Your Distance class is almost perfect! It correctly implements all the required magic methods, supports both Distance objects and numeric operations, and follows Pythonic conventions.
Clear and consistent implementation — all methods are readable and use proper naming conventions. String and repr methods are correct and match the expected output format. Arithmetic operations (__add__, __iadd__, __mul__, __truediv__) correctly return a new Distance instance, but mist option if other won`t be an int or float object.

PR #2276, opened 2025-10-26
Solution

stanislav-kishechko:


app/main.py
```diff
@@ -1,3 +1,232 @@
...
+        instance.
+
+        :param other: The numeric factor by which the distance
+        will be multiplied.
+        :type other: float or int
+        :return: A new Distance instance where the distance
+        in kilometers is multiplied by the given factor.
+        :rtype: Distance
+        """
+        return Distance(self.km * other)
```
Vorongor, line 120: Currently, the method assume other is numeric — if someone does Distance(10) * "5", it will raise a TypeError from Python implicitly.

app/main.py
```diff
@@ -1,3 +1,232 @@
 class Distance:
-    # Write your code here
-    pass
+    """
+    Represents a distance measurement in kilometers.
```
Vorongor, line 6: Docstrings excellent for learning or documentation, but in production you might shorten them slightly (most developers will rely on type hints and IDE popups).

Vorongor (approved):
Your Distance class is well-structured, readable, and richly documented. You’ve clearly put effort into making it both educational and functional, which is great for maintainability or team projects. But it could be better if supplemented by int/float verification  for Comparison and Arithmetic operators

PR #2278, opened 2025-10-27
Soullution

Iventyk:


app/main.py
```diff
@@ -1,3 +1,73 @@
...
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        elif isinstance(other, (int, float)):
```
AngryAndrii: It's a bad practice to use here if/elif statement. You could use
`If statemen:
     do this
do this`
please read the checklist included in the task along with the Readme

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+            return Distance(self.km + other)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km += other.km
+            return self
+        elif isinstance(other, (int, float)):
```
AngryAndrii: The same, here is example:
    return "Valid number!"
return "Invalid number!"

app/main.py
```diff
@@ -1,3 +1,73 @@
...
+
+    def __truediv__(self, number: int | float) -> Distance:
+        return Distance(round((self.km / number), 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        else:
```
AngryAndrii: in these iterative methods you can write shorter, I guess:
return self.km < other

AngryAndrii (changes requested):
Hello, good work! But I think a few changes need to be made

AngryAndrii (approved):
excellent

PR #2279, opened 2025-10-27
added all dunders

AngryAndrii:


app/main.py
```diff
@@ -1,3 +1,48 @@
...
+        return self.km > other
+
+    def __eq__(self, other: Union[int, float]) -> bool:
+        return self.km == other
+
+    def __le__(self, other: Union[int, float]) -> bool:
+        return self.km <= other
+
+    def __ge__(self, other: Union[int, float]) -> bool:
+        return self.km >= other
```
Iventyk: You should add 'Distance' to the type annotations according to the requirements, where comparisons between two 'Distance' instances are possible. Currently, you are accepting only integers and floats.

Iventyk (changes requested):

PR #2282, opened 2025-10-27
Solution

Hellsinger1:


app/main.py
```diff
@@ -1,3 +1,86 @@
...
+        return Distance(self.km + value)
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        value = None
+        if isinstance(other, Distance):
+            value = other.km
+        elif isinstance(other, (float, int)):
+            value = other
+        else:
+            return NotImplemented
```
vsmutok, line 32: You have a lot of code duplication.
Create a separate function for this logic and use it everywhere.

vsmutok (changes requested):

PR #2283, opened 2025-10-27
Solution

SofaD8:


app/main.py
```diff
@@ -1,3 +1,52 @@
...
+            self.km += other
+        return self
+
+    def __mul__(self, other: int | float) -> "Distance":
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int | float) -> "Distance":
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float) -> bool:
```
Taras-Ivashchuk: Hi Sofa.  In bool function headers (lt, gt, eq...)  annotation for paramater 'other' looks like this other: int | float or other: object. It doesn't check Distance type. Correct should be: other : "Distance | int | float"

Taras-Ivashchuk (changes requested):
Please fix parameter annotations for bool methods

Taras-Ivashchuk (approved):
great job!

PR #2299, opened 2025-10-29
created class Distance, implemented all major operation(+-*/><>=<===)

AlexMishKo:


Najttsu (approved):

app/main.py
```diff
@@ -1,3 +1,88 @@
...
+            return NotImplemented
+        return Distance(self.km * other)
+
+    def __rmul__(self, other: Union[int, float, "Distance"]) -> "Distance":
+        return self.__mul__(other)
+
+    def __truediv__(self, other: Union[int, float, "Distance"])\
+            -> Union["Distance", type(ZeroDivisionError)]:
+        if not isinstance(other, (int, float)):
+            return NotImplemented
```
DonDmytron: Please remove the NotImplemented return in all methods, as the methods return None by default. Here is an example of how you can simplify the code: 
    def __truediv__(self, other: int | float) -> Distance | None:
        if isinstance(other, (int, float)) and other != 0:
            return Distance(round(self.km / other, 2)) . Simplify logic in other methods please.

DonDmytron (changes requested):

app/main.py
AlexMishKo: Done it, but i cant see any changes here on git hub. need ur help pls^^

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        return self.__add__(other)
+
+    def __iadd__(self, other: Union[int, float, Distance]) -> Distance:
+        other_km = self._get_km_value(other)
+        self.km += other_km
+        return self
+
+    def __mul__(self, other: Union[int, float]) -> Union[Distance, None]:
+        if isinstance(other, (int, float)):
+            return Distance(self.km * other)
```
oliasymak24, line 38: You don't need this check here, your function anyway takes only int and float types, according to your type annotations

app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        other_km = self._get_km_value(other)
+        return self.km == other_km
+
+    def __le__(self, other: Union[int, float, Distance]) -> bool:
+        other_km = self._get_km_value(other)
+        return self.km <= other_km
+
+    def __ge__(self, other: Union[int, float, Distance]) -> bool:
+        other_km = self._get_km_value(other)
+        return self.km >= other_km
```
oliasymak24: try to rewrite these two methods in one line each. Use previously created magic methods

oliasymak24 (changes requested):

DonDmytron (changes requested):
fix it, please 
Screenshot 2025-10-31 at 15 34 27

oliasymak24 (approved):

PR #2337, opened 2025-11-18
Update main.py

Sashaaaaaa6:


app/main.py
```diff
@@ -1,3 +1,76 @@
...
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", Number]) -> "Distance":
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        if isinstance(other, (int, float)):
+            return Distance(self.km + other)
+        return NotImplemented  # type: ignore
```
viktoria-rybenchuk, line 22: remove redundant comment

viktoria-rybenchuk: fix in all places

app/main.py
```diff
@@ -1,3 +1,76 @@
...
+    def __init__(self, km: float) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Union["Distance", Number]) -> "Distance":
```
viktoria-rybenchuk, line 17:
```suggestion
    def __add__(self, other: Distance | int) -> Distance:
```

viktoria-rybenchuk: fix annotations

viktoria-rybenchuk (changes requested):

PR #2365, opened 2025-12-03
Solution

qw1nch741:


app/main.py
```diff
@@ -1,3 +1,94 @@
...
+            return NotImplemented
+        return self.km < value
+
+    def __gt__(self, other: int | float | "Distance") -> bool:
+        if isinstance(other, Distance):
+            value = other.km
+        elif isinstance(other, (int, float)):
+            value = other
+        else:
+            return NotImplemented
```
vsmutok, line 66: DRY
Try moving this logic into a separate method to avoid duplication.

app/main.py
```diff
@@ -1,3 +1,94 @@
+from __future__ import annotations
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int) -> None:
```
vsmutok, line 5: The type hint int for km is a bit restrictive. Since division can result in a float value being stored in self.km, it's good practice to allow floats here as well. Consider using km: int | float.

vsmutok (changes requested):

app/main.py
vsmutok, line 66: DRY
Try moving this logic into a separate method to avoid duplication.

app/main.py
vsmutok, line 5: The type hint int for km is a bit restrictive. Since division can result in a float value being stored in self.km, it's good practice to allow floats here as well. Consider using km: int | float.

vsmutok (changes requested):

PR #2444, opened 2026-01-11
Add Distance class with magic methods

nazarkravets2007:


nazarkravets2007 (commented):
Add full Distance implementation with type annotations

PR #2499, opened 2026-01-28
complete task

PavloDanylkiv:


app/main.py
```diff
@@ -1,3 +1,78 @@
 class Distance:
-    # Write your code here
-    pass
+    def __init__(self, km: int | float) -> None:
+        self.km = km
+
+    def __add__(self, other: "Distance") -> "Distance":
```
vsmutok:
```suggestion
    def __add__(self, other: Distance | int | float) -> Distance:
```

app/main.py
```diff
@@ -1,3 +1,78 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(
+                km=self.km + other,
+            )
+        elif isinstance(other, Distance):
+            return Distance(
+                km=self.km + other.km,
+            )
```
vsmutok: Try to extract this logic into a separate method and use it everywhere you need it, to avoid duplication

vsmutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,101 @@
...
+        raise TypeError
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, (int, float)):
+            self.km += other
+        elif isinstance(other, Distance):
+            self.km += other.km
+        else:
+            raise TypeError
+        return self
```
DaTrEvTeR, line 22: This line is unreachable. Remove it

DaTrEvTeR (changes requested):

app/main.py
PavloDanylkiv: I just checked and it is reachable.

vsmutok (approved):
Great!

DaTrEvTeR (approved):

PR #2535, opened 2026-02-09
Solution

swirtam:


swirtam:
@mateacademy-ai-mentor I’ve modified my solution based on your advice. I’ve modified __iadd__ method to get rid of else statement. I’ve added the safety check to division in __truediv__ and I’ve written additional methods: __radd__, __rmul__ and __imul__. With __rmul__, __mul__ requires a check of the other’s type to prevent the possibility of multiplying 2 variables of Distance type.

PR #2541, opened 2026-02-12
Development

MykolaLozanDevelopment:
Fix type hints, remove unnecessary else, add future annotations


MykolaLozanDevelopment (commented):
Fix type hints, remove unnecessary else, add future annotations

PR #2556, opened 2026-02-19
Implement these magic methods

Ensenkei:


app/main.py
```diff
@@ -1,3 +1,59 @@
...
+            return self.km > other.km
+        return self.km > other
+
+    def __eq__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km == other.km
+        return self.km == other
+
+    def __le__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
```
vsmutok: Try to move this check into a separate method and reuse it wherever it’s needed.

vsmutok (changes requested):

app/main.py
Ensenkei: Thank you for the tip! I'll look for that

vsmutok (approved):
Well Done!

PR #2560, opened 2026-02-20
'Solution'

AlinaSv88:


app/main.py
```diff
@@ -1,3 +1,57 @@
...
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> object:
+        other = other.km if isinstance(other, Distance) else other
+
+        return Distance(self.km + other)
+
+    __radd__ = __add__
+
+    def __iadd__(self, other: int) -> Self:
```
Dimosphen1, line 22: Instead of using `Self` in the return type annotation, you can directly indicate that you return a `Distance` class instance here and in the other places as well

```suggestion
    def __iadd__(self, other: int) -> Distance:
```

Dimosphen1 (approved):
Looks good to me, well done! One comment is left for consideration.

PR #2563, opened 2026-02-21
Solution

Reverad:


app/main.py
```diff
@@ -1,3 +1,47 @@
+from typing import TypeAlias, SupportsFloat
+
```
elija-xp, line 3: It would be better to add 

 at the top of the file.
This prevents possible NameError caused by forward references in type annotations (e.g. when a class refers to itself).

elija-xp (changes requested):

elija-xp (approved):

elija-xp (approved):

PR #2602, opened 2026-03-10
fix1

ivansinickij29-cpu:


vsmutok (changes requested):
Make sure that your code passes all tests before submitting the task for review.

PR #2614, opened 2026-03-18
Implemented solution for py-distance-class

Neyrony:


Neyrony:
@mateacademy-ai-mentor i think removing else in def __iadd__ will lead to problem like if var is int, then conditional will be true and then after self.km += other.km, will become a error that int or float doesn't have this method

PR #2632, opened 2026-03-24
implement py-distance-class

zubak-vasil:


lubkoserhii (approved):
The code is well-structured and clean. The class is versatile, supporting operations with both other Distance instances and numeric types. Well done!

PR #2646, opened 2026-03-31
solution

Prus-Oleksandr:


amesemyta1 (approved):
def _get_km(self, other: "Distance") -> float:
        return other.km if isinstance(other, Distance) else other

Maybe static (@staticmethod)

PR #2704, opened 2026-04-14
Solution

jlaourayed:


jlaourayed (commented):
from __future__ import annotations    added

jlaourayed:
what is happning

PR #2711, opened 2026-04-16
'Solution'

yakinder46:


app/main.py
```diff
@@ -1,3 +1,64 @@
...
+        return round(self.meters_total, 2) == round(self._get_meters(other), 2)
+
+    def __le__(self, other: Any) -> bool:
+        return self.meters_total <= self._get_meters(other)
+
+    def __ge__(self, other: Any) -> bool:
+        return self.meters_total >= self._get_meters(other)
+
+
+class KiloDistance(Distance):
```
viktoria-rybenchuk, line 58: why do you need this code?

app/main.py
```diff
@@ -1,3 +1,64 @@
...
+
+
 class Distance:
-    # Write your code here
-    pass
+    def __init__(
+            self,
+            meters: int | float,
+            centimeters: int | float = 0,
+    ) -> None:
```
viktoria-rybenchuk, line 9: Distance's __init__ method takes only one argument km and saves it to self.km.

viktoria-rybenchuk (changes requested):

Boyok1337 (changes requested):
Provide answers for previous comments from mentor or fix your solution

PR #2721, opened 2026-04-21
add Distance class with operator overloading

DeeJee-cmd:


app/main.py
```diff
@@ -1,3 +1,68 @@
...
+        if isinstance(other, (int, float)):
+            return Distance(round(self.km / other, 2))
+        return NotImplemented
+
+    def __lt__(self, other: object) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        elif isinstance(other, (int, float)):
+            return self.km < other
+        return NotImplemented
```
vsmutok, line 43: You are performing this check multiple times.

Move this logic into a separate method and reuse it everywhere it is needed.

vsmutok (changes requested):

Boyok1337 (approved):
LGTM!

PR #2727, opened 2026-04-23
Implement Distance class with arithmetic and comparison operations

melmonreaz:


app/main.py
```diff
@@ -1,3 +1,55 @@
...
+
+    def __mul__(self, other: int | float) -> Distance:
+        return Distance(km=self.km * other)
+
+    def __truediv__(self, other: int | float) -> Distance:
+        return Distance(km=round(self.km / other, 2))
+
+    def __lt__(self, other: Distance | int | float) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
```
vsmutok: You perform this check several times.
Extract it into a separate method and use it wherever necessary.

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #2764, opened 2026-05-14
work is done

Sardor-Tozhiyev:
added int | float and from __future__ import annotations


Sardor-Tozhiyev (commented):
added int | float and from __future__ import annotations

PR #2786, opened 2026-05-31
Solution

Teashrock:


app/main.py
```diff
@@ -1,3 +1,53 @@
...
+        return Distance(round(self.km / other, 2))
+
+    def __lt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km < other.km
+        return self.km < other
+
+    def __gt__(self, other: int | float | Distance) -> bool:
+        if isinstance(other, Distance):
+            return self.km > other.km
```
vsmutok: You are performing this check multiple times. I recommend moving this logic into a separate method and reusing it wherever needed.

vsmutok (changes requested):

app/main.py
Teashrock: I thought about moving it to a separate method, thank you for noticing.

UPD: Resolved.

app/main.py
```diff
@@ -1,3 +1,46 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int | float | Distance) -> Distance:
+        return Distance(self.km + self.__check(other))
+
+    def __iadd__(self, other: int | float | Distance) -> Distance:
+        other_km = other.km if isinstance(other, Distance) else other
```
vsmutok: use `__check` here

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #2819, opened 2026-06-19
solution

GavriliukArtem:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            return Distance(km=self.km + other.km)
```
vsmutok, line 16: You perform this check several times.
Please extract it into a separate method and reuse it everywhere you need it.

vsmutok (changes requested):

PR #2820, opened 2026-06-19
feat: implement Distance Class

neeeksa:


app/main.py
```diff
@@ -1,3 +1,61 @@
...
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance | int:
+        if isinstance(other, Distance):
+            return Distance(self.km + other.km)
+        else:
+            return Distance(self.km + other)
```
vsmutok, line 18: You perform this check several times.
Please extract it into a separate method and reuse it everywhere you need it.

app/main.py
```diff
@@ -1,3 +1,61 @@
...
+    def __init__(self, km: int) -> None:
+        self.km = km
+
+    def __str__(self) -> str:
+        return f"Distance: {self.km} kilometers."
+
+    def __repr__(self) -> str:
+        return f"Distance(km={self.km})"
+
+    def __add__(self, other: int) -> Distance | int:
```
```suggestion
    def __add__(self, other: int | float | Distance) -> Distance:
```

vsmutok (changes requested):
f

PR #2921, opened 2026-08-08
Solution

FedirKiiko:


app/main.py
```diff
@@ -1,3 +1,62 @@
...
+
+    def __iadd__(self, other: Distance | int | float) -> Distance:
+        if isinstance(other, Distance):
+            self.km = self.km + other.km
+            return self
+
+        self.km = self.km + other
+        return self
+
+    def __mul__(self, other: int) -> Distance:
```
Avalanche-one: it is better to add float annotation for "other"

app/main.py
```diff
@@ -1,3 +1,62 @@
...
+        return Distance(self.km * other)
+
+    def __truediv__(self, other: int) -> Distance:
```
Avalanche-one: same here

Avalanche-one (changes requested):
Just add float annotations, rest of your solution - good

app/main.py
FedirKiiko: Thanks, added

Avalanche-one (approved):
Good

PR #2923, opened 2026-08-09
Solution

ArGus20:


MaksimDonchyk (approved):
1)__iadd__:Missing type annotation in __iadd__ like(def __iadd__(self, other: Distance | int | float) -> Distance:)
2) in __truediv__ and __mul__ in future add to testing like (if isinstance(other, (int, float)):
        return NotImplemented)

PR #2924, opened 2026-08-09
Solution

MaksimDonchyk:
I know it`s big, so maybe you know, another better solution?


Aragorn-untold (approved):
The size is fine, if you want to return NotImplemented then there is no other way to implement this code other than what you did. """Or I do not know THE way"""

